PR #75, opened 2017-04-13
Fixes #74 update usage of FormatWith

### LINKED ISSUE #74: Update usage of FormatWith

msevestre:


tests/MoBi.Tests/IntegrationTests/CoreRegisterSpecs.cs
```diff
@@ -0,0 +1,25 @@
...
+      protected override void Context()
+      {
+      }
+
+      [Observation]
+      public void should_be_able_to_find_a_diff_builder_for_a_simulation()
+      {
+         var diffBuilderRepository = IoC.Resolve<IDiffBuilderRepository>();
+         var simuationDiffBuilder = diffBuilderRepository.BuilderFor(new MoBiSimulation());
+         simuationDiffBuilder.ShouldBeAnInstanceOf<MoBiSimulationDiffBuilder>();
```
msevestre: ALso added a new test similar to PKsim as I was expecting that to break as well in MoBi. But this was implemented properly

PR #83, opened 2017-04-19
Fixes #80 history rollback command not set

### LINKED ISSUE #80: History does not report resetting a parameter
1. Open the ManualModel.
2. Open simulation S1.
3. Change any parameter value, e.g. MW of molecule A. Check the history, where the change is reported.
4. Click on the orange reset button left of the value.
5. Open the history. 
While the object type says "Parameter", the description field is empty. A parameter path and the value change would be expected here.

msevestre:


src/MoBi.Core/Services/QuantityTask.cs
```diff
@@ -107,14 +107,15 @@ public ICommand SetQuantityDisplayUnit(IQuantity quantity, Unit displayUnit, IMo
...
+         var macroCommand = new MoBiMacroCommand();
 
          //add one before setting the value in the simulation to enable correct undo
          macroCommand.Add(_quantitySynchronizer.Synchronize(quantity, simulation));
          macroCommand.Add(simulationCommand.AsHidden().Run(_context));
          macroCommand.Add(_quantitySynchronizer.Synchronize(quantity, simulation));
 
-         return macroCommand;
+         //needs to be done at the end because description might be set only after run
+         return macroCommand.WithHistoryEntriesFrom(simulationCommand);
```
msevestre: That's the big change here. For some commands, description is only available after RUN is called

tests/MoBi.Tests/Core/QuantityTaskSpecs.cs
```diff
@@ -0,0 +1,57 @@
...
+         _quantitySynchronizer = A.Fake<IQuantitySynchronizer>();
+         _context = A.Fake<IMoBiContext>();
+         sut = new QuantityTask(_context, _quantitySynchronizer);
+      }
+   }
+
+   public class When_resetting_the_value_of_a_quantity_in_a_simulation : concern_for_QuantityTask
+   {
+      private IParameter _parameter;
+      private ICommand _result;
```
msevestre: never too late to add tests

PR #101, opened 2017-05-31
Fixes #24 PI: Allow copying values from the results tab

### LINKED ISSUE #24: PI: Allow copying values from the results tab.
Results are shown in a table, which cells are not active (cannot select and copy the value). Being able to copy values is a must.
aufnahme2

msevestre:


appveyor-coverage.yml
```diff
@@ -1,7 +1,8 @@
 configuration: Debug
+image: Visual Studio 2017
 
 environment:
-  app_version: '7.1.0'
+  app_version: '7.2.0'
```
msevestre: This needed to be updated as well

src/MoBi.Assets/AppConstants.cs
```diff
@@ -1567,6 +1567,7 @@ public static class Captions
          public static readonly string MoleculeObserver = "Molecule Observer";
          public static readonly string ContainerObserver = "Container Observer";
          public static readonly string SelectMolecules = "Select Molecules";
+         public static readonly string LoadingApplication = "Loading Application...";
```
msevestre: This is just so that the caption in the splash screen does not show the name of the view!

src/MoBi.UI/MoBi.UI.csproj
```diff
@@ -525,12 +520,6 @@
     <Compile Include="Views\EditMoleculeBuilderView.Designer.cs">
       <DependentUpon>EditMoleculeBuilderView.cs</DependentUpon>
     </Compile>
-    <Compile Include="Views\EditNeighborhoodView.cs">
```
msevestre: This file was creating GridControl on the fly. Not used anywhere so I deleted it

PR #106, opened 2017-06-06
Fixes #105 - Use CurveNamer to name chart curves

### LINKED ISSUE #105: Use CurveNamer to name chart curves

rwmcintosh:
Also updating Core components and DevExpress

### DIFF
--- a/src/MoBi.Presentation/Presenter/SimulationChartPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/SimulationChartPresenter.cs
@@ -1,16 +1,13 @@
-﻿using System;
-using System.Linq;
-using OSPSuite.Utility.Extensions;
-using MoBi.Core;
-using MoBi.Core.Domain.Model;
+﻿using MoBi.Core.Domain.Model;
 using MoBi.Presentation.Settings;
 using MoBi.Presentation.Tasks;
 using MoBi.Presentation.Views;
 using OSPSuite.Core.Domain.Data;
-using OSPSuite.Core.Domain.Mappers;
+using OSPSuite.Core.Services;
 using OSPSuite.Presentation.Mappers;
 using OSPSuite.Presentation.Presenters.Charts;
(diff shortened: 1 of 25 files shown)


src/MoBi.Presentation/Presenter/SimulationChartPresenter.cs
```diff
@@ -40,9 +37,7 @@ protected override void MarkChartOwnerAsChanged()
 
       protected override string CurveNameDefinition(DataColumn column)
       {
-         var simulationForDataColumn = _simulations[column.Repository];
-
-         return _quantityDisplayPathMapper.DisplayPathAsStringFor(simulationForDataColumn, column);
+         return _curveNamer.CurveNameForColumn(_simulations[column.Repository], column, addSimulationName: false);
```
rwmcintosh: The code change.

There's a namer in ComparisonChartsPresenter too that uses a different override of the mapper. Should we consolidate that into the CurveNamer as well?

msevestre: Let's create another issue for that so that we do not forget. I'd like to avoid doing to many changes in the Curve department at the moment

tests/MoBi.Tests/Presentation/Tasks/ObservedDataTasksSpecs.cs
```diff
@@ -104,9 +107,9 @@ protected override void Because()
       }
 
       [Observation]
-      public void should_change_Name_of_data_repository_to_new_name()
+      public void should_use_the_data_repository_renamer_to_update_data_repository_name()
       {
-         A.CallTo(() => _context.AddToHistory(A<RenameObservedDataCommand>._)).MustHaveHappened();
+         A.CallTo(() => _dataRepositoryNamer.Rename(_dataRepository, _newName)).MustHaveHappened();
```
msevestre: YES!

PR #108, opened 2017-06-08
Fixes #94 - Reset Logic Changed

### LINKED ISSUE #94: Reset logic changed ?!
After commit of all BB in Sim P1 IV, closing and reopening Simulation, Values have still Reset button and are resetted to some earlier value.
As far as I remember, reset buttons vanished in such a situation in earlier versions.

screenshot2

Michael: 
ok so I looked at the issue and here is what I can say:
- This is not new. This was introduced with the commit update logic in 5.6
- This problem is only for parameters that are not defined in PSV and for which an entry is not defined in PSV when overriding a value. Those parameters are for example global reaction parameters or property molecule parameters.

rwmcintosh:


src/MoBi.Core/Commands/ResetFixedConstantParametersToDefaultInSimulationCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+         _simulation = simulation;
+         _buildingBlockFromSimulation = buildingBlockFromSimulation;
+      }
+
+      protected override void ExecuteWith(IMoBiContext context)
+      {
+         _formulaTask = context.Resolve<IMoBiFormulaTask>();
+         _affectedBuildingBlockRetriever = context.Resolve<IAffectedBuildingBlockRetriever>();
+
+         var fixedParameters = fixedConstantParametersFromBuildingBlock().ToList();
```
msevestre: you don't need the ToList here

src/MoBi.Core/Commands/ResetFixedConstantParametersToDefaultInSimulationCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+         return _simulation.Model.Root.GetAllContainersAndSelf<IContainer>()
+            .SelectMany(container => container.AllParameters(parameterIsConstantFixedAndFromBuildingBlock));
+      }
+
+      private bool parameterIsConstantFixedAndFromBuildingBlock(IParameter parameter)
+      {
+         return parameter.IsFixedValue && parameter.Formula.IsConstant() && isFromBuildingBlock(parameter);
+      }
+
+      private bool isFromBuildingBlock(IParameter x)
```
msevestre: rename parameter `x` to `parameter`

src/MoBi.Core/Commands/ResetFixedConstantParametersToDefaultInSimulationCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+      {
+         return Equals(_affectedBuildingBlockRetriever.RetrieveFor(x, _simulation).UntypedBuildingBlock, _buildingBlockFromSimulation);
```
msevestre: that retrieve for may be null?

src/MoBi.Core/Commands/ResetFixedConstantParametersToDefaultInSimulationCommand.cs
```diff
@@ -0,0 +1,71 @@
...
+      }
+
+      private void resetQuantity(IQuantity fixedQuantity)
```
msevestre: `fixexQuantity` => `quantity`

msevestre:
Looking good. Serious testing on real model should be done however to make sure that this is working as expected

src/MoBi.Core/Commands/ResetFixedConstantParametersToDefaultInSimulationCommand.cs
rwmcintosh: That means the enumerable will evaluate more than once no?

rwmcintosh: fixedConstantParametersFromBuildingBlock can return IReadOnlyList

PR #117, opened 2017-06-16
Fixes #116 - Update Core to support new charts

### LINKED ISSUE #116: Update Core to support new charts

rwmcintosh:


src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -69,52 +70,68 @@ public abstract class ChartPresenter : AbstractCommandCollectorPresenter<IChartV
       private readonly IDataColumnToPathElementsMapper _dataColumnToPathElementsMapper;
       private readonly IChartEditorLayoutTask _chartEditorLayoutTask;
       protected readonly ICache<DataRepository, IMoBiSimulation> _simulations;
-      private readonly IChartDisplayPresenter _displayPresenter;
-      private readonly IChartEditorPresenter _editorPresenter;
+      private IChartDisplayPresenter _displayPresenter;
+      private IChartEditorPresenter _editorPresenter;
       private readonly string _nameProperty;
```
msevestre: That should not be required anymore

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -69,52 +70,68 @@ public abstract class ChartPresenter : AbstractCommandCollectorPresenter<IChartV
...
       {
-         _chartEditorAndDisplayPresenter.SetNoCurvesSelectedHint(string.Empty);
+         _displayPresenter.SetNoCurvesSelectedHint(string.Empty);
       }
 
       private void initializeNoCurvesHint()
       {
-         _chartEditorAndDisplayPresenter.SetNoCurvesSelectedHint(AppConstants.PleaseSelectCurveInChartEditor);
+         _displayPresenter.SetNoCurvesSelectedHint(AppConstants.PleaseSelectCurveInChartEditor);
       }
```
msevestre: put that into initializeDisplayPresenter and remove this methid

msevestre: unless this method is used somewhere else that is :)

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -190,54 +207,76 @@ protected virtual void OnDragDrop(object sender, DragEventArgs e)
...
-               Chart.PropertyChanged += onPropertyChanged;
-         }
+         _editorPresenter.Clear();
+         _displayPresenter.Clear();
+      }
+
+      private void releasePropertyChanged()
+      {
+         if (Chart != null)
+            Chart.PropertyChanged -= onPropertyChanged;
```
msevestre: remove that. That should not be used anymore and won't be fired as it used to anyways

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -190,54 +207,76 @@ protected virtual void OnDragDrop(object sender, DragEventArgs e)
...
+      }
+
+      private void subscribePropertyChanged()
```
msevestre: same with that

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -54,9 +54,10 @@ public interface IChartPresenter : IPresenter<IChartView>,
...
       ///    This method ensures the correct order of parameter setting
       /// </remarks>
-      void Show(ICurveChart chart, IReadOnlyList<DataRepository> dataRepositories, CurveChartTemplate defaultTemplate = null);
+      void Show(CurveChart chart, IReadOnlyList<DataRepository> dataRepositories, CurveChartTemplate defaultTemplate = null);
 
       void UpdateTemplatesBasedOn(IWithChartTemplates withChartTemplates);
+      void Edit(CurveChart curveChart);
    }
 
    public abstract class ChartPresenter : AbstractCommandCollectorPresenter<IChartView, IChartPresenter>, IChartPresenter
```
msevestre: Probably that guy should actually inherit from the OSPSuite.Core ChartPresenter no?

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -190,54 +207,76 @@ protected virtual void OnDragDrop(object sender, DragEventArgs e)
...
+      {
+         releasePropertyChanged();
+
+         _editorPresenter.Edit(curveChart);
+         _displayPresenter.Edit(curveChart);
+
+         subscribePropertyChanged();
       }
 
       private void onPropertyChanged(object sender, PropertyChangedEventArgs e)
```
msevestre: There is the method ChartChanged that can be overriden in the base ChartPresenter.
If for some reason, this cannot be used, you need to react to the event ChartChanged event of the ChartEditorPresenter

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -274,14 +313,17 @@ public virtual void UpdateTemplatesBasedOn(IWithChartTemplates withChartTemplate
...
       }
 
-      private void addDataRepositorySimulation(DataRepository dataRepository)
+      private void addDataRepositoriesToSimulation(IReadOnlyCollection<DataRepository> dataRepositories)
       {
-         if (_simulations.Contains(dataRepository))
-            return;
+         dataRepositories.Each(dataRepository =>
+         {
+            if (_simulations.Contains(dataRepository))
```
msevestre: use a where condition before the Each. `dataRepositories.Where(x=>!_simulation.Contains(x)).Each`

src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
```diff
@@ -94,13 +93,14 @@ public void Edit(IMoBiSimulation simulation)
          _solverSettingsPresenter.Edit(_simulation);
          _editOutputSchemaPresenter.Edit(_simulation);
          _favoritesPresenter.Edit(_simulation);
-         _chartPresenter.UpdateTemplatesBasedOn(_simulation);
+         _chartPresenter.UpdateTemplatesFor(_simulation);
          _view.SetEditView(_favoritesPresenter.BaseView);
          UpdateCaption();
          _view.Display();
+         loadChart();
```
rwmcintosh: that means we are going to load the chart when the simulation is edited

src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
```diff
@@ -114,50 +114,46 @@ public override void Edit(object subject)
...
       }
 
-      public override object Subject
-      {
-         get { return _simulation; }
-      }
+      public override object Subject => _simulation;
 
-      public void ShowData()
+      private void loadChart()
```
rwmcintosh: does not get called from outside the presenter anymore

src/MoBi.UI/Views/SimulationView/EditSimulationView.cs
```diff
@@ -74,19 +65,22 @@ public void SetTreeView(IView view)
 
       public void SetChartView(IView view)
       {
+         view.CaptionChanged += (o, e) => OnEvent(() => tabResults.Text = captionNameFrom(view));
```
rwmcintosh: Binding the tab text to the caption of the chart view

src/MoBi.Presentation/Presenter/ChartPresenter.cs
rwmcintosh: it's used in two spots

src/MoBi.Core/Reporting/ChartsReporter.cs
```diff
@@ -1,28 +1,26 @@
...
 using System.Linq;
-using OSPSuite.TeXReporting.Items;
 using OSPSuite.Core.Chart;
 using OSPSuite.Infrastructure.Reporting;
+using OSPSuite.TeXReporting.Items;
 
 namespace MoBi.Core.Reporting
 {
-   internal class ChartsReporter : OSPSuiteTeXReporter<IEnumerable<ICurveChart>>
+   internal class ChartsReporter : OSPSuiteTeXReporter<IReadOnlyCollection<CurveChart>>
```
msevestre: Did you check that this is working? If what's passed to the Reporter is an actual IEnumerable, this TeXReporter won't be found

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -54,67 +50,74 @@ public interface IChartPresenter : IPresenter<IChartView>,
...
-         _chartEditorAndDisplayPresenter.SetCurveNameDefinition(CurveNameDefinition);
-         _nameProperty = MoBiReflectionHelper.PropertyName<CurveChart>(x => x.Name);
+
          _observedDataDragDropBinder = new ObservedDataDragDropBinder();
 
-         AddSubPresenters(_chartEditorAndDisplayPresenter);
+         AddSubPresenters(chartPresenterContext.EditorAndDisplayPresenter);
+      }
+
+      private void initializeEditorPresenter(IChartEditorAndDisplayPresenter chartEditorAndDisplayPresenter)
```
msevestre: why do you pass that guy as a parameter?Cannot you use your member instead?

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -190,76 +193,56 @@ protected virtual void OnDragDrop(object sender, DragEventArgs e)
...
-      }
-
-      private void onPropertyChanged(object sender, PropertyChangedEventArgs e)
+      protected override void ChartChanged()
       {
+         base.ChartChanged();
          MarkChartOwnerAsChanged();
-         if (e.PropertyName.Equals(_nameProperty))
-         {
-            _context.PublishEvent(new RenamedEvent(Chart));
```
msevestre: Is someone else relying on this event being fired?

src/MoBi.UI/Views/SimulationView/EditSimulationView.cs
msevestre: That could be moved to the presenter so that the behavior can also be tested...

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -54,67 +50,74 @@ public interface IChartPresenter : IPresenter<IChartView>,
...
+         initializeDisplayPresenter(chartPresenterContext.EditorAndDisplayPresenter);
+         initializeEditorPresenter(chartPresenterContext.EditorAndDisplayPresenter);
 
          _chartTemplatingTask = chartTemplatingTask;
-         _simulations = new Cache<DataRepository, IMoBiSimulation>(onMissingKey: x => null);
+         _dataRepositoryCache = new Cache<DataRepository, IMoBiSimulation>(onMissingKey: x => null);
 
-         _dataColumnToPathElementsMapper = dataColumnToPathElementsMapper;
-         _chartEditorLayoutTask = chartEditorLayoutTask;
+         _dataColumnToPathElementsMapper = chartPresenterContext.DataColumnToPathElementsMapper;
```
msevestre: no need to save all those guys as members. That's why the context is all about

src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
```diff
@@ -40,7 +39,7 @@ public class EditSimulationPresenter : SingleStartPresenter<IEditSimulationView,
       private bool _diagramLoaded;
       private readonly IHeavyWorkManager _heavyWorkManager;
       private readonly IChartFactory _chartFactory;
-      private bool _dataShown;
+      private bool _chartLoaded;
```
msevestre: why do we still need this flag? i thought we said this was not required anymore and we should load the chart on Edit?

src/MoBi.Presentation/Presenter/ChartPresenter.cs
rwmcintosh: Not for the chart. I think this was left when the tab caption was being updated with chart name. Probably some time after that the chart was loaded only when the tab was clicked and this stopped working.

When I stepped through the handlers of RenamedEvent, it's not used for chart.

PR #134, opened 2017-07-03
Fixes #111 Favorites are not automatically specified for parameter id…

### LINKED ISSUE #111: Favorites are not automatically specified for parameter identification
When starting a parameter identification by right click on simulation, parameters defined as favorites were automatically specified as parameters to identify. That does not work in the Version 7.1, Build 66

msevestre:
…entification

### DIFF
--- a/src/MoBi.Core/Commands/UpdateSimulationCommand.cs
+++ b/src/MoBi.Core/Commands/UpdateSimulationCommand.cs
@@ -11,6 +11,7 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Assets;
+using OSPSuite.Core.Services;
 
 namespace MoBi.Core.Commands
 {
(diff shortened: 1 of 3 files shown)


src/MoBi.Core/Commands/UpdateSimulationCommand.cs
```diff
@@ -73,6 +73,14 @@ protected override void ExecuteWith(IMoBiContext context)
          context.PublishEvent(new SimulationReloadEvent(_simulationToUpdate));
       }
 
+      private void updateReferencesToSimulation(IMoBiContext context)
+      {
+         var simulationReferenceUpdater = context.Resolve<ISimulationReferenceUpdater>();
+         simulationReferenceUpdater.SwapSimulationInParameterAnalysables(_simulationToUpdate, _simulationToUpdate);
+
+         var simulationParameterOriginIdUpdater = context.Resolve<ISimulationParameterOriginIdUpdater>();
+         simulationParameterOriginIdUpdater.UpdateSimulationId(_simulationToUpdate);
```
msevestre: That needed to be called after updating the simulation as new parameters do not have the reference to the simulation

PR #167, opened 2017-08-24
Fixes #140 multiple conversion

### LINKED ISSUE #140: Diagram conversion from older version done each time
This is an issue we had before (see YouTrack # 46-7209). When loading a model created with an older version of MoBi, a conversion is done even though the project had been saved several times already with MoBi 7.1. This will lead to the safety question "Do you want to save the current project?" each time the project is only opened but nothing has been changed. It also will bloat the History list with entries as shown below whenever the project is really saved again, i.e., after making real changes.

State After Action	Command Type	Object Type	Description	User	Time
3615	Info	Diagram	Diagram 'Diagram' converted from version '6.3.1' to '7.1.0'	<user> 	2017-07-10 16:16

msevestre:
Very similar to what happened in PKSim. We only throw the message if the conversion REALLY happened as opposed to every time that a version is being changed

### DIFF
--- a/src/MoBi.Core/Serialization/Converter/v3_5/Converter341To351.cs
+++ b/src/MoBi.Core/Serialization/Converter/v3_5/Converter341To351.cs
@@ -1,12 +1,10 @@
 ﻿using System.Linq;
 using System.Xml.Linq;
 using MoBi.Assets;
-using OSPSuite.Serializer.Xml.Extensions;
-using OSPSuite.Utility.Extensions;
-using OSPSuite.Utility.Visitor;
 using MoBi.Core.Domain.Builder;
 using MoBi.Core.Domain.Model;
 using MoBi.Core.Domain.Services;
+using OSPSuite.Assets;
 using OSPSuite.Core.Converter.v5_5;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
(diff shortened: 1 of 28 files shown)


src/MoBi.Core/Serialization/Converter/v3_5/Converter341To351.cs
```diff
@@ -93,14 +93,17 @@ public  int ConvertXml(XElement element, IMoBiProject project)
          //only happens when loading an actual MoBiSimulation from a project
          _loadingMoBiSimulation = (element.Name == AppConstants.XmlNames.MoBiSimulation);
 
-         return ProjectVersions.V3_5_1;
+         //too many conversion happening here! set to true for compatibility
+         return (ProjectVersions.V3_5_1, true);
```
msevestre: setting to true here is just a simple way to be compatible. Better safe than sorry

src/MoBi.Core/Serialization/Converter/v6_0/Converter36To601.cs
```diff
@@ -16,32 +16,42 @@ public class Converter36To601 : IMoBiObjectConverter,
...
+         var (_, coreConversionHappened) = _coreConverter56To601.ConvertXml(element);
          element.DescendantsAndSelfNamed("DisplayUnitMap").Each(displayUnitMapElement => convertPerTimeDimension(displayUnitMapElement, "dimension"));
          element.DescendantsAndSelfNamed("Formula").Each(displayUnitMapElement => convertPerTimeDimension(displayUnitMapElement, "dim"));
          element.DescendantsAndSelfNamed("Map").Each(displayUnitMapElement => convertPerTimeDimension(displayUnitMapElement, "s"));
          element.DescendantsAndSelfNamed("Parameter").Each(convertParameterQuantityType);
-         return ProjectVersions.V6_0_1;
+         return (ProjectVersions.V6_0_1, _converted || coreConversionHappened);
+      }
+
+      public void Visit(IMoBiProject project)
```
msevestre: moving public method to the top is a simple way to see what's going on with the component

PR #175, opened 2017-09-18
Fixes #172 Read license agreement without function

### LINKED ISSUE #172: "Read license agreement" without function
When selecting "File -> About -> Read license agreement", nothing happens.

msevestre:


msevestre:
merging

PR #190, opened 2017-10-24
188 draft watermark support

msevestre:


msevestre:
@Yuri05 very similar implementation to PKSim. Calling the checker and making sure that the option is available in the UI. I had to create the application settings for that

PR #206, opened 2017-11-22
Fixes #205 Load Container From Template does not work

### LINKED ISSUE #205: Load Container From Template does not work
Example workflow to reproduce
* Open **PK_Manual_Diclofenac** project installed with MoBi
* Open Spatial Structure building block save **Organism|Bone** to **Bone.pkml**
* In the Spatial Structure building block go to **Organism** and select **Load Container from Template...** from its context menu
* Select saved **Bone.pkml**

Now a rename dialog should appear (in order to rename Bone because it already exists under Organism).
Instead, **Select parameters to load ...** dialog appears.
At the end, no new container is imported; all selected parameters are added directly under organism.

msevestre:


src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForContainer.cs
```diff
@@ -68,7 +68,7 @@ protected override IMenuBarItem CreateDeleteItemFor(TContainer objectToRemove)
       private IMenuBarItem createAddExistingFromTemplateItemFor<T>(TContainer parent) where T : class
       {
          var typeName = _objectTypeResolver.TypeFor<T>();
-         return CreateMenuButton.WithCaption(AppConstants.MenuNames.AddExistingFromTemplate(ObjectTypeName))
+         return CreateMenuButton.WithCaption(AppConstants.MenuNames.AddExistingFromTemplate(typeName))
```
msevestre: the wrong object type was used. It should be the child to load (e.g. typename) as opposed to containing object name (container)

PR #240, opened 2018-02-13
Fixes #239 order of favorites should be editable

### LINKED ISSUE #239: The order of Favorites should be editable
Open-Systems-Pharmacology/OSPSuite.Core#237

msevestre:


src/MoBi.Presentation/Presenter/EditFavoritesPresenter.cs
```diff
@@ -1,102 +1,73 @@
...
+      void MoveDown();
+
    }
 
    public interface IEditFavoritesPresenter<T> : IEditFavoritesPresenter, IEditPresenter<T> where T : IObjectBase
    {
    }
 
-   public abstract class EditFavoritesPresenter<T> : AbstractParameterBasePresenter<IEditParameterListView, IEditParameterListPresenter>,
+   public abstract class EditFavoritesPresenter<T> : AbstractCommandCollectorPresenter<IEditFavoritesView, IEditFavoritesPresenter>,
```
msevestre: Same idea as in PKSim. Using composition over inheritance to drive the design

src/MoBi.Presentation/Presenter/EditFavoritesPresenter.cs
```diff
@@ -1,102 +1,73 @@
...
-         IParameterToParameterDTOMapper parameterDTOMapper, IFavoriteRepository favoriteRepository,
-         IInteractionTasksForParameter parameterTask, IFavoriteTask favoriteTask, IEntityPathResolver entityPathResolver, IViewItemContextMenuFactory viewItemContextMenuFactory)
-         : base(
-            view, quantityTask, interactionTaskContext,
-            formulaMapper, parameterTask, favoriteTask)
+      protected EditFavoritesPresenter(
+         IEditFavoritesView view,
+         IFavoriteRepository favoriteRepository,
+         IEntityPathResolver entityPathResolver,
+         IEditParameterListPresenter editParameterListPresenter,
```
msevestre: We did not have any stand alone implementation of the EditParameterListPresenter... now we do :)

src/MoBi.Presentation/Presenter/EditFavoritesPresenter.cs
```diff
@@ -197,5 +141,22 @@ public void Handle(RemovedEvent eventToHandle)
...
          refresh();
       }
+
+      public void MoveUp() => moveFavorites(x => x.MoveUp);
+
+      public void MoveDown() => moveFavorites(x => x.MoveDown);
+
+      private void moveFavorites(Func<IFavoriteTask, Action<IEnumerable<string>>> moveActionFunc)
+      {
+         var selectedParameters = _editParameterListPresenter.SelectedParameters;
```
msevestre: Same code as in PKSim. Not sure how to not duplicate the code since we have complete different presenters

src/MoBi.UI/Views/EditParameterListView.cs
```diff
@@ -215,10 +216,30 @@ public void SetVisibility(PathElement pathElement, bool isVisible)
          _pathBinder.SetVisibility(pathElement, isVisible);
       }
 
+      public IReadOnlyList<ParameterDTO> SelectedParameters
+      {
+         get { return _gridView.GetSelectedRows().Select(rowHandle => _gridViewBinder.ElementAt(rowHandle)).ToList(); }
+         set
```
msevestre: THis code should be moved in one single place. This is exactly the same in PKSim and MoBi. I thin GridViewBinder would be  a good place but I need to make sure that this works before moving it. Can be done later

PR #249, opened 2018-04-06
Fixes #248 Add an option to disable Circular Reference Check

### LINKED ISSUE #248: Add an option to disable Circular Reference Check
see https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/295

msevestre:


src/MoBi.Assets/AppConstants.cs
```diff
@@ -1579,6 +1579,7 @@ public static class Captions
          public static readonly string ExportObservedDataToExcel = "Export observed data to Excel";
          public static readonly string ContainerPath = "Container Path";
          public static readonly string ValidateRules = "Validate value constraints";
+         public static readonly string PerformCircularReferenceCheck = "Perform circular reference check (uncheck only if you have some performance issue while creating a model)";
```
msevestre: On by default. To turn off for big big models (Pavel had one that was taking a lot of time and sometime even out of memory...). Because it's so rare, let's not spend too much time to find out why and allow disable instead

Yuri05: I would extend the message with something like "Please make sure to save your work before starting any simulation/PI".
Because meanwhile this check is also turned off in SimModel and starting a calculation in case of circular references will result in stack overflow and full program crash.

src/MoBi.Presentation/Presenter/FinalOptionsPresenter.cs
```diff
@@ -1,48 +1,35 @@
...
-
+using OSPSuite.Assets;
 using OSPSuite.Presentation.Presenters;
 
 namespace MoBi.Presentation.Presenter
 {
    public interface IFinalOptionsPresenter : ISimulationItemPresenter
    {
-      void DoFinalActions(IMoBiSimulation simulation);
    }
```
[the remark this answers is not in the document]
msevestre: Still use to register the presenter with the container.

src/MoBi.Assets/AppConstants.cs
msevestre: @Yuri05 This will be too long for the check box text I fear.Will try

msevestre: done

PR #254, opened 2018-04-23
Fixes #170 Parameter with flag CanBeVaried set to false should be exc…

### LINKED ISSUE #170: Parameter with flag CanBeVaried set to false should be excluded from PI
This should be in implementation of https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/316
Possibly ReadOnly parameter as well. (A ReadOnly Parameter should never be CanBeVaried anyways)

msevestre:
…luded from PI


msevestre:
Can be varied parameter cannot be used in PI

msevestre:
@Yuri05 Review when you have time. Trivial change

PR #271, opened 2018-05-15
Fixes #269 Automatic switch of ValueOrigin to unknown and Fixes #268

### LINKED ISSUE #268: Chart editor - many data points and scrolling
I have many experimental data points (n=100+). When trying to make them visible or not visible on the chart, if the row is not within the view box, if 'visible' is clicked, it scrolls back up to the top. Then I have to scroll down to find the next row that I want to have visible. Disabling the automatic scrolling would be very helpful.

### LINKED ISSUE #269: Automatic switch of ValueOrigin to unknown

msevestre:


msevestre:
@Yuri05 Same changes as in PK-Sim .Please review

PR #278, opened 2018-05-18
Fixes #218 update building block templates

### LINKED ISSUE #218: Generate new PKML templates from PK-Sim
This should be done at the end of the release

msevestre:


msevestre:
@Yuri05 Also update Groups

PR #302, opened 2018-07-10
Fixes #142 migration to x64 bit

### LINKED ISSUE #142: Feature request: x64 bit version
Due to high memory demands, a 64-bit version of PK-Sim/MoBi is needed.

msevestre:


src/MoBi/MoBi.csproj
```diff
@@ -268,8 +268,8 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>call "$(VS140COMNTOOLS)vsvars32.bat"
-editbin /largeaddressaware "$(TargetPath)"</PostBuildEvent>
```
msevestre: @Yuri05  This should not be required anymore correct?

msevestre:
@Yuri05 Just updated all dll to use 64bit. and changes project type
I had to update sbml to a newer version as well and I remove the postbuild events applying the largeaddress patch

msevestre:
@Yuri05 Do not merge yet .Some integrations tests are failing.  also SBML does not seem to work...fun fun

msevestre:
Actually it works. Wrong dll loaded

src/MoBi/MoBi.csproj
Yuri05: correct

PR #364, opened 2019-03-24
Fixes #363 migrate to new osp suite

### LINKED ISSUE #363: Migrate to new OSPSuite Structure

msevestre:


MoBi.sln
```diff
@@ -31,6 +31,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MoBi.UI.Tests", "tests\MoBi.UI.Tests\MoBi.UI.Tests.csproj", "{2710573C-9216-4817-AD81-97099597A780}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MoBi.Engine", "src\MoBi.Engine\MoBi.Engine.csproj", "{15ED932C-EDEC-407B-91D9-7BCB8E7F95F5}"
```
msevestre: In MoBi. The code is so well split as in PKSim. IT uses a lot of core objects directly. So we need to have a specific dll
Also libscml is also x64 so it's required

msevestre:
yep worked. Merging For now

src/MoBi.Presentation/Presenter/EditExplicitFormulaPresenter.cs
```diff
@@ -201,7 +200,7 @@ public void Validate(string formulaString)
             _view.SetParserError(null);
             _context.PublishEvent(new FormulaValidEvent(_formula, BuildingBlock));
          }
-         catch (FuncParserException parserException)
+         catch (OSPSuiteException parserException)
```
Yuri05: What is the reason for this change?

msevestre: FuncParserException  comes from FuncParser which cannot be referenced from the presenter layer (AnyCPU vs x64 bit).

msevestre: And the validation method now throws an OSPSuiteException so this would not even work anymore :)

PR #372, opened 2019-04-17
Fixes #351 configure simulation

### LINKED ISSUE #351: (Re-)configuration of existing simulations in MoBi such as in PK-Sim
In PK-Sim, when clicking on an exisiting simulation, an option to configure the simulation is presented.
The configure option allows to exchange building blocks very efficiently. In MoBi such an option does not exist. However MoBi opens the configuration menu of a simulation only, when for example the reaction and molecule building blocks have been changed and 'update from building block' is clicked.
The configuration menu should be available as in PK-Sim.

msevestre:


src/MoBi.Core/Commands/UpdateSimulationCommand.cs
```diff
@@ -46,6 +51,11 @@ public UpdateSimulationCommand(IMoBiSimulation simulationToUpdate, IModel newMod
          _changedBuildingBlockType = changedBuildingBlockType;
          ObjectType = ObjectTypes.Simulation;
          CommandType = AppConstants.Commands.UpdateCommand;
+
+         if (string.IsNullOrEmpty(_changedBuildingBlockName))
```
msevestre: We had all the logic already to update a single building block. I am just ensuring that we reuse the same logic for configuration by changing the description of the command

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForSimulation.cs
```diff
@@ -40,10 +40,11 @@ public IContextMenu InitializeWith(IObjectBaseDTO dto, IPresenter presenter)
          {
             createEditItem(simulation),
             createRenameItem(simulation),
-            
 
-            createNewSimulationMenuBarItem(),
-            createAddExistingSimulationMenuBarItem(),
+            createConfigure(simulation),
+            
+//            createNewSimulationMenuBarItem(),
```
msevestre: @Yuri05  I am not sure WHY we have those guys where on a simuation? It does not make sense to be to have those menus at the Folder level and the simulation level, especially because there is absolutely no link with the selected simulation

src/MoBi.Presentation/Tasks/SimulationUpdateTask.cs
```diff
@@ -82,33 +80,59 @@ public ICommand<IMoBiContext> UpdateSimulationFrom(IMoBiSimulation simulationToU
             fixedValueQuantities.AddRange(simulationToUpdate.Model.Root.GetAllChildren<IQuantity>(x => x.IsFixedValue));
          }
 
-         return updateSimulation(simulationToUpdate, templateBuildingBlock, buildConfigurationReferencingTemplate, configurationCommands, fixedValueQuantities);
+         return updateSimulation(simulationToUpdate, buildConfigurationReferencingTemplate, configurationCommands, templateBuildingBlock, fixedValueQuantities);
       }
 
-      private ICommand<IMoBiContext> updateSimulation(IMoBiSimulation simulationToUpdate, IBuildingBlock templateBuildingBlock, IMoBiBuildConfiguration buildConfigurationReferencingTemplates, IMoBiCommand configurationCommands, PathCache<IQuantity> fixedValueQuantities)
+      public ICommand ConfigureSimulation(IMoBiSimulation simulationToConfigure)
```
msevestre: The configuree simulation is exactly the same as update without the quick update

src/MoBi.Assets/ToolTips.cs
```diff
@@ -98,6 +98,8 @@ public static class SimulationRibbon
          public static readonly string RunWithSettingsDescription = "Select simulation settings and simulate the active simulation";
          public static readonly string StopSimulation = "Stop the currently running Simulation";
          public static readonly string CalculateScaleFactors = "Calculate scale factors";
+         public static readonly string ConfigureSimulationDescription = "Configure Simulation";
```
Yuri05: In `RunWithSettingsDescription `it's _**s**imulation_,
in `StopSimulation `and `ConfigureSimulationDescription ``it's _**S**imulation_
Should be consistent

src/MoBi.Presentation/Tasks/Edit/EditTasksForSimulation.cs
```diff
@@ -132,23 +131,28 @@ public void CalculateScaleFactors(IMoBiSimulation simulation)
...
                return;
 
-            AddCommand(command);
+            addCommand(command);
          }
       }
 
+      public void Configure(IMoBiSimulation simulation)
+      {
+         throw new System.NotImplementedException();
```
Yuri05: Is this PR ready to merge or should I wait for more upcoming commits?

msevestre: This method should be gone. THat's where I started the implementation. Good catch. Will remove and update caption

msevestre: This is so weird. I don't see this code... hum going to update again

msevestre: oh it's because there is EditTaskForSimulation and EditTasksForSimulation in MoBi...

src/MoBi.Presentation/Tasks/Edit/EditTasksForSimulation.cs
```diff
@@ -164,5 +170,18 @@ public void ExportSimModelXml(IMoBiSimulation simulation)
          if (fileName.IsNullOrEmpty()) return;
          _simModelExporter.Export(simulation, fileName);
       }
+
+      public override void Rename(IMoBiSimulation simulationToRename, IEnumerable<IObjectBase> existingObjectsInParent, IBuildingBlock buildingBlock)
```
msevestre: That was the only code in EditTaskForSimulation. This class is now gone!

src/MoBi.Assets/ToolTips.cs
msevestre: actually nothing was consistent here. I have renamed everything to be consistent (Simulation, Building Block etc.. all low caps)

PR #383, opened 2019-05-03
Fixes #379 dimensionless dealt with properly

### LINKED ISSUE #379: Export from MoBi into population simulation with dimensionless parameter state variable
Hello all,

Please ignore my posting in the Forum, I think this is the wrong location.

I am having an issue when exporting a MoBi file into the population simulator.

I needed to introduce a dimensionless parameter state variable (fraction) but the export to the population simulator seem not to allow such step. Is there a solution to this problem or can I work around it?

The message: Could not load simulation from ...
Dimension dimensionless per time not available in DimensionFactory.

Picture1

Thank you.

Best regards,

Rudiger

msevestre:


src/MoBi.Assets/AppConstants.cs
```diff
@@ -1969,7 +1969,11 @@ public static string MoleculeNodeAlreadyExistsForMolecule(string moleculeName)
 
       public static string RHSDefaultUnitName(IDimension dimension)
       {
-         return $"{dimension.BaseUnit.Name}/min";
+         var numerator = string.IsNullOrEmpty(dimension.BaseUnit.Name) ? "1" : dimension.BaseUnit.Name;
```
msevestre: Such as stupid mistake here. The unit returned was "/min" instead of "1/min"!
Same with min/min => I return string.empty

PR #394, opened 2019-06-12
fixes #370 SumFormulaDescription should be updated

### LINKED ISSUE #370: SumFormulaDescription should be updated
see https://github.com/Open-Systems-Pharmacology/MoBi/pull/369/files

Yuri05:


src/MoBi.Assets/AppConstants.cs
```diff
@@ -1599,7 +1599,7 @@ public static class Captions
          //TODO MOVE TO CORE
          public static readonly string UserDefined = "User Defined";
 
-         public static string SumFormulaDescription(string iterationPattern) => $"This is a description of Sum formula using {iterationPattern}: TODO. This is a description of Sum formula : TODO";
+         public static string SumFormulaDescription(string iterationPattern) => $"Sum formula is defined as R1*…*Rm*∑{iterationPattern}*Q1_#i*…Qn_#i where R1…Rm (m>=0) are the quantities of an independent object (as absolute path or relative path); {iterationPattern} is a control variable (parameter, molecule amount, … defined by certain conditions); and Q1_#i…Qn_#i (n>=0) are the quantities that are obtained from a path relative to {iterationPattern}";
```
Yuri05: really hard to describe the formula behaviour in a short tooltip. we should describe it in the manual with more details and maybe some examples

msevestre: yeah. I have no idea what this description means :)
Can we describe it with an example instead? At least in the manual
Also this is not shown as tool tip but as description

Yuri05: In the manual for sure. In the UI, having a very limited space, hard to describe.
OK I will merge then

PR #397, opened 2019-06-13
Fixes #396 issue with curve

### LINKED ISSUE #396: Bug with remove columns called differently than in PKSim

msevestre:


src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -232,10 +232,10 @@ public void Show(CurveChart chart, IReadOnlyList<DataRepository> dataRepositorie
          try
          {
             clearNoCurvesHint();
+            InitializeAnalysis(chart);
             //do not validate template when showing a chart as the chart might well be without curves when initialized for the first time.
             var currentTemplate = defaultTemplate ?? _chartTemplatingTask.TemplateFrom(chart, validateTemplate: false);
             replaceSimulationRepositories(dataRepositories);
-            InitializeAnalysis(chart);
```
msevestre: Some data repositories where removed from the editor before the chart was initialized. I'lll add a check in core to make sure that the order does not matter. But this should be good like that anyways
https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/576

PR #462, opened 2020-03-26
Fixes #452 cursor amount field

### LINKED ISSUE #452: Placing the cursor into the "Amount"-field overwrites the formula
In the Simulation tree view, simply placing the curso into the field "Amount" of a molecule "overrides" the formula.

image

Changin e.g. the expression value in the parameters-field has then no effect on the amount of the molecule.

msevestre:


src/MoBi.UI/Views/ValueEdit.cs
```diff
@@ -51,20 +52,27 @@ public override void InitializeBinding()
          tbValue.EnterMoveNextControl = true;
       }
 
-      public override bool HasError
+      private void onValueUpdating(ValueEditDTO dto, PropertyValueSetEventArgs<double> e)
       {
-         get { return _screenBinder.HasError; }
+         // Some artifact with this control when values have a lot of digits could lead to an event being raised even though nothing has changed.
+         // This should be fixed in OSPSuite.DataBinding but we'll be done at another time
+         if (ValueComparer.AreValuesEqual(e.OldValue, e.NewValue))
```
msevestre: @Yuri05 This is an issue with the DataBinding component. Only fixes this here for now and I created an issue in DataBinding to fix it later on

src/MoBi.UI/Views/ValueEdit.cs
```diff
@@ -51,20 +52,27 @@ public override void InitializeBinding()
...
+            return;
+
+         ValueChanged(dto, e.NewValue);
```
Yuri05: is line 62 the replacement for deleted line `OnPropertyChanged` event in `ValueEditDTO.setValue` ?

msevestre: No the OnPropertyChanged was sent twice in that case (see comment)

Yuri05: ok

msevestre: The problem is that the value in the object is something like 5.123456789123456798E-5
The control has 5.12345678912345E-5 

The function then believes that something has changed. I am surprised that we did not find this bug before. I guess it does not happen often that we have so many values after the digit

msevestre: see https://github.com/Open-Systems-Pharmacology/OSPSuite.DataBinding/issues/10

Yuri05: yep, so many digits do not make much sense, it's below double precision

PR #463, opened 2020-03-26
Fixes #427 wrong selection

### LINKED ISSUE #427: Spatial Structures Parameters tab: wrong selection when "Show advanced parameters" is turned off
If the checkbox "Show advanced parameters" in the Parameters-tab is unchecked, it is sometimes impossible to select a non-advanced parameter.

E.g.:
Open any PK-Sim project in MoBi, navigate to "Organism|Bone|Plasma", check "Show advanced parameters":
image

Now uncheck "Show advanced parameters" and try to select "Volume". The parameter "FcRn concentration in container" is still selected:

image

msevestre:


src/MoBi.Presentation/Presenter/EditParameterPresenter.cs
```diff
@@ -144,16 +143,6 @@ IEditValueOriginPresenter editValueOriginPresenter
          _editValueOriginPresenter.ShowCaption = false;
       }
 
-      public void SetPropertyValueFromView<TProp>(string propertyName, TProp newValue, TProp oldValue)
```
msevestre: not used anymore

src/MoBi.Presentation/Presenter/EditParametersInContainerPresenter.cs
```diff
@@ -162,8 +161,10 @@ private ParameterDTO dtoFor(IParameter parameter)
 
       private void showParameters()
       {
-         var parametersToShowDTO = _allParametersDTO.Where(shouldShowParameter);
+         var parametersToShowDTO = _allParametersDTO.Where(shouldShowParameter).ToList();
          _view.BindTo(parametersToShowDTO);
+         setupEditPresenter(parametersToShowDTO.FirstOrDefault()?.Parameter);
```
msevestre: I always select the first parameter. In theory we could try to see if the selected parameter is still in the list so we do not reselect but the ref is not saved so I don't want to open a big can of worms

PR #528, opened 2021-02-25
Plugging in "Importer Branch" Core Package to MoBi

georgeDaskalakis:


msevestre:
@georgeDaskalakis Going to merge eventhough the build is failing as I am assuming that this will be dealt with later on.
Thanks :)

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -257,7 +257,7 @@ private IEnumerable<ColumnInfo> createColumnInfos()
             NullValuesHandling = NullValuesHandlingType.DeleteRow,
          };
 
-         timeColumn.DimensionInfos.Add(new DimensionInfo {Dimension = timeDimension, IsMainDimension = true});
+         timeColumn.DimensionInfos.Add(new DimensionInfo.DimensionInfo {Dimension = timeDimension, IsMainDimension = true});
```
msevestre, line 260: @georgeDaskalakis  Why do we have TWO DimensionInfo? Isn't it weird? I'll comment it just to have a discussion

msevestre (approved):

PR #535, opened 2021-03-12
487 tag for event group

msevestre:


msevestre:
Need to add some tests. 
Typical implementation where I thought I would be done in 30min. But the code was duplicated everywhere! So I created a TagPresenter abstraction and I can now use it everywhere. 
I also removed some DTO that where nonsensical

src/MoBi.Presentation/Presenter/TagsPresenter.cs
```diff
@@ -0,0 +1,64 @@
...
+   //AppConstants.Captions.ContainerTags.FormatForLabel
+
+   public interface ITagsPresenter : IEditPresenter<IEntity>, IPresenter<ITagsView>
+   {
+      void AddNewTag();
+      void RemoveTag(TagDTO tagDTO);
+      IBuildingBlock BuildingBlock { get; set; }
+   }
+
+   public class TagsPresenter : AbstractEditPresenter<ITagsView, ITagsPresenter, IEntity>, ITagsPresenter
```
msevestre, line 23: Simple  implementation. Very easy to reason about, test in isolation etc.
We should try to do the same for all presenters function we will create in the future.

src/MoBi.Presentation/Tasks/EntityTask.cs
```diff
@@ -31,20 +30,16 @@ public EntityTask(IMoBiContext context, ITagVisitor tagVisitor, IDialogCreator d
          _dialogCreator = dialogCreator;
       }
 
-      public IMoBiCommand AddNewTagTo(IEntity entity, ITaggedEntityDTO dto, IBuildingBlock buildingBlock)
+      public IMoBiCommand AddNewTagTo(IEntity entity, IBuildingBlock buildingBlock)
       {
-         var tag = _dialogCreator.AskForInput("New Tag for Container", "New Tag", string.Empty, Enumerable.Empty<string>(), getUsedTags());
+         var tag = _dialogCreator.AskForInput("New Tag for Container", "New Tag", string.Empty, entity.Tags.Select(x => x.Value), getUsedTags());
```
msevestre, line 35: This was in fact a bug. One could enter the same tag twice!

msevestre (commented):
Need to add some tests but the function is here
Let's make sure that we do not hesitate to refactor stuff when we need to in the future otherwise MoBi will still be very hard to extend

Yuri05 (approved):

PR #537, opened 2021-03-12
Fixes #525 enhance list of observers

### LINKED ISSUE #525: Extend the list of "Standard PK-Sim observers" used to hide warnings
When the setting “Show warnings from PK-Sim standard observers” is disabled: warnings from some PK-Sim observers are hidden (otherwise they would appear every time when creating a new simulation in MoBi. For the standard PK-Sim observers those warnings are irrelevant and should not be shown)
grafik

However with the latest model structure modifications some warnings are now shown, which should be hidden:
grafik

The observers above should be added to the MoBi Ignore-List:
https://github.com/Open-Systems-Pharmacology/MoBi/blob/02a5e9290aaf871504310837ed39b9feb53fc6e4/src/MoBi.Assets/AppConstants.cs#L103-L109

msevestre:


msevestre (commented):
all static and one dynamic

Yuri05 (commented):
approved

PR #538, opened 2021-03-12
Fixes #485 manual input event

### LINKED ISSUE #485: Manual input for event assignment path
Currently, the only way to change an event assignment's entity path to use the keyword "MOLECULE" (as in the particle dissolution model, shown below) is by manually editing the pkml file. Perhaps, it would be more user-friendly to allow for manual input of the entity path?

Capture

msevestre:


src/MoBi.Presentation/Presenter/EditEventBuilderPresenter.cs
```diff
@@ -19,20 +19,22 @@
 using OSPSuite.Core.Services;
```
msevestre, line 19: This whole code is a mess. REtrieving object from context etc... 
I don't want to touch it too much because we'll probably revisit event stuff eventually...but wow

Yuri05 (approved):

PR #539, opened 2021-03-15
fixes #507 drag and drop

### LINKED ISSUE #507: Drag and drop of parameter to the "references area" does not work
Mobi, OPS version: 9.1:
Attached model (file "Model.zip"): In the spatial structure at the organism level a new parameter "AAA" is added. 
At the molecule "Rez" at the molecules building block a parameter BBB was added. In order to define the equation for parameter BBB I wanted to drag and drop the parameter AAA to the references area. The drop does not work, see the black "forbidden" sign in the picture attached. An analogous drop worked in a previous version (now only adding the reference manually works). 

[Model.zip](https://github.com/Open-Systems-Pharmacology/MoBi/files/5167270/Model.zip)
DragDrop

msevestre:


msevestre:
@Yuri05 Going to merge this. Please review

PR #540, opened 2021-03-15
Fixes #459 import wrong dimension

### LINKED ISSUE #459: Import of simulation parameters: Wrong dimension
When importing simulation parameters from an Excel-file via right-clicik on the PSV-BB -> "Import...", entries with the unit "1/min" are imported as dimension "Becquerel".  Most of the time, it is "Inversed time".

A solution would be to specify an extra column "Dimension" in the import file.

msevestre:


src/MoBi.Presentation/Mappers/AbstractDataTableProviderToQuantityDTOMapper.cs
```diff
@@ -36,16 +36,32 @@ protected AbstractDataTableProviderToQuantityDTOMapper(IMoBiDimensionFactory dim
...
+
+         if (table.Columns.Count <= dimensionColumn)
+            return dimension;
+
+         //now we may have a dimension specified
+         var dimensionName = row[dimensionColumn].ToString();
+         if (string.IsNullOrEmpty(dimensionName))
+            return dimension;
+
+         if (!_dimensionFactory.Has(dimensionName))
```
msevestre, line 60: @Yuri05 Idea is pretty simple.. Same as before. But if a dimension column was specified, we use this instead.
If the dimension does not match the unit, the importer will throw an error as well so all is covered

msevestre (commented):
@Yuri05 Going to merge this . Please review

PR #541, opened 2021-03-15
Fixes #486 display of formula name

### LINKED ISSUE #486: Suggestion: Display of actual formula name instead of formula
There are two places in the Event building blocks where the formula itself is displayed instead of the formula name: 1) in the application molecule builder and 2) in event assignments. 
This can make it difficult to track down the formula under the Formulas tab. Also, one cannot distinguish between formulas with the same equation but different references (e.g. ParticleBinDissolvedDrugMassBinDecrease formulas in the particle dissolution model). I would think that displaying the formula name would be more consistent with how other formulas are handled and avoid these issues.

InkedCapture2_LI
InkedCapture3_LI
InkedCapture_LI
InkedCapture4_LI

msevestre:


msevestre:
@Yuri05 Trivial change. We can update if we need easily

PR #543, opened 2021-03-16
Fixes #530 illegal chars

### LINKED ISSUE #530: Illegal characters in parameter/container names
At the moment some illegal characters can be included into parameter names, which might lead to problems, e.g.
* Double quotes
* Pipes
* ... (anything else?)

Naming check should be extended, so that those characters will be forbidden.

grafik

grafik

msevestre:


src/MoBi.Presentation/DTO/ObjectBaseDTO.cs
```diff
@@ -90,9 +99,15 @@ private static class AllRules
             .WithRule((dto, name) => dto.IsNameUnique(name))
             .WithError(AppConstants.Validation.NameAlreadyUsed);
 
+         private static IBusinessRule nameDoesNotContainIllegalCharacters { get; } = CreateRule.For<IObjectBaseDTO>()
```
msevestre, line 102: @Yuri05 Copied the rule from Core. I wanted to use the base object, but there are some stuff in this ObjectBaseDTO that I don't quite understand so I'd rather not mess stuff around too much

src/MoBi.Presentation/Tasks/Interaction/ObjectBaseTask.cs
```diff
@@ -40,23 +39,27 @@ internal class ObjectBaseTask : IObjectBaseTask
...
       }
 
-      public IMoBiCommand Rename(IObjectBase objectBase, IEnumerable<string> allreadyNames, IBuildingBlock buildingBlock)
+      public IMoBiCommand Rename(IObjectBase objectBase, IEnumerable<string> alreadyUsedNames, IBuildingBlock buildingBlock)
       {
-         var unallowedNames = new List<string>(allreadyNames);
+         var unallowedNames = new List<string>(alreadyUsedNames);
          unallowedNames.AddRange(AppConstants.UnallowedNames);
          var objectName = _objectTypeResolver.TypeFor(objectBase);
-         string newName = _dialogCreator.AskForInput(AppConstants.Dialog.AskForNewName(objectBase.Name), AppConstants.Captions.NewName, objectBase.Name, unallowedNames);
```
msevestre, line 59: Here the DTO was bypassed altogether. Absolutely no name checked performed!

msevestre:
@Yuri05 Merging this. All standard illegal chars are forbidden now

 ":", "*", "?", "<", ">", "|", "{", "}", "\""

PR #555, opened 2021-04-28
Fixes #554 file is readonly

### LINKED ISSUE #554: File is readonly when sent from PKSim
/cc @Christoph27, @Yuri05 

Going to check this out

msevestre:


src/MoBi.Core/Serialization/ORM/ContextPersistor.cs
```diff
@@ -114,7 +114,7 @@ public void Save(IMoBiContext context)
       private void verifyFileNotReadOnly(string projectFilePath)
       {
          var fileInfo = new FileInfo(projectFilePath);
-         if (!fileInfo.IsReadOnly)
+         if (!fileInfo.Exists || !fileInfo.IsReadOnly)
```
msevestre, line 117: @Yuri05 A file that does not exist IS readonly. WTF

Yuri05: I guess it's more a philosopical question :)

PR #567, opened 2021-05-26
fixes #566

### LINKED ISSUE #566: manual entry of molecule in importer not working
The manual entry of the molecule is not possible with the new version of core.

georgeDaskalakis:


src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -495,7 +496,7 @@ private void addInfoToCategory(DimensionInfo.MetaDataCategory metaDataCategory,
             metaDataCategory.ListOfImages.Add(container.Name, icon.IconName);
       }
 
-      private DimensionInfo.MetaDataCategory createMetaDataCategory<T>(string categoryName, bool isMandatory = false, bool isListOfValuesFixed = false, Action<DimensionInfo.MetaDataCategory> fixedValuesRetriever = null, string description = null)
+      private DimensionInfo.MetaDataCategory createMetaDataCategory<T>(string categoryName, bool isMandatory = false, bool isListOfValuesFixed = false, Action<DimensionInfo.MetaDataCategory> fixedValuesRetriever = null, string description = null, bool allowsManualInput = false)
```
msevestre, line 499: Is there a reasoin why we still have the name space DimensionInfo?

msevestre (changes requested):
one question. I think this should be removed (namespace DimensionInfo)

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
msevestre: well at least in core :)

PR #576, opened 2021-06-25
Fixes #537 project conversion

msevestre:


src/MoBi.Core/Serialization/Converter/v10/Converter90To100.cs
```diff
@@ -0,0 +1,84 @@
...
+         Visit(simulationTransfer?.Simulation);
+      }
+
+      private void convertFormula(IFormula formula)
+      {
+         var explicitFormula = formula as ExplicitFormula;
+         if (explicitFormula == null)
+            return;
+
+         if (!explicitFormula.NameIsOneOf(
```
msevestre, line 63: @PavelBal @Yuri05 These were the only one that I could find in v9 with a division

msevestre: @PavelBal I'd love if you could check it out and let me know if anything was missed?

msevestre:
@PavelBal @Yuri05 Going to merge so that @PavelBal cna try the nightly. I'll leave the issue open however

PR #577, opened 2021-06-29
573 update core

msevestre:


src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -442,7 +442,7 @@ private void addNamingPatterns(DimensionInfo.DataImporterSettings dataImporterSe
             Constants.ObservedData.GENDER,
             Constants.ObservedData.DOSE,
             Constants.ObservedData.ROUTE,
-            Constants.ObservedData.PATIENT_ID
+            Constants.ObservedData.SUBJECT_ID
```
msevestre, line 445: @abdelr @georgeDaskalakis Shouldn't that come from Core ? At least the default that we can override?

Issue #287, opened 2018-05-31
SBML Import : Wrong import of model equations


msevestre:
MoBi-Import of [BIOMD0000000528](http://www.ebi.ac.uk/biomodels-main/BIOMD0000000528) and [BIOMD0000000489](http://www.ebi.ac.uk/biomodels-main/BIOMD0000000489) Biomodels fails due to wrong import of model equations

[SBML_Import.zip](https://github.com/Open-Systems-Pharmacology/PK-Sim/files/1495512/SBML_Import.zip)

Steps to reproduce:
1. Create new MoBi project
2. Import one of the SBML models attached
3. Create empty Observers building block
4. Create empty Events building block
5. Create default simulation settings
6. Try to create a simulation - it fails with errors

(MoBi projects created in step 5 also attached for both models)

see original in https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/472

Yuri05:
One problem is: current SBML import does not handle `<functionDefinition>` properly.
E.g. in the problematic example **BIOMD0000000489** from above, kinetic law of the reaction **"reaction_2"** makes use of the function definition **"function_1"** . The part of the equation corresponding to **"function_1"** is imported just as an empty string...

```
<reaction id="reaction_2" metaid="COPASI56" name="MYD882) IRAK1 and TRAF6 Phosphorylation" reversible="true">
  ...
  <kineticLaw>
    <math xmlns="http://www.w3.org/1998/Math/MathML">          
	  <apply>
	    <times/>
		<ci> compartment_1 </ci>
		<apply>
		  <ci> function_1 </ci>
		  <ci> k1 </ci>
		  <ci> species_5 </ci>
		  <ci> species_9 </ci>
		  <ci> species_10 </ci>
		  <ci> k2 </ci>
		  <ci> species_11 </ci>
		</apply>
	  </apply>
	</math>
  </kineticLaw>
... 
<functionDefinition id="function_1" metaid="COPASI119" name="Modified Mass Action">
```

msevestre:
Isn't it what the sbml lib is supposed to do?

Yuri05:
[...]

PR #590, opened 2021-07-30
Updating dlls

### LINKED ISSUE #589: MoBi should be able to import level 3 sbml files

abdelr:
WIP #589 (loading level 3 sbml files)


msevestre:
hum..... Are you sure swapping the dll is enough to FIX the issue? There is a ton load of mapping going on and if new constructs are necessary, new code will be required

abdelr:
This is why it is on draft mode ;). We are testing ....

msevestre:
sorry did not see. The commit message said "Fixes #589" that's why:)

Yuri05:
@abdelr Which libsml version is it now? Seems to be 5.18. 
Also I would add an integration test with SBML level 3 model.

abdelr:
> @abdelr Which libsml version is it now? Seems to be 5.18.

5.18.0 indeed.

> Also I would add an integration test with SBML level 3 model.
>

Sure, I could already add a couple of integration tests with simple level 2 and level 3 already working tests. But I would say it is a work in progress draft PR for now. We are planning to test more thoroughly and create issues as they are detected. It was not possible to lead level 3 files at all. It was crashing. Now, it does not crash but I am not sure how it handles all the information. We will be testing and fixing issues that could come down the road.
 
> And: does it fix #287 as well?

It crashes when loading the level 2 version 4 file. This might be a good point to try to start solving already.

abdelr:
> sorry did not see. The commit message said "Fixes #589" that's why:)

I added the comment to link the issue but it is not solving the issue yet. I actually thought on not creating the PR but I went for creating the PR on draft mode and work on it so there is somewhere on Github in case we need to cooperate.

msevestre:
That's fine. My suggestion would be to say WIP instead of Fixes (I was on my mail client and did not see the draft mode)

abdelr:
I just added handling invalid conversion. Still, many situations are not handled so I would say this is far from finished (the old importer too). I suggest we start testing it with real files and tackle the issues as they are discovered. I tried some online files and that are not supported. I will try at least that it does not crash but still, we need to describe what information is been discarded that we still need.

src/MoBi.Engine/Sbml/SBMLTask.cs
```diff
@@ -104,9 +104,14 @@ public Model GetModel(string filename)
...
             throw new MoBiException(SBMLConstants.ModelNotRead(sbmlDoc.getErrorLog().ToString()));
          }
-
-         convertSBML(sbmlDoc);
-
+         try
+         {
+            convertSBML(sbmlDoc);
+         }
+         catch (Exception _)
```
msevestre, line 111: Are we sure this is always the reason for an exception? Cannot we be more explicit here? What if the code in `convertSBML` (our code) crashes?

msevestre (changes requested):
Asking to see if we can not catch all exception but a more specific one

abdelr:
This is off dated... The comment from @msevestre 

> Asking to see if we can not catch all exception but a more specific one

is fixed on #599

PR #593, opened 2021-08-12
592 parse user defined functions

### LINKED ISSUE #592: Parse user defined functions when loading reactions from SBML files
Derived from #287 
When loading SBML files, reading reactions should parse user defined functions

abdelr:
Fixes #592 (parsing user defined functions when loading reactions from SBML file)


abdelr:
This problem is not new with updating the dlls to handle level 3 files. It is a known old problem actually. Since this is scoped to the SBML importer, we might consider including this into V10 since the risk is null to me and it actually solves something that has been wrong for some time. It is still not complete. In addition to this, there are problems with the paths so still, it is not possible to create the simulations but at least we are getting closer ;)

msevestre (commented):
I am fine with it
@Yuri05 ?

Yuri05 (approved):

Yuri05:
looks good for me

PR #595, opened 2021-08-17
594 create aliases when loading reactions from sbml files

### LINKED ISSUE #594: Create aliases when loading reactions from SBML files
Derived from #287
When loading SBML files, reading reactions should create all aliases

abdelr:
Fixes #594 (Create aliases when loading reactions from SBML files)

### DIFF
--- a/src/MoBi.Engine/Sbml/ASTHandler.cs
+++ b/src/MoBi.Engine/Sbml/ASTHandler.cs
@@ -56,6 +56,7 @@ namespace MoBi.Engine.Sbml
          try
          {
             _objectPaths.Clear();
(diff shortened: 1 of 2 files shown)


abdelr:
Following on #592, this should also be included on V10. Actually, with this, it is now possible to create simulations after importing an SBML file so I would recommend to close #287 after merging this PR. I tried with the sample xml file on tests and it does not crash anymore when creating the simulation but when running it, there are problems. I would say those problems are not related to the import ignoring some important data though. I am more biased to think I configured something wrong on the rest of the project so I would recommend a more experience users to try the feature ;).

Yuri05 (approved):

PR #597, opened 2021-08-31
596 units handling when importing sbml file

### LINKED ISSUE #596: SBML import: dimensions/units that are not supported
When importing a SBLM model, the dimensions/units must be mapped to the entries supported in OSPS. As there is no strict list of allowed units for SBLM, automatic mapping is not always possible. As a solution:

On load, show the list of all units in the SBML, and the user can select the mapping to the OSP units and a scaling factor.

abdelr:
Fixes #596 (SBML Importer does not create new dimensions, instead converts SBML units into MoBi units)


abdelr:
@msevestre @PavelBal @Yuri05 with this last PR, I was able to successfully run simulation imported from SBML (for the first time) so I would like to evaluate to include this on V10 as discussed before. Notice though that I did modify somethings from our last talk. The importer is currently not prompting the user to check if the units are correctly mapped since with the enum for the kind field we have a well defined set of units and those units are not translated to MoBi units successfully. I guess we will discover some bugs as the users start using it but at least, it is ready for use now I think. Please, let me know your thoughts on the implementation and I will gladly address your feedback.

msevestre:
Gladly....lol;-)

src/MoBi.Engine/Sbml/ParameterImporter.cs
```diff
@@ -7,53 +7,62 @@
...
-        ///     Creates a MoBi Parameter from a given SBML Parameter.
-        /// </summary>
-        public IFormulaUsable CreateParameter(Parameter sbmlParameter)
-        {
-            var value = 0.0;
-            if (sbmlParameter.isSetValue()) value = sbmlParameter.getValue();
-            IFormula formula = ObjectBaseFactory.Create<ConstantFormula>().WithValue(value);
+      public void SetUnitDefinitionImporter(UnitDefinitionImporter unitDefinitionImporter)
+      {
+         _unitDefinitionImporter = unitDefinitionImporter;
```
msevestre, line 32: Why is the service not injected in constructor?

abdelr: There were no clear interfaces for that. I refactor some parts of the code and we can inject the importers but in general the SBML Importer old code is not following some principles of inversion of control. I think we should not address this now but maybe later we should revisit the topic.

PR #599, opened 2021-09-02
598 create default building blocks on sbml import

### LINKED ISSUE #598: SBML importer: create missing building block
When importing an SBLM model, not all building blocks necessary for model creation are created (e.g., simulation settings). In this case, empty or default BB of the type should be created, so the user can directly create a simulation.

abdelr:
Fixes #598


src/MoBi.Engine/Sbml/EventImporter.cs
```diff
@@ -42,11 +42,14 @@ protected override void Import(Model model)
...
          CreateEGBandEGBB();
-
-         for (long i = 0; i < model.getNumEvents(); i++)
+         long i;
+         for (i = 0; i < model.getNumEvents(); i++)
          {
             CreateEvent(model.getEvent(i));
          }
+         if (i == 0)
+            CreateEmptyEvent();
```
Yuri05, line 51: I think something like below would be more clear
```
if (model.getNumEvents() == 0)
    CreateEmptyEvent();
else
{
    for (var i = 0; i < model.getNumEvents(); i++)
        CreateEvent(model.getEvent(i));
}
```

src/MoBi.Engine/Sbml/EventImporter.cs
```diff
@@ -90,6 +93,15 @@ internal void CreateEvent(Event sbmlEvent)
...
       }
 
+      private void CreateEmptyEvent()
+      {
+         EventGroupBuilder.Add(
+            new EventBuilder()
+               .WithId(SBMLConstants.SBML_EVENTBUILDER)
+               .WithName(SBMLConstants.SBML_DEFAULTEVENTNAME)
+         );
+      }
```
Yuri05, line 103: this looks suspicios for me. Why not creating the new event builder via `Context.Create()` ?
Especially if we import >1 sbml simulations - it might create several event builders with the same Id (or even crash?).
@msevestre Maybe you can comment on this.

src/MoBi.Engine/Sbml/EventImporter.cs
abdelr: Thanks

src/MoBi.Engine/Sbml/EventImporter.cs
abdelr: @msevestre and @Yuri05, I am using the same code style. I would say if we need to change this, there are several places where we need to do the same. Do notice that at this point, the object is not yet added to the project. It will later be added when the method AddToProject is invoked (which also updates the history). Could this be the cause?

PavelBal: `CreateEmptyEvent()` is only called if no Events-BB is created during import, so there should be no issues about IDs.

Yuri05: If I import >1 SBML simulations into the same project - no issues as well?

PavelBal: Not possible - "import sbml" always starts an new project. We might want to change this later but now it is like selecting "Open simulation"

src/MoBi.Engine/EngineRegister.cs
```diff
@@ -11,15 +12,34 @@ public override void RegisterInContainer(IContainer container)
...
             x.AssemblyContainingType<EngineRegister>();
+            x.WithConvention(new OSPSuiteRegistrationConvention(registerConcreteType: true));
             x.ExcludeType<UnitDefinitionImporter>();
+            x.ExcludeType<CompartmentImporter>();
+            x.ExcludeType<SpeciesImporter>();
+            x.ExcludeType<ParameterImporter>();
             x.ExcludeType<FunctionDefinitionImporter>();
-            x.WithConvention(new OSPSuiteRegistrationConvention(registerConcreteType: true));
+            x.ExcludeType<ReactionImporter>();
+            x.ExcludeType<AssignmentImporter>();
```
msevestre, line 22: We do not need this. See current change in develop where we register by convention

src/MoBi.Engine/Sbml/SBMLTask.cs
```diff
@@ -105,14 +105,7 @@ public Model GetModel(string filename)
...
          }
-         try
-         {
-            convertSBML(sbmlDoc);
-         }
-         catch (Exception _)
-         {
-            throw new MoBiException(SBMLConstants.NoValidConversionToLevel3Version2);
-         }
+         convertSBML(sbmlDoc);
```
msevestre, line 108: I don't get it. Now you do not catch anything at all? Don't you want to catch when you have the wrong version

msevestre (changes requested):
Registration is wrong.... and no catch anymore?

tests/MoBi.Tests/Core/SBML/EventImporterSpecs.cs
```diff
@@ -24,6 +24,22 @@ protected override void Context()
...
+      {
+         var sbmlModel = new Model(3, 1);
+
+         sut.DoImport(sbmlModel, new MoBiProject(), A.Fake<SBMLInformation>(), new MoBiMacroCommand());
+      }
+
+      [Observation]
+      public void LoadsDefaultIfNoEvent()
+      {
+         ObjectBaseExtensions.AllNames(sut.EventGroupBuilder.Children).ShouldNotBeEmpty();
```
msevestre, line 39: use `sut.EventGroupBuilder.Children.AllNames()`

src/MoBi.Engine/EngineRegister.cs
abdelr: Right, only ASTHandler was missing.

src/MoBi.Engine/Sbml/SBMLTask.cs
abdelr: It does throw the exception. convertSBML is actually checking and throwing the exception. Here we are just catching everything and translating to no valid conversion when actually this is already done in line 137 where the return from setLevelAndVersion is been checked.

src/MoBi.Engine/EngineRegister.cs
```diff
@@ -11,13 +11,13 @@ public override void RegisterInContainer(IContainer container)
...
             x.AssemblyContainingType<EngineRegister>();
+            x.WithConvention(new OSPSuiteRegistrationConvention(registerConcreteType: true));
             x.ExcludeType<UnitDefinitionImporter>();
             x.ExcludeType<FunctionDefinitionImporter>();
-            x.WithConvention(new OSPSuiteRegistrationConvention(registerConcreteType: true));
          });
 
-         //No idea why this is required explicitly. Maybe because the class only has yield operators?
          container.Register<SBMLImporterRepository, SBMLImporterRepository>();
+         container.Register<ASTHandler, ASTHandler>();
```
msevestre, line 20: hum...no idea why this is required also... oh well

msevestre (approved):

Issue #521, opened 2020-11-24
Clone simulation


PavelBal:
Definitely would like to see this feature in MoBi.

PavelBal:
https://github.com/Open-Systems-Pharmacology/MoBi/tree/521_clone_simulation

PR #602, opened 2021-09-09
Cloning simulations

abdelr:
WIP #521 (Clone simulations)

### DIFF
--- a/src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
+++ b/src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
@@ -8,6 +8,12 @@ using MoBi.Presentation.Tasks.Interaction;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Presentation.Presenters;
+using System;
+using OSPSuite.Core.Serialization;
+using IContainer = OSPSuite.Utility.Container.IContainer;
+using OSPSuite.Core.Domain.UnitSystem;
+using OSPSuite.Core.Domain.Services;
+using MoBi.Core.Serialization.Xml.Services;
 
 namespace MoBi.Presentation.Tasks.Edit
 {
@@ -15,6 +21,8 @@ namespace MoBi.Presentation.Tasks.Edit
    {
       void Edit(T objectToEdit);
       void Save(T entityToSerialize);
+
+      T Clone(T entityToClone);
 
       /// <summary>
       ///    Gets the forbidden names for the given object, from the local list the objects name is removed.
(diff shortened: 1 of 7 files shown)


abdelr:
@PavelBal @msevestre and @Yuri05 I have marked this as a draft because I am not sure about the whole specifications. Currently, it does clone the simulation based on the entity to clone but maybe we would like to make a deeper or shallower copy. I am struggling a bit here since we have not worked too much into MoBi direction yet so please, take a look and let me know what do you think is still missing or completely wrong. Also, the tests are for sure, not complete. Here I would like to hear a bit more about what are the expectations after cloning a simulation so I could turn those requirements into tests ;).

msevestre:
Let's postpone this until after the release. Cloning is far from trivial because of all the references to building blocks going on. The implementation might be working but it would require a lot of testing. I'll look at this later

PavelBal:
Yes, definitely not for V10. We'll continue working on it in the branch, it is also a good opportunity to learn more about how simulations are set up in MoBi.

abdelr:
Definitely not for V10. I think we all agreed V10 should be sealed now (unless some bug fixing). I also have the feeling this is going to take some time and testing and this is why I opened the discussion ;)

src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
```diff
@@ -42,22 +50,59 @@ public abstract class EditTaskFor<T> : IEditTaskFor<T> where T : class, IObjectB
...
+         _objectBaseFactory = objectBaseFactory;
+         _cloneManagerForModel = cloneManagerForModel;
       }
 
       public virtual void Edit(T objectToEdit)
       {
          _context.PublishEvent(new EntitySelectedEvent(objectToEdit, this));
       }
 
+      protected TEntity Clone<TEntity>(TEntity entityToClone) where TEntity : class
```
msevestre, line 85: @abdelr This is not correct. A clone method like this one will keep all internal ids resulting in different object with same id. I know this is a draft but you cannot rely on this method

msevestre: I open the PR by mistake. Let me know when you want me to review

georgeDaskalakis:
@msevestre is this PR up to date? should we actually delet or is there something to finalize?

msevestre:
I know nothing about this PR. Maybe @PavelBal knows what the status is?

PavelBal:
Abdel started to work on implementation of simulation cloning in MoBi, but it was not finished. Cloning simulations is an important feature, however, it might get obsolete with OSMOSES. If it does not bother you, we could let this PR live for a while until we either a) solve this in OSMOSES or b) decide to implement it separately.

PavelBal:
Let's close the PR, but keep the branch for now.

PR #612, opened 2021-09-21
Assignment Importer should use absolute paths so it can be used so set values to other building blocks

### LINKED ISSUE #607: Formulas are parsed on initial parameter values
Load the model on https://www.ebi.ac.uk/biomodels/MODEL1904090001#Files and try to create a simulation. The action results in errors relating the spatial structure.

abdelr:
Fixes #607  Assignment Importer should use absolute paths so it can be used so set values to other building blocks

### DIFF
--- a/tests/MoBi.Tests/Core/SBML/AssignmentImporterSpecs.cs
+++ b/tests/MoBi.Tests/Core/SBML/AssignmentImporterSpecs.cs
@@ -78,7 +78,11 @@ namespace MoBi.Core.SBML
       {
          var psvbb = _moBiProject.ParametersStartValueBlockCollection.FirstOrDefault();
          psvbb.ShouldNotBeNull();
-         psvbb.Where(psv => psv.Formula != null).Each(psv => psv.Formula.ObjectPaths.Each((p => p.Resolve<IFormulaUsable>(psv).ShouldNotBeNull())));
+         psvbb
+            .Where(psv => psv.Formula != null)                          //if it contains a formula
+            .Each(psv => psv.Formula.ObjectPaths.Each(                  //then all its objectPaths
+               p => p.Resolve<IFormulaUsable>(psv).ShouldNotBeNull())   //should be resolvable from where they are used
+            );
       }
 
    }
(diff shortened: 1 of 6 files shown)


tests/MoBi.Tests/Core/SBML/ContextForSBMLIntegration.cs
```diff
@@ -23,7 +23,7 @@ protected override void Context()
          var context = IoC.Resolve<IMoBiContext>();
          context.NewProject();
          _moBiProject = context.CurrentProject;
-         context.LoadFrom(_moBiProject);
+         //context.LoadFrom(_moBiProject);
```
msevestre, line 26: why this change

tests/MoBi.Tests/Core/SBML/AssignmentImporterSpecs.cs
```diff
@@ -63,4 +63,23 @@ public void Compartment_InitialAssignmentCreationTest()
...
+         _fileName = Helper.TestFileFullPath("tiny_example_12.xml");
+      }
+
+
+      [Observation]
+      public void FormulaParameter_InitialAssignmentCreationTest()
+      {
+         var psvbb = _moBiProject.ParametersStartValueBlockCollection.FirstOrDefault();
+         psvbb.ShouldNotBeNull();
+         psvbb.Where(psv => psv.Formula != null).Each(psv => psv.Formula.ObjectPaths.Each((p => p.Resolve<IFormulaUsable>(psv).ShouldNotBeNull())));
```
msevestre, line 81: even here.... it makes sense to simplify some of  the query to be more readable.

tests/MoBi.Tests/Core/SBML/ContextForSBMLIntegration.cs
abdelr: I am not sure why this line is needed but I have just restored it. I was experimenting a bit and forgot to restore it on my last commit.

msevestre (commented):
it works. But I am personally in favor or defining a variable
var allObjectPaths = psvbb.Where(x=> x.Formula != null).SelectMany(x=>x.objectPath);

and then checking. Then there is not need for my (clearly obsolete brain) to analyse for loop within for loop

PR #613, opened 2021-09-22
607 formulas are parsed on initial parameter values

### LINKED ISSUE #601: Nighlty build failing
This has been going on for a while
https://ci.appveyor.com/project/open-systems-pharmacology-ci/mobi/history
Due to SBML stuff. Please fix asap (it might already be fixed with PR merged this morning)
@abdelr @georgeDaskalakis

abdelr:
Fixes #601 (Removing obsolete test)


abdelr:
Hi @msevestre, we discussed yesterday about this test. It actually was obsolete. It was moved to SBMLTaskSpecs.cs since it is concerning the whole integration of packages. This is also why was a bit cryptic (it was not finished). So trying to explain it with comments was not a good idea after all. I believe, the test in SBMLTaskSpecs makes complete sense and is much more clear.

PR #615, opened 2021-09-28
Setting volume from size

### LINKED ISSUE #614: When importing from sbml file, volume is not set when size is recognized

abdelr:
Fixes #614 (Import volume from sbml file)

### DIFF
--- a/src/MoBi.Engine/Sbml/CompartmentImporter.cs
+++ b/src/MoBi.Engine/Sbml/CompartmentImporter.cs
@@ -22,16 +22,18 @@ namespace MoBi.Engine.Sbml
       internal IContainer _eventsTopContainer;
       private readonly IDimensionFactory _dimensionFactory;
       private IFormulaFactory _formulaFactory;
+      private IUnitDefinitionImporter _unitDefinitionImporter;
 
       public CompartmentImporter(IObjectPathFactory objectPathFactory, IObjectBaseFactory objectBaseFactory,
          IMoBiSpatialStructureFactory spatialStructureFactory, IMoBiDimensionFactory moBiDimensionFactory,
-         ASTHandler astHandler, IMoBiContext context, IFormulaFactory formulaFactory)
+         ASTHandler astHandler, IMoBiContext context, IFormulaFactory formulaFactory, IUnitDefinitionImporter unitDefinitionImporter)
          : base(objectPathFactory, objectBaseFactory, astHandler, context)
       {
          _objectBaseFactory = objectBaseFactory;
          _spatialStructureFactory = spatialStructureFactory;
          _dimensionFactory = moBiDimensionFactory;
          _formulaFactory = formulaFactory;
+         _unitDefinitionImporter = unitDefinitionImporter;
       }
 
       protected override void Import(Model sbmlModel)
@@ -125,12 +127,14 @@ namespace MoBi.Engine.Sbml
             volume = compartment.getVolume();
          else if (compartment.isSetSize())
             volume = compartment.getSize();
-         var dimension = _dimensionFactory.Dimension(Constants.Dimension.VOLUME);
-         IFormula formula = _formulaFactory.ConstantFormula(volume, dimension);
+
+         var unit = compartment.getUnits();
+         var baseValue = _unitDefinitionImporter.ToMobiBaseUnit(unit, volume);
+         IFormula formula = _formulaFactory.ConstantFormula(baseValue.value, baseValue.dimension);
 
          var volumeParameter = _objectBaseFactory.Create<IParameter>()
             .WithName(SBMLConstants.VOLUME)
-            .WithDimension(dimension)
+            .WithDimension(baseValue.dimension)
             .WithFormula(formula);
 
          return volumeParameter;
--- a/src/MoBi.Engine/Sbml/CompartmentImporter.cs
+++ b/src/MoBi.Engine/Sbml/CompartmentImporter.cs
@@ -21,15 +21,17 @@ namespace MoBi.Engine.Sbml
       internal IContainer _topContainer;
       internal IContainer _eventsTopContainer;
       private readonly IDimensionFactory _dimensionFactory;
+      private IFormulaFactory _formulaFactory;
 
       public CompartmentImporter(IObjectPathFactory objectPathFactory, IObjectBaseFactory objectBaseFactory,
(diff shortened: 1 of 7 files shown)


src/MoBi.Engine/Sbml/CompartmentImporter.cs
```diff
@@ -116,13 +116,18 @@ private void CreateMoleculeProperties(IContainer container)
...
-      private IEntity CreateVolumeParameter()
+      private IEntity CreateVolumeParameter(Compartment compartment)
       {
-         IFormula formula = ObjectBaseFactory.Create<ConstantFormula>().WithValue(1);
+         var volume = 1.0;
+         if (compartment.isSetVolume())
+            volume = compartment.getVolume();
+         else if (compartment.isSetSize())
+            volume = compartment.getSize();
+         IFormula formula = ObjectBaseFactory.Create<ConstantFormula>().WithValue(volume);
```
msevestre, line 126: @abdelr There is in core a FormulaFactory with a `ConstantFormula` method. This should be used here
Specifically, you will see that you need to set the dimension of the formula as well when creating a formula. At the moment this is missing. Dimension of the formula will be  the one from the containing object

tests/MoBi.Tests/Core/SBML/CompartmentImporterSpecs.cs
```diff
@@ -103,6 +105,12 @@ public void ContainerCreationTest3()
             containsSizeParam.ShouldBeTrue();
             containsVolumeParam.ShouldBeTrue();
          }
+
+         [Observation]
+         public void VolumeIsSetFromSize()
+         {
+            ((_container.Children.FirstOrDefault(x => x.Name == SBMLConstants.VOLUME) as IParameter).Formula as ConstantFormula).Value.ShouldBeEqualTo(5);
```
msevestre, line 112: Can we do sthg with less () and as
```
var parameter = _container.Parameter(SBMLConstants.VOLUME)
parameter.Value.ShouldBeEqualTo(5);
```
and if you want to check that the formula is a constant you can 
var formula = parameter.Formula as ConstantFormula (I prefer here to use Downcast<ConstantFormula> because it crashes if this is not the right type and as returns null
formula.Value etc..

msevestre (changes requested):

src/MoBi.Engine/Sbml/CompartmentImporter.cs
abdelr: Thanks for the catch. I did not know about the FormulaFactory ;)

src/MoBi.Engine/Sbml/CompartmentImporter.cs
```diff
@@ -116,13 +118,19 @@ private void CreateMoleculeProperties(IContainer container)
...
+            volume = compartment.getVolume();
+         else if (compartment.isSetSize())
+            volume = compartment.getSize();
+         var dimension = _dimensionFactory.Dimension(Constants.Dimension.VOLUME);
+         IFormula formula = _formulaFactory.ConstantFormula(volume, dimension);
 
          var volumeParameter = _objectBaseFactory.Create<IParameter>()
             .WithName(SBMLConstants.VOLUME)
-            .WithDimension(_dimensionFactory.Dimension(OSPSuite.Core.Domain.Constants.Dimension.VOLUME))
+            .WithDimension(dimension)
```
Yuri05, line 133: I miss the unit definition for the volume parameter.
As far as I understand this should be the default unit of the "volume" unit definition?

```
<unitDefinition id="volume" name="volume" metaid="volume">
        <listOfUnits>
          <unit metaid="CDMT00051" kind="litre"/>
        </listOfUnits>
</unitDefinition>
...
<listOfCompartments>
      <compartment id="default" metaid="default" sboTerm="SBO:0000290" units="volume" size="1"/>
      <compartment id="c2" name="Environment" metaid="c2" sboTerm="SBO:0000290" units="volume" outside="default" size="1"/>
    </listOfCompartments>
```

tests/MoBi.Tests/Core/SBML/CompartmentImporterSpecs.cs
```diff
@@ -103,6 +105,13 @@ public void ContainerCreationTest3()
             containsSizeParam.ShouldBeTrue();
             containsVolumeParam.ShouldBeTrue();
          }
+
+         [Observation]
+         public void VolumeIsSetFromSize()
+         {
+            var parameter = _container.Children.FirstOrDefault(x => x.Name == SBMLConstants.VOLUME) as IParameter;
```
msevestre, line 112: same as the other pr 
This could be greatly simplified. `var parameter = _container.Parameter(SBMLConstants.VOLUME)`

src/MoBi.Engine/Sbml/CompartmentImporter.cs
msevestre: @Yuri05 is right. The unit should be set as unit for the parameter. Now, is the value 1 in litre? Then we need to make sure we always convert the value in baseunit. @abdelr Is this done somewhere?

msevestre (commented):
Potentially, the DisplayUnit is not setproperly

src/MoBi.Engine/Sbml/CompartmentImporter.cs
```diff
@@ -116,13 +118,19 @@ private void CreateMoleculeProperties(IContainer container)
...
+      private IEntity CreateVolumeParameter(Compartment compartment)
       {
-         IFormula formula = ObjectBaseFactory.Create<ConstantFormula>().WithValue(1);
+         var volume = 1.0;
+         if (compartment.isSetVolume())
+            volume = compartment.getVolume();
+         else if (compartment.isSetSize())
+            volume = compartment.getSize();
+         var dimension = _dimensionFactory.Dimension(Constants.Dimension.VOLUME);
+         IFormula formula = _formulaFactory.ConstantFormula(volume, dimension);
```
msevestre, line 129: @Yuri05's point is valid. Is the `volume` value always in baseunit? or could the value be mL for instance? In this case, the value needs to be converted to baseunit. So it's not only a display issue but the value might be wrong

msevestre (changes requested):

src/MoBi.Engine/Sbml/CompartmentImporter.cs
abdelr: Definitely!! Nice catch. Thanks.

Yuri05 (approved):

PR #619, opened 2021-09-28
Parse combined units

### LINKED ISSUE #618: When importing an sbml file, rate parameters should have amount per time dimension

abdelr:
Fixes #618 (parse combined units when importing from sbml)

### DIFF
--- a/src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
+++ b/src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
@@ -12,7 +12,7 @@ using Unit = OSPSuite.Core.Domain.UnitSystem.Unit;
 
 namespace MoBi.Engine.Sbml
 {
-   class CombinedUnit
+   internal class CombinedUnit
    {
       public string Name { get => (_direct?.Name ?? "1") + (_inverse != null ? $"/{_inverse.Name}" : ""); }
       public double Rate { get; private set; } = 1;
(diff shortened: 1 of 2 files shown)


src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -103,30 +120,29 @@ public IDimension ConvertUnit(UnitDefinition unitDefinition)
...
-               return unitDimension;
-            }
+            combinedUnit.AddUnit(
+               sbmlUnitDefinition.getKind(), 
+               sbmlUnitDefinition.getExponent(), 
+               sbmlUnitDefinition.getMultiplier(),
+               _baseUnitsDictionary);
          }
+         var unitName = combinedUnit.Name;
+         var unitDimension = _moBiDimensionFactory.DimensionForUnit(unitName);
```
msevestre, line 134: ` public (IDimension dimension, Unit unit) FindUnit(string unitName)` returns the dimension and unit... as an FYI

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -12,6 +12,24 @@
 
 namespace MoBi.Engine.Sbml
 {
+   class CombinedUnit
+   {
+      public string Name { get => (_direct?.Name ?? "1") + (_inverse != null ? $"/{_inverse.Name}" : ""); }
+      public double Rate { get; private set; } = 1;
+      public void AddUnit(int kind, double exponent, double multiplier, IDictionary<int, Unit>  baseUnitsDictionary)
```
msevestre, line 19: Looks like a constructor to me no?

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -12,6 +12,24 @@
 
 namespace MoBi.Engine.Sbml
 {
+   class CombinedUnit
```
msevestre, line 15: set the type of class (internal, public, private)

msevestre (commented):
A few remarks. I would run resharper in this file because spacing is not respected

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
abdelr: Not really, if you check how it is used, you will see that it could be invoked twice. Once for setting the direct unit and a second time for setting the inverse unit (or in the other other, it does not really matter). Initially, I thought that unit definitions in an sbml file would only have one single unit but if you have "mole/l" for example you will have a unit definition with two units, one for mole with positive exponent and a second for liter with negative exponent. So we need to iterate first on all the units two collect all information (with AddUnit) to know if it is a plain inverse dimension (1/s for inverse time), a regular dimension (mole) or a combined dimension (mole/l).

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
abdelr: Good to know.

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
msevestre: yep you are correct. Did not see the loop

PR #623, opened 2021-09-30
622 newly defined units properly translate when importing from sbml files

### LINKED ISSUE #622: Newly defined units should be properly translated when importing from sbml files
should be calculated as multiplier times 10 to the power of scale and the exponent should only means direct or inverse unit

abdelr:
Fixes #622 (Using scale properly when importing unit definitions from SBML Files)

### DIFF
--- a/src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
+++ b/src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
@@ -12,7 +12,7 @@ using Unit = OSPSuite.Core.Domain.UnitSystem.Unit;
 
 namespace MoBi.Engine.Sbml
 {
-   class CombinedUnit
+   internal class CombinedUnit
    {
       public string Name { get => (_direct?.Name ?? "1") + (_inverse != null ? $"/{_inverse.Name}" : ""); }
       public double Rate { get; private set; } = 1;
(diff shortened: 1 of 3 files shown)


src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -31,16 +49,19 @@ public class UnitDefinitionImporter : SBMLImporter, IStartable, IUnitDefinitionI
       private readonly IMoBiDimensionFactory _moBiDimensionFactory;
       private readonly IDictionary<int, Unit> _baseUnitsDictionary;
       private readonly IDictionary<string, UnitConvertionInfo> _unitConvertionDictionary;
+      private IDimensionFactory _dimensionFactory;
+      public IReadOnlyDictionary<string, IDimension> ConvertionDictionary { get => _unitConvertionDictionary.ToDictionary(kv => kv.Key, kv => kv.Value.Dimension); }
       private IDictionary<string, string> _sbmlUnitsSynonyms = new Dictionary<string, string>()
```
msevestre, line 54: @abdelr Try to think Cache over Dictionary . cache exposes just so many more convenience methods
for instance you can remove the methods where you check if the value is in the cache completely by defining the behavior at the cache level

```
_sbmlUnitsSynonyms  = new Cache<string, string>(onMissingkeys = k=>k)
```
e.g what happens when the key is missing. we return the key.

msevestre (commented):
Merging but let's try to avoid using Dictionary unless absolutely required please

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
abdelr: Check.

msevestre:
@abdelr So many merge conflictts that I resolved by hand... Going to merge if the PR builds But can you check that this is all correct

msevestre:
the diff is still there in the PR
Otherwise yes, check the content of the file UnitDefinitionImporter.cs: and make sure it is correct

PR #625, opened 2021-10-04
WIP Fixes failign build

msevestre:
@abdelr I tried to fix the code but I have no idea what's going on .Why aren't you using the capability of the internal unit system to convert from one unit to another? If feels that you are multiplying TWICE with the rate when you care converting. 
I have run resharper on the file (it would be awesome if you could run it as well) to simplify the code and I also implemented all the suggestion.
Then I removed the usage of two dimensionFactory in the same file (they should be the same instance anyways). I am sure you can figure out what's going on

### DIFF
--- a/src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
+++ b/src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
@@ -14,23 +14,28 @@ namespace MoBi.Engine.Sbml
 {
    internal class CombinedUnit
    {
-      public string Name { get => (_direct?.Name ?? "1") + (_inverse != null ? $"/{_inverse.Name}" : ""); }
+      public string Name => (_direct?.Name ?? "1") + (_inverse != null ? $"/{_inverse.Name}" : "");
+
       public double Rate { get; private set; } = 1;
+
+      private Unit _direct;
+      private Unit _inverse;
+
       public void AddUnit(int kind, double exponent, double multiplier, double scale, IDictionary<int, Unit> baseUnitsDictionary)
       {
          if (!baseUnitsDictionary.ContainsKey(kind))
             return;
+      
          if (exponent < 0)
             _inverse = baseUnitsDictionary[kind];
          else
             _direct = baseUnitsDictionary[kind];
+
          Rate *= multiplier * Math.Pow(10, scale);
       }
-      private Unit _direct { get; set; }
-      private Unit _inverse { get; set; }
    }
 
-   internal class UnitConvertionInfo
+   internal class UnitConversionInfo
    {
       public IDimension Dimension { get; set; }
       public Unit Unit { get; set; }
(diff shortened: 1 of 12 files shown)


src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -14,23 +14,28 @@ namespace MoBi.Engine.Sbml
...
             return;
+      
          if (exponent < 0)
             _inverse = baseUnitsDictionary[kind];
          else
             _direct = baseUnitsDictionary[kind];
+
          Rate *= multiplier * Math.Pow(10, scale);
       }
-      private Unit _direct { get; set; }
```
msevestre, line 29: private menber. Why a get set?

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -48,37 +53,39 @@ public class UnitDefinitionImporter : SBMLImporter, IStartable, IUnitDefinitionI
    {
       private readonly IMoBiDimensionFactory _moBiDimensionFactory;
       private readonly IDictionary<int, Unit> _baseUnitsDictionary;
-      private readonly IDictionary<string, UnitConvertionInfo> _unitConvertionDictionary;
-      private IDimensionFactory _dimensionFactory;
```
msevestre, line 52: why this guy here sinec we have a MoBiDimensionFactory (which is a specification of IDimensionFActory)

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -48,37 +53,39 @@ public class UnitDefinitionImporter : SBMLImporter, IStartable, IUnitDefinitionI
...
 
-      private void InitBaseUnitsDictionary()
+      private void initBaseUnitsDictionary()
       {
          _baseUnitsDictionary.Add(libsbml.UNIT_KIND_AMPERE, _moBiDimensionFactory.Dimension("Ampere").Unit("A"));
          _baseUnitsDictionary.Add(libsbml.UNIT_KIND_BECQUEREL, _moBiDimensionFactory.Dimension("Becquerel").Unit("Bq"));
          _baseUnitsDictionary.Add(libsbml.UNIT_KIND_CANDELA, _moBiDimensionFactory.Dimension("Candela").Unit("cd"));
          _baseUnitsDictionary.Add(libsbml.UNIT_KIND_COULOMB, _moBiDimensionFactory.Dimension("Coulomb").Unit("C"));
-         _baseUnitsDictionary.Add(libsbml.UNIT_KIND_DIMENSIONLESS, _moBiDimensionFactory.NoDimension.DefaultUnit);
+         _baseUnitsDictionary.Add(libsbml.UNIT_KIND_DIMENSIONLESS, _moBiDimensionFactory.NoDimension.BaseUnit);
```
msevestre, line 88: I am pretty sure you mean BaseUnit and not DefaultUnit. We always need to convert to BaseUnit.

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -131,7 +140,7 @@ public IDimension ConvertUnit(UnitDefinition unitDefinition)
          if (dimension != Constants.Dimension.NO_DIMENSION)
          {
             _sbmlInformation.MobiDimension[sbmlUnit] = dimension;
-            _unitConvertionDictionary.Add(sbmlUnit, new UnitConvertionInfo() { Dimension = dimension, Unit = dimension.DefaultUnit, Rate = 1 });
+            _unitConversionDictionary.Add(sbmlUnit, new UnitConversionInfo {Dimension = dimension, Unit = dimension.BaseUnit, Rate = 1});
```
msevestre, line 143: I am pretty sure you mean BaseUnit and not DefaultUnit. We always need to convert to BaseUnit.

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -146,45 +155,48 @@ public IDimension ConvertUnit(UnitDefinition unitDefinition)
...
          if (dimensionAndUnit.dimension != Constants.Dimension.NO_DIMENSION)
          {
             _sbmlInformation.MobiDimension[sbmlUnit] = dimensionAndUnit.dimension;
-            _unitConvertionDictionary.Add(sbmlUnit, new UnitConvertionInfo() { Dimension = dimensionAndUnit.dimension, Unit = dimensionAndUnit.unit, Rate = combinedUnit.Rate });
+            _unitConversionDictionary.Add(sbmlUnit, new UnitConversionInfo {Dimension = dimensionAndUnit.dimension, Unit = dimensionAndUnit.unit, Rate = combinedUnit.Rate});
             return dimensionAndUnit.dimension;
          }
 
-         _unitConvertionDictionary.Add(sbmlUnit, new UnitConvertionInfo() { Dimension = dimension, Unit = dimension.DefaultUnit, Rate = 1 });
+         _unitConversionDictionary.Add(sbmlUnit, new UnitConversionInfo {Dimension = dimension, Unit = dimension.BaseUnit, Rate = 1});
```
msevestre, line 168: I am pretty sure you mean BaseUnit and not DefaultUnit. We always need to convert to BaseUnit.

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
```diff
@@ -146,45 +155,48 @@ public IDimension ConvertUnit(UnitDefinition unitDefinition)
...
 
       public (double value, IDimension dimension) ToMobiBaseUnit(string unit, double value)
       {
-         if (_unitConvertionDictionary.ContainsKey(unit))
+         if (_unitConversionDictionary.ContainsKey(unit))
          {
-            var convertionData = _unitConvertionDictionary[unit];
-            return (convertionData.Dimension.UnitValueToBaseUnitValue(convertionData.Unit, value * convertionData.Rate), convertionData.Dimension);
+            var conversionData = _unitConversionDictionary[unit];
+            return (conversionData.Dimension.UnitValueToBaseUnitValue(conversionData.Unit, value * conversionData.Rate), conversionData.Dimension);
```
msevestre, line 189: Why do we need this rate? It feels that yo are converting twice since we are also converting using the value

tests/MoBi.Tests/Core/SBML/UnitImporterSpecs.cs
```diff
@@ -17,95 +17,78 @@ namespace MoBi.Core.SBML
...
+
+         _sbmlModel.addUnitDefinition(_unitDef);
+         sut.DoImport(_sbmlModel, new MoBiProject(), A.Fake<SBMLInformation>(), new MoBiMacroCommand());
+      }
+
+      [Observation]
+      public void NewDimensionsTests()
+      {
+         sut.ToMobiBaseUnit("substance", 3e6).value.ShouldBeEqualTo(3e12);
+         sut.ToMobiBaseUnit("inverse_day", 0.83).value.ShouldBeEqualTo(0.83 / 1440); //base unit is in minutes so 1440 minutes in one day
```
msevestre, line 87: but isn't the value in second based on _unit3?

tests/MoBi.Tests/Core/SBML/UnitImporterSpecs.cs
```diff
@@ -117,7 +100,7 @@ protected override void Context()
       [Observation]
       public void ParameterWithVolumeUnitTest()
       {
-         var tc = _moBiProject.SpatialStructureCollection.FirstOrDefault().TopContainers.FirstOrDefault();
+         var tc = _moBiProject.SpatialStructureCollection.First().TopContainers.First();
```
msevestre, line 103: It should always be First since you are calling something after. FirstOrDefault does not make sense. Probably old code but we might as well fix it

msevestre:
@abdelr The test is still failing. Just some cleanup

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
abdelr: @msevestre There are indeed two conversions. The first conversion is based on the description on the sbml file. The second is to the mobi base unit. Suppose there is a description from unit A to a (sbml) base unit B described in the file. This will be captured by the rate. Unit A does not exist on mobi but unit B does. Additionally, unit B needs to be translated to mobi base unit C. More specifically, we have an entry on the sbml file that specifies that "unit A" is going to be used and it has a given rule to be translated to "mole" (which is a conventional unit known in mobi too). This translation is captured in rate. But when we translate in mobi, we actually need to translate from "unit A" to "micro mole" which is the base unit, so we need to translate from "unit A" to "mole" and then to "micro mole". Does it make sense? It is not a clean explanation but we could talk about this tomorrow if you like. I will take a look on the changes though and also on the failing test.

tests/MoBi.Tests/Core/SBML/UnitImporterSpecs.cs
abdelr: Sure, you are right.

tests/MoBi.Tests/Core/SBML/UnitImporterSpecs.cs
abdelr: Minutes is the base unit, right?

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
abdelr: Sure

src/MoBi.Engine/Sbml/UnitDefinitionImporter.cs
abdelr: Not really needed, sometimes I do this for finding easier where it is been used. So even when it is private, it is a full property and you can easily navigate but you are right, not necessary.

msevestre: Resharper navigates with private members...plus it follows the rest of the code

msevestre (commented):
I think I found the issue. The factor was wrong. I have also added some comments. Please let me know if they are accurate

abdelr (approved):

msevestre:
@abdelr I forgot to publish my comment yesterday. Since you approved, I am assuming it's fine

abdelr:
@msevestre Yes, I looked into the changes and made perfect sense for me.

PR #629, opened 2021-10-05
Fixes #628 mapping issue

### LINKED ISSUE #628: Importer configuration created in PK-Sim not used correctly
A configuration has been created in PK-Sim for the following mapping:

image

Using the same configuration with the same file in MoBi:

image

1) The column "Molecular weight" cannot be selected for molecular weight
2) "Mapping Name" is "Molecular **W**eight" in PK-Sim and "Molecular **w**eight" in MoBi
3) Columns "Species", "Study Id", "Subject Id", "Gender", "Dose", "Route" are not shown as additional grouping (why do the lists of mapping settings differ betwen PK-Sim and MoBi at all?)
4) After import, all these meta data are actually present (which is good!)

image

But "Molecular Weight" is also considered as meta data and not the actual molecular weight property.

msevestre:


src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -10,17 +10,17 @@
...
 using OSPSuite.Utility.Extensions;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using OSPSuite.Infrastructure.Import.Core;
 using ColumnInfo = OSPSuite.Infrastructure.Import.Core.ColumnInfo;
 using OSPSuite.Infrastructure.Import.Services;
 using static OSPSuite.Core.Domain.Constants;
 using Command = OSPSuite.Assets.Command;
-using DimensionInfo = OSPSuite.Infrastructure.Import.Core;
```
msevestre, line 23: Still some old stuff here. In fact most of the code in this file seemed to be outdated. It was not using the default configuration etc...

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -158,16 +159,30 @@ private static MoBiMacroCommand deleteAllResultsFromSimulationCommand(IMoBiSimul
...
          };
-         addNamingPatterns(settings);
-         settings.NameOfMetaDataHoldingMolecularWeightInformation = AppConstants.Parameters.MOLECULAR_WEIGHT;
-         return settings;
+
+         addNamingPatterns(dataImporterSettings);
+         dataImporterSettings.NameOfMetaDataHoldingMoleculeInformation = Constants.ObservedData.MOLECULE;
+         dataImporterSettings.NameOfMetaDataHoldingMolecularWeightInformation = Constants.ObservedData.MOLECULAR_WEIGHT;
+
+         var metaDataCategories = _dataImporter.DefaultMetaDataCategories().ToList();
```
msevestre, line 174: Creating the default list and populate from predefined values

msevestre (commented):
Issue is also fixed Mapping is workign as expected

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -275,15 +290,14 @@ where result
 
       private IEnumerable<DataRepository> getObservedDataFromImporter(ImporterConfiguration configuration)
       {
-         var dataImporterSettings = createDataImportSettings();
+         var (metaDataCategories, dataImporterSettings)  = initializeSettings();
 
          //do we really need this in MoBi????
-         dataImporterSettings.NameOfMetaDataHoldingMoleculeInformation = Constants.ObservedData.MOLECULE;
```
msevestre, line 281: Yes we need it

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -158,16 +159,30 @@ private static MoBiMacroCommand deleteAllResultsFromSimulationCommand(IMoBiSimul
...
+      private (IReadOnlyList<MetaDataCategory>, DataImporterSettings) initializeSettings()
       {
-         var settings = new DimensionInfo.DataImporterSettings
+         var dataImporterSettings = new DataImporterSettings
          {
             IconName = ApplicationIcons.MoBi.IconName,
             Caption = $"{AppConstants.PRODUCT_NAME} - {AppConstants.Captions.ImportObservedData}"
          };
-         addNamingPatterns(settings);
-         settings.NameOfMetaDataHoldingMolecularWeightInformation = AppConstants.Parameters.MOLECULAR_WEIGHT;
```
msevestre, line 169: This was using the wrong constant! Name of the parameters instead of the ObservedDataConstant defined.

PR #635, opened 2021-10-07
634 translate constants when importing reactions from sbml

### LINKED ISSUE #634: When importing SBML Files, the constants on the reactions should be also translated to base units

abdelr:
Fixes #634 (translate constants to base units when importing reactions from sbml)


abdelr:
> Looks like there are many changes from the other PR. @abdelr Can you do the quick changes that I proposed and then I'll merge (merely code consistency and not using magic string)

@msevestre Done already

PR #680, opened 2022-01-28
666 importing mol weight

### LINKED ISSUE #666: Importer: setting molweight of imported observed data not possible
**Problem Description**

When mapping to a project molecule during the import process, the Molecular weight value does not get set according the value of the molecule. After the import, it is not possible to edit the MW of observed data that does not already have a molecular weight. This means that when the user tries to import a file with the following mapping:

image

the result is the following observed data:

image

missing the possibilty to edit the MW. Instead there should be the possibility to edit the MW, like in the following case where there is one set:

image

The background to the difference between this behaviour in PK-Sim and in MoBi, is that in MoBi it is possible to have more than one molecules with the same name from different building blocks and with different Molecular weights. As a result we do not assign the MW of the Molecule to the observed data.

**Proposed Solution** 

a)	In MoBi we should get the MW from the molecule if there is only one molecule with that name, or many molecules with the same MW value. If there are more than one MW values we leave the value empty. This value is just an initial state in MoBi and not linked to the molecule and possible changes of it.
 
b)	In MoBi if we have mapped both a Molecule and a MW then the MW value coming from the excel should overwrite the MW value coming from the molecule in the project.

c)	Both in MoBi and PK-Sim we should always be able to see and edit the MW in the observed data. Only exception: if in PK-Sim the MW comes from a molecule, we should still be able to see it but not edit it.

### LINKED ISSUE #667: Importer: MoBi: Cannot import molecular weight if it does not match the molecule
Importing observed data with a column for molecular weight and mapping a molecule from the project. If the MW in the column does not match the MW in the molecule, an error is thrown.

image

I think we agreed on not checking it for MoBi! (you can use the "Aciclivir.pkml" from the r package for testing: https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/inst/extdata/Aciclovir.pkml)
[Acicloriv.xlsx](https://github.com/Open-Systems-Pharmacology/MoBi/files/7838382/Acicloriv.xlsx)

### LINKED ISSUE #668: Imorter: MoBi: cannot import MW
https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/develop/inst/extdata/Aciclovir.pkml

with [Acicloriv.xlsx](https://github.com/Open-Systems-Pharmacology/MoBi/files/7838382/Acicloriv.xlsx)

with this mapping:

image

MW is just not applied:

image

georgeDaskalakis:
closes #666
closes #667 
closes #668

### DIFF
--- a/src/MoBi.Presentation/Tasks/ObservedDataTask.cs
+++ b/src/MoBi.Presentation/Tasks/ObservedDataTask.cs
@@ -164,7 +164,8 @@ namespace MoBi.Presentation.Tasks
          var dataImporterSettings = new DataImporterSettings
          {
             IconName = ApplicationIcons.MoBi.IconName,
-            Caption = $"{AppConstants.PRODUCT_NAME} - {AppConstants.Captions.ImportObservedData}"
+            Caption = $"{AppConstants.PRODUCT_NAME} - {AppConstants.Captions.ImportObservedData}",
+            CheckMolWeightAgainstMolecule = false
          };
 
          addNamingPatterns(dataImporterSettings);
@@ -182,7 +183,7 @@ namespace MoBi.Presentation.Tasks
       {
          addPredefinedOrganValues(metaDataCategories.FindByName(Constants.ObservedData.ORGAN));
          addPredefinedCompartmentValues(metaDataCategories.FindByName(Constants.ObservedData.COMPARTMENT));
-         addPredefinedMoleculeNames(metaDataCategories.FindByName(Constants.ObservedData.MOLECULE));
+         addPredefinedMoleculesForImporter(metaDataCategories.FindByName(Constants.ObservedData.MOLECULE));
       }
 
       public override void Rename(DataRepository dataRepository)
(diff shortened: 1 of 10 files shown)


src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -384,6 +385,26 @@ private void addDimensionsTo(ColumnInfo columnInfo)
...
+      private void addPredefinedMoleculesForImporter(MetaDataCategory metaDataCategory)
+      {
+         if (metaDataCategory == null)
+            return;
+
+         metaDataCategory.ShouldListOfValuesBeIncluded = true;
+
+         foreach (var molecule in allMolecules())
+         {
+            var molWeightParameter = molecule.AllParameters().FirstOrDefault(x => x.Name == AppConstants.Parameters.MOLECULAR_WEIGHT);
```
Yuri05, line 397: you could use 
`var molWeightParameter = molecule.Parameter(AppConstants.Parameters.MOLECULAR_WEIGHT)`

georgeDaskalakis: done

PR #696, opened 2022-03-08
Fixes #690 deleting node

### LINKED ISSUE #690: Deleting a molecule node connected to a reaction does not remove the molecul from reaction
[TwinNodeRemoval.zip](https://github.com/Open-Systems-Pharmacology/MoBi/files/8186201/TwinNodeRemoval.zip)

If a twin node of a molecule is present in the Reactions-chart, deleting the connected node does not remove the molecule from reaction:

image

- Selected the connected molecule and press "del":

image

Also in the simulation, the molecule is produced.

msevestre:


src/MoBi.UI/Presenters/ReactionDiagramPresenter.cs
```diff
@@ -179,18 +180,16 @@ private bool moleculeNodeWillBeUnlinkedAfterDelete(MoleculeNode moleculeNode, IE
...
+            RemoveReactionNode((ReactionNode) itemToDelete);
       }
 
       private bool anyLinkedNodes(MoleculeNode moleculeNode)
       {
-         var moleculeNodes = reactionDiagramManager.GetMoleculeNodes(moleculeNode.Name);
          var reactionNodes = getLinkedReactionsForMoleculeNode(moleculeNode);
-
-         return reactionNodes.Any() && moleculeNodes.Count() == 1;
+         return reactionNodes.Any();
```
msevestre, line 192: This is the one change. If there were multiple nodes, the prompt was not shown

msevestre: I do not know why it was like this and there are not tests to explain the behavior.  I tried delete and multiple delete, seems to be working fine

PR #705, opened 2022-03-15
703 crash psv

msevestre:


src/MoBi.Presentation/Tasks/Interaction/AbstractStartValuesTask.cs
```diff
@@ -166,10 +166,16 @@ protected bool HasEquivalentFormula(IStartValue startValue, IFormula targetFormu
          return _startValuePathTask.HasEquivalentFormula(startValue, targetFormula);
       }
 
+      protected static bool HasEquivalentStartValue(IStartValue startValue, IParameter parameter)
+      {
+         var (value, _) = parameter.TryGetValue();
+         return HasEquivalentStartValue(startValue, value);
```
msevestre, line 172: This was causing the crash before. Sometimes, we can have formula that cannot be evaluated and it was crashign the code. I am using our new TryGetValue here in the overload

src/MoBi.UI/Views/EditParameterListView.cs
```diff
@@ -120,7 +120,7 @@ public override void InitializeBinding()
 
          _pathBinder.InitializeBinding(_gridViewBinder);
          _gridViewBinder.Bind(dto => dto.Value)
-            .WithFormat(dto => dto.ParameterFormatter())
+            .WithFormat(dto => dto.ParameterFormatter(checkForEditable:false))
```
msevestre, line 123: Thsi is the other issue that was showing different value for NaN depending in editable or not. MoBi does not care for this so I want to make sure it's consistent.

msevestre (commented):
Also solve some issues with layout

Yuri05 (approved):

PR #717, opened 2022-04-22
Update dll after breakiong change in core

msevestre:


msevestre:
@Yuri05 Letting the tests run. Very little changes required here

PR #733, opened 2022-07-01
732 reset layout

msevestre:


src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -250,12 +245,21 @@ protected override void ChartChanged()
          MarkChartOwnerAsChanged();
       }
 
+      public override void InitializeAnalysis(CurveChart chart)
```
msevestre, line 248: @Yuri05
I am only initializing the chart once. In PKSim, this is indeed done only once that's why we did not see it
I am going to update the templates also to ensure that the pin is disabled

Yuri05 (approved):

PR #735, opened 2022-07-03
Fixes #725 MSV problems

### LINKED ISSUE #725: Layout problems in MSV-BB
image

msevestre:


src/MoBi.UI/Views/BaseStartValuesView.cs
```diff
@@ -66,30 +64,6 @@ protected BaseStartValuesView(ValueOriginBinder<TStartValue> valueOriginBinder)
          _screenBinder = new ScreenBinder<IStartValuesPresenter<TStartValue>>();
       }
 
-      public override void InitializeResources()
-      {
-         base.InitializeResources();
-
-         scaleApplyToSelectionLayoutItem(layoutItemIsPresent);
```
msevestre, line 73: @Yuri05 Trying to do too much sometimes..., Here it seems to work when we don't try to size ourselves
image

PR #760, opened 2022-09-27
746 update svg

msevestre:


msevestre:
Also added the version number. Can be helpful

PR #761, opened 2022-09-29
Update core to 11.1.22 and simmodel to 4.0.0.52

rwmcintosh:


src/MoBi.Core/Services/SimModelManagerFactory.cs
```diff
@@ -35,7 +35,7 @@ public ISimModelManager Create()
 
       private DataFactory createDataFactory()
       {
-         return new DataFactory(_context.DimensionFactory, _context.ObjectPathFactory, _displayUnitRetriever, _dataRepositoryTask);
+         return new DataFactory(_context.DimensionFactory, _displayUnitRetriever, _dataRepositoryTask);
```
msevestre, line 38: what the heck. Why is this used in MoBi?
Does PK-Sim creates its own as well?!?

@rwmcintosh Can you create an issue for this to remove this weird stuff if possible

PR #770, opened 2022-10-06
Fixes #769 parameter should be global for interaction container

### LINKED ISSUE #769: Parameter added to an inhibition/induction process should be GLOBAL
I created a new parameter in the Induction-Process container of a molecule in the Molecules-BB:

image

When I create a simulation with this building block, the parameter is not there:

image

@Yuri05 @msevestre Could you explain? Am I doing something wrong?

msevestre:


src/MoBi.Presentation/Mappers/MoleculeBuilderDTOToTreeNodeMapper.cs
```diff
@@ -1,83 +1,83 @@
...
       {
-         _activeTransportBuilderContaienrToMolceculeTreeNodeMapper = activeTransportBuilderContaienrToMolceculeTreeNodeMapper;
+         _activeTransportBuilderContainerToMoleculeTreeNodeMapper = activeTransportBuilderContainerToMoleculeTreeNodeMapper;
          _interactionContainerMapper = interactionContainerMapper;
       }
 
-      public IMoleculeTreeNode MapFrom(MoleculeBuilderDTO input)
+      public IMoleculeTreeNode MapFrom(MoleculeBuilderDTO moleculeBuilderDTO)
       {
-         var node = new MoleculeTreeNode(input) {Text = input.Name, Icon = ApplicationIcons.IconByName(input.Icon)};
```
msevestre, line 28: All the changes here are just a million typos in the file. Not related to actual change

tests/MoBi.Tests/Core/Commands/AddParameterToContainerCommandSpecs.cs
```diff
@@ -62,6 +62,19 @@ protected override ParameterBuildMode GetBuildModeType()
...
+   public class When_adding_a_new_parameter_to_interaction_container : When_testing_for_appropriate_parameter_build_modes
+   {
+      protected override IContainer GetContainer()
+      {
+         return new InteractionContainer();
+      }
+
+      protected override ParameterBuildMode GetBuildModeType()
+      {
+         return ParameterBuildMode.Global;
```
msevestre, line 74: the test

src/MoBi.Core/Domain/Extensions/ContainerExtensions.cs
```diff
@@ -27,7 +27,8 @@ public static IEnumerable<ParameterBuildMode> AvailableBuildModeForParameters(th
       public static ParameterBuildMode DefaultParameterBuildMode(this IContainer container)
       {
          var needsGlobalParameter = container.IsAnImplementationOf<TransporterMoleculeContainer>() ||
-                                    container.IsAnImplementationOf<IReactionBuilder>();
+                                    container.IsAnImplementationOf<IReactionBuilder>() ||
+                                    container.IsAnImplementationOf<InteractionContainer>();
```
msevestre, line 31: the fix

PavelBal (approved):

tests/MoBi.Tests/Core/Commands/AddParameterToContainerCommandSpecs.cs
Yuri05: Observations from the base test class are automatically executed when running the tests in the derived class?
I did not know it

msevestre: yeah it's because it's derived and we try to find an Obervation in each class. Pretty neat (but potentially way over engineered in this case)

Yuri05: I think it's pretty usefull. 
In the case like this I have always defined a function `testXY()` in the base class and the observation in each derived class which just called `testXY()` - which could be avoided with this feature :)

Issue #768, opened 2022-10-06
Wrong icon for "Events"-BB


PavelBal:
When editing an Events-BB, the "Passive Transports"-icon is shown in the tab.

![image](https://user-images.githubusercontent.com/25903040/194297010-1cf486f9-3f40-4f9a-a8ec-048e8c114562.png)

msevestre:
I can confirm. WTF

msevestre:
this is fixed

PR #771, opened 2022-10-06
Fxies #768 wrong icon

msevestre:


msevestre:
@Yuri05 tRivial fix here. Merging

PR #777, opened 2022-10-20
Pluging in new Simulation OutputMapings and Charts

georgeDaskalakis:


georgeDaskalakis:
@msevestre converting this to a draft, since moving code from PK-Sim to Core will also affect this PR.

georgeDaskalakis:
@msevestre this is also ready with the latest Core dlls.

src/MoBi.Core/Serialization/Xml/Serializer/MoBiSimulationXmlSerializer.cs
```diff
@@ -16,7 +16,7 @@ public MoBiSimulationXmlSerializer() : base(SerializationConstants.MoBiSimulatio
       public override void PerformMapping()
       {
          base.PerformMapping();
-         Map(x => x.Results);
+         Map(x => x.ResultsDataRepository);
```
msevestre, line 19: This is not good here. You are breaking serialization

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -75,6 +86,7 @@ public abstract class ChartPresenter : ChartPresenter<CurveChart, IChartView, IC
 
          _chartTemplatingTask = chartTemplatingTask;
          _dataRepositoryCache = new Cache<DataRepository, IMoBiSimulation>(onMissingKey: x => null);
+         _OutputMappingMatchingTask = OutputMappingMatchingTask;
```
msevestre, line 89: Casing is off here. Follow coding standard here

msevestre (changes requested):
The serialization isa problem
You need to name the mapping explicitly to be like the old one

src/MoBi.Presentation/Presenter/ChartPresenter.cs
georgeDaskalakis: done

src/MoBi.Core/Serialization/Xml/Serializer/MoBiSimulationXmlSerializer.cs
georgeDaskalakis: Not sure how you mean. Let's briefly discuss this today.

georgeDaskalakis:
@msevestre is there something more to be changed for this PR? I would say this is ready for merging

PR #788, opened 2022-11-02
Fixes #787 define domain object for expression profile

### LINKED ISSUE #787: Define expression profile domain object
@rwmcintosh commented on [Tue Oct 25 2022](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/1731)

Properties of this object:
- Name of the protein
- Type (enzyme/binding partner/transporter)
- PK-Sim version
- List of StartValue type objects (See StartValueBase)
   - Name => CoreExpressionProfile
   - Name each list item => ExpressionParameter or something like this
   - Don't use too many Interfaces (eg. IStartValueBase is probably not necessary)

rwmcintosh:


src/MoBi.Core/OSPSuite.Core/Domain/Builder/ExpressionParameter.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace OSPSuite.Core.Domain.Builder
+{
+   public class ExpressionParameter : PathAndValueEntity
```
rwmcintosh, line 3: I think ParameterStartValue will look the same. MoleculeStartValue has some extra stuff.

src/MoBi.Core/OSPSuite.Core/Domain/Builder/PathAndValueEntity.cs
```diff
@@ -0,0 +1,157 @@
+﻿using System;
+using System.Linq;
+using OSPSuite.Core.Domain.Formulas;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Core.Domain.UnitSystem;
+
+namespace OSPSuite.Core.Domain.Builder
```
rwmcintosh, line 7: I think this is what you had in mind with a common object. Basically, we'd just need to rename the x.StartValue to x.Value all over the place in Core and MoBi.

tests/MoBi.Tests/OSPSuiteCore.Tests/CoreExpressionProfileSpecs.cs
```diff
@@ -0,0 +1,44 @@
+﻿using OSPSuite.BDDHelper;
+using OSPSuite.BDDHelper.Extensions;
+using OSPSuite.Core.Domain.Builder;
+
+namespace MoBi.OSPSuiteCore.Tests
+{
+   public class concern_for_CoreExpressionProfile : ContextSpecification<CoreExpressionProfile>
```
rwmcintosh, line 7: Most of the stuff in these objects are just setting properties. So I'm not testing it.

appveyor.yml
```diff
@@ -53,6 +53,7 @@ skip_tags: true
 branches:
   only:
     - develop
+    - OSMOSES
```
rwmcintosh, line 56: Not sure if this is what you meant by enabling the branch on appveyor

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System.Collections;
+using OSPSuite.Utility.Collections;
+using System.Collections.Generic;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Extensions;
+using static OSPSuite.Core.CoreConstants.ContainerName;
+
+namespace OSPSuite.Core.Domain.Builder
+{
+   public class CoreExpressionProfile : BuildingBlock, IEnumerable<ExpressionParameter>
```
msevestre, line 10: I think this should implement the BuildingBlockStartValue T no? There is something like this for MoBi. Check ParameterStartValueBuildingBlock

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
```diff
@@ -0,0 +1,88 @@
...
+   {
+      private readonly ICache<IObjectPath, ExpressionParameter> _allValues;
+      public CoreExpressionProfile()
+      {
+         _allValues = new Cache<IObjectPath, ExpressionParameter>(x => x.Path, x => null);
+      }
+
+      public virtual string MoleculeName { set; get; }
```
msevestre, line 18: I don't think this is ok. We are inferring from the name so let's always retrieve from the name as well so that it's always consistent

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
```diff
@@ -0,0 +1,88 @@
...
+
+      public string Species { set; get; }
+
+      public ExpressionType Type { set; get; }
+
+      public int PKSimVersion { set; get; }
+
+      public virtual string Category { set; get; }
```
msevestre, line 26: I don't think this is ok. We are inferring from the name so let's always retrieve from the name as well so that it's always consistent

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
```diff
@@ -0,0 +1,88 @@
...
+      {
+         get => ExpressionProfileName(MoleculeName, Species, Category);
+         set
+         {
+            if (string.Equals(Name, value))
+            {
+               return;
+            }
+
+            var (moleculeName, _, category) = NamesFromExpressionProfileName(value);
```
msevestre, line 38: probably MoleculeName should be assigned too. Thisis not done in PKSim because it's the name of the molecule but needs to be done here. (Molecule private set same as Category)

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
```diff
@@ -0,0 +1,88 @@
...
+      }
+
+      public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager cloneManager)
+      {
+         base.UpdatePropertiesFrom(source, cloneManager);
+         var sourceExpressionProfile = source as CoreExpressionProfile;
+         if (sourceExpressionProfile == null) return;
+
+         sourceExpressionProfile.Each(value => Add(cloneManager.Clone(value)));
+         MoleculeName = sourceExpressionProfile.MoleculeName;
```
msevestre, line 53: PRobably not required as setting Name should set Molecule and CAtegory

src/MoBi.Core/OSPSuite.Core/Domain/Builder/PathAndValueEntity.cs
msevestre: Renaming probably not at the MoleculeSV and PSV level as we'll have to write converter
Or we can do it but we need to start adding an ongoing task for stuff to convert
Alternatively you have Value in the base class and keep StartValue <=> Value

src/MoBi.Core/OSPSuite.Core/Domain/Builder/PathAndValueEntity.cs
```diff
@@ -0,0 +1,157 @@
...
+         Value = sourcePathAndValueEntity.Value;
+         ContainerPath = sourcePathAndValueEntity.ContainerPath.Clone<IObjectPath>();
+         DisplayUnit = sourcePathAndValueEntity.DisplayUnit;
+         Dimension = sourcePathAndValueEntity.Dimension;
+         Formula = cloneManager.Clone(sourcePathAndValueEntity.Formula);
+         ValueOrigin.UpdateAllFrom(sourcePathAndValueEntity.ValueOrigin);
+      }
+
+      public override string ToString()
+      {
```
msevestre, line 153: this can be inlined

tests/MoBi.Tests/OSPSuiteCore.Tests/CoreExpressionProfileSpecs.cs
```diff
@@ -0,0 +1,44 @@
...
+
+   public class when_setting_the_name_of_the_building_block : concern_for_CoreExpressionProfile
+   {
+      protected override void Because()
+      {
+         sut.Name = "Molecule|Species|Phenotype";
+      }
+
+      [Observation] 
+      public void the_name_should_set_the_category_of_the_buidling_block()
```
msevestre, line 39: same with Species and Moleucle.

msevestre (changes requested):

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
rwmcintosh: I think you meant BuildingBlock<T> which is what both StartValueBuildingBlocks do. If so, then you are right. Then I can get rid of all the enumerator stuff.

msevestre: no I just checked the code in core
it's called StartValueBuildingBlock<T>

Which ha all the behavior you need. While this building block base class can be renamed to, I think we should use this instead of reimplementing the mgt of list etc?

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
```diff
@@ -0,0 +1,56 @@
+﻿using OSPSuite.Core.Domain.Services;
+using static OSPSuite.Core.CoreConstants.ContainerName;
+
+namespace OSPSuite.Core.Domain.Builder
+{
+   public class CoreExpressionProfile : BuildingBlock<ExpressionParameter>
```
msevestre, line 6: Oh and this is not that here. This now makes the expressionPArameters  as list. We want to be able to access them by path too using a cache
StartValueBuildingBlock<T> is the one (look at PSVBB and MSVBB)

msevestre (changes requested):

src/MoBi.Core/OSPSuite.Core/Domain/Builder/CoreExpressionProfile.cs
rwmcintosh: you are saying to use StartValueBuildingBlock<T> as the base here? On promotion we can rename this base class PathAndValueBuildingBlock or something like that?

rwmcintosh: Done

PR #797, opened 2022-11-12
796 make sure we load

msevestre:


msevestre:
I have fixed quite a few bugs at the same time. 
For instance, OutputMappings were not serilaized at all in MoBi

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -74,7 +74,7 @@ public bool IsCreatedBy(IBuildingBlock templateBuildingBlock)
 
       public bool UsesObservedData(DataRepository dataRepository)
       {
-         return Charts.Any(x => chartUsesObservedData(dataRepository, x));
+         return OutputMappings.Any(x => x.UsesObservedData(dataRepository)) || Charts.Any(x => chartUsesObservedData(dataRepository, x));
```
msevestre, line 77: I think this is a good way to check if an observed data is used. Either it's used in the chart or it's referenced in the output mappings.

src/MoBi.Core/Domain/Services/CloneManagerForSimulation.cs
```diff
@@ -31,15 +31,15 @@ public IMoBiSimulation CloneSimulation(IMoBiSimulation simulationToClone)
...
-            MoleculesInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.MoleculesInfo),
-            ReactionsInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.ReactionsInfo),
-            SpatialStructureInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.SpatialStructureInfo),
-            PassiveTransportsInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.PassiveTransportsInfo),
-            ObserversInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.ObserversInfo),
-            EventGroupsInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.EventGroupsInfo),
-            ParameterStartValuesInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.ParameterStartValuesInfo),
-            MoleculeStartValuesInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.MoleculeStartValuesInfo),
-            SimulationSettingsInfo = CloneBuidingBlockInfo(simulationToClone.MoBiBuildConfiguration.SimulationSettingsInfo)
+            MoleculesInfo = CloneBuildingBlockInfo(simulationToClone.MoBiBuildConfiguration.MoleculesInfo),
```
msevestre, line 34: typo

src/MoBi.Core/Serialization/Services/DeserializedReferenceResolver.cs
```diff
@@ -41,18 +41,30 @@ public void ResolveFormulaAndTemplateReferences(object deserializedObject, IMoBi
...
       }
 
+      private void updateOutputMappings(IMoBiSimulation simulation, OutputMappings outputMappings)
+      {
+         if (outputMappings == null)
+            return;
+
+         simulation.OutputMappings = outputMappings;
+         //It is required to update the references as the simulation created may have another reference as the one deserialized
+         outputMappings.Each(x => x.UpdateSimulation(simulation));
```
msevestre, line 65: I am not quite SURE why this is required. But MoBi does not use the SimulationLoader. Instead something else is used and that creates issue with references

rwmcintosh (changes requested):
Just that typo to fix please.

msevestre:
@rwmcintosh I think those changes can be done by the reviewer if it's only a little thing? This is probably much faster than back and forth. We could discuss what makes sense

msevestre (commented):
Update typos

rwmcintosh:
> @rwmcintosh I think those changes can be done by the reviewer if it's only a little thing? This is probably much faster than back and forth. We could discuss what makes sense

Oh, like in the editor if it's just a comment typo? Yeah, that makes sense.

PR #803, opened 2022-11-16
update core to latest

rwmcintosh:


src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -122,6 +123,17 @@ public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager clone
          return null;
       }
 
+      public void RemoveUsedObservedData(DataRepository dataRepository)
```
rwmcintosh, line 126: @georgeDaskalakis these guys need implementation in MoBi

msevestre:
@georgeDaskalakis I am going to merge so that I can fix my problem. The TODO remains

PR #809, opened 2022-11-25
Fixes #1733 Comparison of Expression Profiles

rwmcintosh:


src/MoBi.Core/CoreRegister.cs
```diff
@@ -92,11 +93,14 @@ public override void RegisterInContainer(IContainer container)
          registerCommitTasks(container);
 
          registerConverters(container);
+
+         // Temporary registration while Core components reside in MoBi
+         container.AddRegister(x => x.FromType<MoBiRegisterOfCore>());
```
msevestre, line 98: Good idea

msevestre (approved):
Works. Wondering about the implementation since most of it will be deleted? What's left at the end?

rwmcintosh:
> Works. Wondering about the implementation since most of it will be deleted? What's left at the end?

There will just be these 3 DiffBuilders all based on what is now called StartValuesDiffBuilder or something.

rwmcintosh:
> Works. Wondering about the implementation since most of it will be deleted? What's left at the end?

If you mean, why did you paste that code - it's because the StartValuesDiffBuilder is an internal class in Core, not protected or public

PR #827, opened 2022-12-08
Import expression profile as pkml

georgeDaskalakis:


Yuri05:
sbml or pkml? :)

rwmcintosh:
> sbml or pkml? :)

pkml

src/MoBi.Assets/ToolTips.cs
```diff
@@ -22,6 +22,11 @@ public static class BuildingBlockMolecule
          public static readonly string AddPKSimMolecule = "Add a PK-Sim Molecule";
       }
 
+      public static class BuildingBlockExpressionProfile
+      {
+         public static readonly string LoadExpressionProfile = $"Load Expression Profile from Expression Profile Building Block file (*{Constants.Filter.PKML_FILE_FILTER})";
```
msevestre, line 27: What is this tooltip? WE do not have everything like this

msevestre (commented):
Please create an issue to remember to undo the changes in appveyor file

src/MoBi.Assets/ToolTips.cs
msevestre: This tooltip look weird with the * etc. Not sure why we have this (or do we have this for other building blocks?)(

PR #833, opened 2022-12-13
Fixes #784 expression profile from pksim Fixes #842

### LINKED ISSUE #784: Create expression profile from PK-sim
You should see only the view of PK-sim where expression profiles are created

### LINKED ISSUE #842: Adding multiple expression profiles from through MoBi throws registration exception
@rwmcintosh commented on [Wed Dec 14 2022](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/2481)

rwmcintosh:


src/MoBi.Assets/AppConstants.cs
```diff
@@ -2068,6 +2071,11 @@ public static class PKSim
          public static readonly string PopulationSimulationArgument = "/pop";
          public static readonly string JournalFileArgument = "/j";
          public static readonly string NotInstalled = "PK-Sim was not found on current system. Please make sure that PK-Sim was installed using the provided setup. Alternatively, you can specify where PK-Sim is installed on your system under Utilities -> Options";
+
+         public static string EntryPointNotFound(string fileName)
+         {
+            return $"Could not find entry point {fileName}. Please make sure that PK-Sim was installed using the provided setup.";
```
msevestre, line 2077: We have something like this where we try to start PKSim already (start population workflow). Maybe we could reuse the wording instead (it also specifies that path can be set explicitly in the user settings)

src/MoBi.Core/Services/IPKSimStarter.cs
```diff
@@ -1,8 +1,13 @@
-﻿namespace MoBi.Core.Services
+﻿using OSPSuite.Core.Domain.Builder;
+
+namespace MoBi.Core.Services
 {
    public interface IPKSimStarter
    {
       void StartPopulationSimulationWithSimulationFile(string simulationFilePath);
       void StartWithWorkingJournalFile(string journalFilePath);
+      ExpressionProfileBuildingBlock CreateMetabolizingEnzymeExpression();
```
msevestre, line 9: I am wondering why these are 3 methods and not one with a parameter since we already have an enum etc?

src/MoBi.Presentation/Tasks/PKSimStarter.cs
```diff
@@ -31,9 +40,45 @@ public void StartWithWorkingJournalFile(string journalFilePath)
...
+      {
+         var assemblyFile = retrievePKSimUIStarterPath();
+
+         if (_externalAssembly == null)
+         {
+            _externalAssembly = Assembly.LoadFrom(assemblyFile);
+         }
+
+         if (_expressionCreatorType == null)
+            _expressionCreatorType = _externalAssembly.GetType("PKSim.UI.Starter.ExpressionProfileCreator");
```
msevestre, line 64: Can we define those strings as constant in the class a the top?

msevestre, line 64: why is this guy a member of the class? Should'it be transient?

src/MoBi.Presentation/UICommand/AddNewExpressionProfileBuildingBlock.cs
```diff
@@ -0,0 +1,73 @@
+﻿using MoBi.Core.Domain.Model;
+using MoBi.Core.Services;
+using MoBi.Presentation.Tasks.Interaction;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Presentation.UICommands;
+
+namespace MoBi.Presentation.UICommand
+{
+   public abstract class AddExpressionProfileBuildingBlock : ObjectUICommand<AddExpressionProfileBuildingBlock>
```
msevestre, line 9: yeah using the enum you would only have one command

src/MoBi.Presentation/UICommand/AddNewExpressionProfileBuildingBlock.cs
```diff
@@ -0,0 +1,73 @@
...
+      {
+         _pkSimStarter = pkSimStarter;
+         _context = context;
+         _interactionTasksForExpressionProfileBuildingBlock = interactionTasksForExpressionProfileBuildingBlock;
+      }
+
+      protected override void PerformExecute()
+      {
+         var expressionFromPKSim = CreateExpressionFromPKSim();
+         if (expressionFromPKSim != null)
```
msevestre, line 26: don't hesitate to exit early
https://github.com/Open-Systems-Pharmacology/Suite/blob/develop/CODING_STANDARDS.md#coding-style

src/MoBi.Presentation/UICommand/AddNewExpressionProfileBuildingBlock.cs
```diff
@@ -0,0 +1,73 @@
...
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Presentation.UICommands;
+
+namespace MoBi.Presentation.UICommand
+{
+   public abstract class AddExpressionProfileBuildingBlock : ObjectUICommand<AddExpressionProfileBuildingBlock>
+   {
+      protected IPKSimStarter _pkSimStarter;
+      private readonly IMoBiContext _context;
+      private readonly IInteractionTasksForExpressionProfileBuildingBlock _interactionTasksForExpressionProfileBuildingBlock;
```
msevestre, line 13: we typically call the parameter `interactionTask`

rwmcintosh:
Also fixes #842

PR #846, opened 2022-12-20
821 drag and drop same obs data on chart duplicates mapping

### LINKED ISSUE #821: Drag and drop same obs data on chart duplicates mapping
image
I believe this is NOT by design
Using project from #800

IMO, obs data should only be added once to the list

Thoughts? @rwmcintosh @Yuri05 @PavelBal @georgeDaskalakis ?

georgeDaskalakis:
closes #821

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
@@ -136,7 +136,7 @@ namespace MoBi.Presentation.Presenter
          UpdateCaption();
          _view.Display();
          loadChart();
-     }
+      }
 
       private void addObservedDataRepositories(IList<DataRepository> data, IEnumerable<Curve> curves)
       {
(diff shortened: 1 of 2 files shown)


src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
```diff
@@ -328,11 +328,23 @@ private bool canHandle(IObjectBaseEvent objectBaseEvent)
...
             _outputMappingMatchingTask.AddMatchingOutputMapping(dataRepository, _simulation);
             _context.PublishEvent(new ObservedDataAddedToAnalysableEvent(_simulation, dataRepository, false));
          }
       }
+
+      private List<DataRepository> observedDataNotMappedInTheSimulation(ObservedDataAddedToChartEventArgs e)
+      {
+         var observedDataToAdd = e.AddedDataRepositories
+            .Where(dataRepository => !_simulation.OutputMappings.Any(x => x.UsesObservedData(dataRepository))).ToList();
+         return observedDataToAdd;
```
rwmcintosh, line 347: You can just return right away without the variable

rwmcintosh (approved):

src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
georgeDaskalakis: true. corrected.

georgeDaskalakis:
@msevestre if no further comments, I am going to merge

PR #848, opened 2022-12-20
847 cannot delete observed data fixed

### LINKED ISSUE #847: Cannot delete observed data from simulaiton OutputMappings tab
The newly added way of deleting observed data (from the "x" in the "Observed Data" tab in a Simulation ) is not functioning in MoBi.

georgeDaskalakis:
closes #847

### DIFF
--- a/src/MoBi.Core/Domain/Model/MoBiSimulation.cs
+++ b/src/MoBi.Core/Domain/Model/MoBiSimulation.cs
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Linq;
+using FluentNHibernate.Conventions;
 using OSPSuite.Core.Chart;
 using OSPSuite.Core.Chart.Simulations;
 using OSPSuite.Core.Diagram;
@@ -134,11 +135,11 @@ namespace MoBi.Core.Domain.Model
          if (!UsesObservedData(dataRepository))
             return;
 
-         var curveToRemove = Chart.Curves.FirstOrDefault(c => Equals(c.yData.Repository, dataRepository));
-         if (curveToRemove == null)
+         var curveToRemove = Chart.Curves.Where(c => Equals(c.yData.Repository, dataRepository)).ToList();
+         if (curveToRemove.IsEmpty())
             return;
 
-         Chart.RemoveCurve(curveToRemove.Id);
+         curveToRemove.ForEach(curve => Chart.RemoveCurve(curve.Id) );
 
          HasChanged = true;
       }
--- a/src/MoBi.Core/Domain/Model/MoBiSimulation.cs
+++ b/src/MoBi.Core/Domain/Model/MoBiSimulation.cs
@@ -1,6 +1,5 @@
 using System.Collections.Generic;
 using System.Linq;
-using FluentNHibernate.Conventions;
 using OSPSuite.Core.Chart;
 using OSPSuite.Core.Chart.Simulations;
 using OSPSuite.Core.Diagram;
@@ -136,7 +135,7 @@ namespace MoBi.Core.Domain.Model
             return;
 
          var curveToRemove = Chart.Curves.Where(c => Equals(c.yData.Repository, dataRepository)).ToList();
-         if (curveToRemove.IsEmpty())
+         if (!curveToRemove.Any())
             return;
 
          curveToRemove.ForEach(curve => Chart.RemoveCurve(curve.Id) );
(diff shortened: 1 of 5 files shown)


src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -131,13 +131,22 @@ public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager clone
 
       public void RemoveUsedObservedData(DataRepository dataRepository)
       {
-         
+         if (!UsesObservedData(dataRepository))
+            return;
 
+         var curveToRemove = Chart.Curves.FirstOrDefault(c => Equals(c.yData.Repository, dataRepository));
```
msevestre, line 137: Why First or default ?why are you assuming only one curve belongs to the repository? I think this is wrong

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -354,5 +355,10 @@ public override void ReleaseFrom(IEventPublisher eventPublisher)
          displayPresenter.DragOver -= OnDragOver;
          Clear();
       }
+
+      public void Handle(ObservedDataRemovedFromAnalysableEvent e)
+      {
+         Refresh();
```
msevestre, line 361: hum. you will refresh all the time even if the chart has nothing to do with the event. I think this is not ok

msevestre (changes requested):

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
georgeDaskalakis: true, was assuming one curve because of one dataRepository. Fixed to work for multiple.

georgeDaskalakis:
@msevestre this is ready for review again

src/MoBi.Presentation/Presenter/ChartPresenter.cs
```diff
@@ -23,6 +23,7 @@
 using OSPSuite.Utility.Collections;
 using OSPSuite.Utility.Events;
 using OSPSuite.Utility.Extensions;
+using static OSPSuite.Assets.Captions;
```
Yuri05, line 26: why this?
(wondering because there are no other code changes in the file)

georgeDaskalakis: true, this slipped my attention - just something that was left there. Removed.

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Linq;
+using FluentNHibernate.Conventions;
```
msevestre, line 3: Please remove. This has nothing to do in s simulation file

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -131,13 +132,22 @@ public override void UpdatePropertiesFrom(IUpdatable source, ICloneManager clone
...
       {
-         
+         if (!UsesObservedData(dataRepository))
+            return;
 
+         var curveToRemove = Chart.Curves.Where(c => Equals(c.yData.Repository, dataRepository)).ToList();
+         if (curveToRemove.IsEmpty())
+            return;
+
+         curveToRemove.ForEach(curve => Chart.RemoveCurve(curve.Id) );
```
msevestre, line 142: Our method is each

msevestre (changes requested):
Looks good aside from.the nhibernate import which is not acceptable

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
georgeDaskalakis: removed

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
msevestre: Each not ForEach

msevestre (changes requested):

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
georgeDaskalakis: @msevestre the nhibernate was actually because of the .IsEmpty on the list. .ForEach comes from the ms list

msevestre (approved):

Issue #783, opened 2022-11-02
Create individual from PK-Sim


rwmcintosh:
You should see only the view of PK-sim where individuals are created

PR #852, opened 2023-01-04
Fixes #645 Individual Management

### LINKED ISSUE #645: Individual Management
* Create
* Delete
* Rename

All of these actions are using Commands.

rwmcintosh:
The only thing missing is Adding a new individual. That will be covered by a separate issue because the only way to create a new one is from #783


src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForIndividualBuildingBlock.cs
```diff
@@ -0,0 +1,13 @@
+﻿using MoBi.Core.Domain.Model;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Services;
+
+namespace MoBi.Presentation.MenusAndBars.ContextMenus
```
rwmcintosh, line 5: This is the only code. The interaction tasks were already in place.

PR #861, opened 2023-01-17
Fixes #860 Slow performance when scrolling start values

### LINKED ISSUE #860: Slow performance when scrolling start values
If there are a lot of start value objects then the scrolling performance is very poor.

The reason is that validation is occurring for newly painted rows. The validation for a start value DTO includes checking that it's got a unique path among all DTO's so, more DTO's mean this rule takes longer to validate.

﻿﻿https://supportcenter.devexpress.com/ticket/details/q361417/gridview-scrolling-slow

rwmcintosh:


tests/MoBi.Tests/Presentation/MoleculeStartValueDTOSpecs.cs
```diff
@@ -20,26 +20,6 @@ protected override void Context()
       }
    }
 
-   public class When_validating_a_start_value_with_proposed_path_element_removed_from_end : concern_for_MoleculeStartValueDTO
```
rwmcintosh, line 23: This test becomes invalid because it tests if the building block is valid by construction rather than through a user edit

rwmcintosh: image

The test still works if triggered through a user edit

rwmcintosh:
@PavelBal FYI

msevestre (approved):
Lol. This check was terrible
Well done.
Perfect comments too

PR #865, opened 2023-01-19
Fixes #840 Put PKSim Version in views

### LINKED ISSUE #840: Put PKSim Version in views

rwmcintosh:


rwmcintosh:
Little different implementations due to how we are displaying this read-only data from PK-Sim.

PK-Sim Version is not really 'Origin Data', but I didn't think this display warranted another layout control group for just one text box.

image

image

msevestre (approved):

Issue #838, opened 2022-12-14
Cloning and renaming the cloned Expression Profile


rwmcintosh:
Naming is important. 
- Naming to be done field by field
- Allow renaming enzyme but change moleculename for all parameters to match. 
- Append 'Cloned' to the category eg. "AADAC|Human|den Holländer 1992 ind1 male 0.92y 6.5kg - Cloned"

PR #869, opened 2023-01-30
Fixes ﻿#838 Rename and cloned rename working

rwmcintosh:
For clone rename or just rename, we are using this new dialog for expression profiles. You don't rename an expression profiles, instead you have to change the individual fields that comprise the name

image

As you type, the readonly 'Name' is updated in real time, as is the validation indicator.

image


src/MoBi.Presentation/Tasks/Interaction/ObjectBaseTask.cs
```diff
@@ -1,128 +0,0 @@
-﻿using System.Collections.Generic;
```
rwmcintosh, line 1: A lot of the changed files here are because I removed this task. It mostly renamed objects and up to now, all the renaming was completely common for all building blocks.

I needed to rename building blocks now in a type-sensitive way because the renaming presenter for Expression Profile would be different from most other cases.

Ultimately, this code was added to one of the base 'Interaction' tasks rather than delegated to this type-insensitive class.

There were also a couple of dependencies that needed to be injected as a result.

src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
```diff
@@ -66,7 +75,89 @@ public virtual void Edit(T objectToEdit)
       public virtual void Rename(T objectBase, IEnumerable<IObjectBase> existingObjectsInParent, IBuildingBlock buildingBlock)
       {
          var forbiddenNames = GetForbiddenNames(objectBase, existingObjectsInParent);
-         _context.AddToHistory(_interactionTask.Rename(objectBase, forbiddenNames, buildingBlock));
+         _context.AddToHistory(rename(objectBase, forbiddenNames, buildingBlock));
```
rwmcintosh, line 78: Here now instead of calling _interactionTask, which calls ObjectBaseTask, we can just do all the renaming here because this is a generic class.

src/MoBi.Presentation/Tasks/Interaction/InteractionTask.cs
```diff
@@ -76,11 +73,6 @@ public class InteractionTask : IInteractionTask
          _serializationTask.SaveModelPart(entityToSerialize, fileName);
       }
 
-      public IMoBiCommand Rename<T>(T objectBase, IEnumerable<string> forbiddenNames, IBuildingBlock buildingBlock) where T : IObjectBase
```
rwmcintosh, line 79: Previously, InteractionTask.Rename just called ObjectBaseTask.Rename

src/MoBi.Core/Commands/RenameExpressionProfileBuildingBlockCommand.cs
```diff
@@ -0,0 +1,48 @@
...
+      public override void RestoreExecutionData(IMoBiContext context)
+      {
+         base.RestoreExecutionData(context);
+         _expressionProfileBuildingBlock = context.Get<ExpressionProfileBuildingBlock>(ObjectId);
+      }
+
+      protected override void RenameBuildingBlock(IMoBiContext context)
+      {
+         base.RenameBuildingBlock(context);
+         var objectPathFactory = context.ObjectPathFactory;
```
msevestre, line 39: I'd suggest to add a comment here4 as this is not quite obvious why you need to mess around with path

src/MoBi.Core/Commands/RenameExpressionProfileBuildingBlockCommand.cs
```diff
@@ -0,0 +1,48 @@
...
+         _expressionProfileBuildingBlock.Each(x =>
+         {
+            var objectPath = x.Path;
+            objectPath.Replace(_oldMoleculeName, _expressionProfileBuildingBlock.MoleculeName);
+            x.Path = objectPathFactory.CreateObjectPathFrom(objectPath);
```
msevestre, line 44: yeah this 3 lines of code feel really weird. You replace in the reference (same object) and then create a new entry ?

src/MoBi.Core/Commands/RenameObjectBaseCommand.cs
```diff
@@ -39,15 +39,20 @@ protected override void ClearReferences()
       protected override void ExecuteWith(IMoBiContext context)
       {
          base.ExecuteWith(context);
+         RenameBuildingBlock(context);
```
msevestre, line 42: this feels weird. Why a RenameBuildingBlock method in a generic objectBaseCommand

src/MoBi.Presentation/DTO/RenameExpressionProfileDTO.cs
```diff
@@ -0,0 +1,103 @@
...
+         {
+            _species = value;
+            OnPropertyChanged(() => Name);
+            OnPropertyChanged(() => Species);
+         }
+      }
+
+      private static class RenameExpressionProfileDTORules
+      {
+         public static IEnumerable<IBusinessRule> NonEmptyFieldRules()
```
msevestre, line 64: change ths to be initalized once and not all the time.
Also return IReadOnly

src/MoBi.Presentation/DTO/RenameExpressionProfileDTO.cs
```diff
@@ -0,0 +1,103 @@
...
+         {
+            yield return createNonEmptyRule(item => item.Species, Assets.AppConstants.Validation.SpeciesCannotBeEmpty);
+            yield return createNonEmptyRule(item => item.Category, Assets.AppConstants.Validation.CategoryCannotBeEmpty);
+            yield return createNonEmptyRule(item => item.MoleculeName, Assets.AppConstants.Validation.MoleculeNameCannotBeEmpty);
+         }
+
+         private static IBusinessRule createNonEmptyRule(Expression<Func<RenameExpressionProfileDTO, string>> propertyToCheck, string errorCaption)
```
msevestre, line 71: change those method to use a get so that it's created only once and not every time

src/MoBi.Presentation/DTO/RenameExpressionProfileDTO.cs
```diff
@@ -0,0 +1,103 @@
...
+
+         public static IBusinessRule UniqueName { get; } = CreateRule.For<RenameExpressionProfileDTO>()
+            .Property(item => item.Name)
+            .WithRule((dto, newName) => dto.isNameUnique(newName))
+            .WithError((dto, newName) => dto.nameAlreadyExistsError(newName));
+      }
+
+      private string nameAlreadyExistsError(string newName)
+      {
+         return Error.NameAlreadyExists(newName);
```
msevestre, line 87: hum.. not sure about the gain of a method to call the same method somewhere else

src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
msevestre: what's the reason of doing this? You are adding dependencies in all derives class. There is probably a good reasons for this but I don't see it

src/MoBi.UI/Views/RenameExpressionProfileBuildingBlockView.cs
```diff
@@ -0,0 +1,75 @@
...
+      {
+         _screenBinder.Bind(x => x.Category).To(tbCategory);
+         _screenBinder.Bind(x => x.MoleculeName).To(tbMoleculeName);
+         _screenBinder.Bind(x => x.Species).To(tbSpecies);
+         _screenBinder.Bind(x => x.Name).To(tbName);
+         _screenBinder.SavingMode = SavingMode.Always;
+
+         // We need to validate the text box when the text changes so that the Name property is validated
+         // even thought that text box is readonly. The Name is composed of each of the other text boxes content
+         tbCategory.TextChanged += (o,e) => OnEvent(() => validateTextBox(o));
```
msevestre, line 38: I don't get this. why is this required? Isn't it what the property changes is all about?

msevestre (changes requested):
See comments in PR

src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
rwmcintosh: It could be done another way.

What we had before was an ObjectBaseTask that did only this renaming in the same way for all objects. That dependency was part of InteractionTask which was part of InteractionTaskContext.

These two general dependencies could go in InteractionTaskContext so that not so many classes have direct dependency on these objects, only InteractionTaskContext would have to change.

I'll move it around and see how that fits.

src/MoBi.UI/Views/RenameExpressionProfileBuildingBlockView.cs
rwmcintosh: That's only updated when the field loses focus. To get this feature working the way I liked, I wanted to run validation on every keystroke.

tests/MoBi.Tests/Presentation/Tasks/InteractionTasksForRootApplicationBuilderSpecs.cs
```diff
@@ -52,7 +50,7 @@ protected override void Because()
       [Observation]
       public void should_not_return_builder_name_as_forbidden()
       {
-         _result.ShouldOnlyContain(_builderName);
+         _result.ShouldNotContain(_builderName);
```
rwmcintosh, line 53: Trying to infer this test from the name and content. It feels like the original test was just using the wrong assertion.

src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
rwmcintosh: It's nicer this way. The dependencies are still the dependencies, but now these ones are bundled into InteractionTaskContext instead of in the constructors of all these EditTasks.

src/MoBi.UI/Views/RenameExpressionProfileBuildingBlockView.cs
rwmcintosh: I should say that the validation for that field already runs on each keystroke, but the underlying field is only updated on lost focus.

I want to update the underlying field and validate the 'Name' property on every keystroke on any one of these text boxes.

msevestre: This should work out of the box. Not a big deal but if you have your DTO raising the right changed event...it happens automatically like for all other boxes.

msevestre (approved):

PR #870, opened 2023-01-31
783 create individual from pk sim

### LINKED ISSUE #783: Create individual from PK-Sim

georgeDaskalakis:
closes #783


rwmcintosh (approved):

src/MoBi.Presentation/Tasks/PKSimStarter.cs
```diff
@@ -86,6 +105,11 @@ private MethodInfo getExpressionCreatorMethod(string methodName)
          return _externalAssembly.GetType(PKSIM_UI_STARTER_EXPRESSION_PROFILE_CREATOR).GetMethod(methodName);
       }
 
+      private MethodInfo getIndividualCreatorMethod(string methodName)
+      {
+         return _externalAssembly.GetType(PKSIM_UI_STARTER_INDIVIDUAL_CREATOR).GetMethod(methodName);
```
msevestre, line 110: Is there a reason to have one starter for each? Cannot we have one single entry point that stats the right stuff based on the method? Because if feels like we have one class with one method for each. Soemhow, there is one too many

src/MoBi.Presentation/UICommand/AddNewIndividualCommand.cs
```diff
@@ -0,0 +1,33 @@
...
+         IInteractionTasksForIndividualBuildingBlock interactionTask)
+      {
+         _pkSimStarter = pkSimStarter;
+         _context = context;
+         _interactionTask = interactionTask;
+      }
+
+      protected override void PerformExecute()
+      {
+         var individualFromPKSim = _pkSimStarter.CreateIndividual();
```
msevestre, line 25: this can probably also be generic completely and use the type. 
pksimStarter.Create<Individual> etc.. not a lof of code so that's ok but this will be duplicated EXACTLY For each type we add. Not sure if we plan to load more from PKSim ( I could see molecule being one of them, events etc.. So making it reusable would make sense IMO

msevestre (commented):
I think this works but there is a lot of `almost the same` that can {(I believe, please correct me if I am wrong) be refactored so that we can add many more PKSim type easily

src/MoBi.Presentation/Tasks/PKSimStarter.cs
georgeDaskalakis: I agree. I suggest we do this refactoring in a new PR. I have created a separate issue for this:

https://github.com/Open-Systems-Pharmacology/MoBi/issues/882

since for the issue I am currently worrking on ( https://github.com/Open-Systems-Pharmacology/MoBi/issues/839 ) we will need to make the class even more extendable - to be able to pass furtherr arguments than just the IShell fe

georgeDaskalakis:
@msevestre this was refactored accortding to your suggestions to become more generic - feel free to review it again.

Issue #823, opened 2022-12-07
Implement context menu entries for each building block type


rwmcintosh:
This relies on completion of #817

rwmcintosh:
This should include adding building blocks to the module

PR #877, opened 2023-02-07
Fixes #816 View and Presenter for the Module Explorer

### LINKED ISSUE #816: View and Presenter for the Module Explorer
We decided to try using a nested structure provided by grid view to implement the module explorer view.

That gives the user more flexibility in how they will interact with modules.

By default, the grid should organize under each 'module' collecting the building blocks associated with the module.

Users could decide to organize the view differently, showing the building blocks organized by type, therefore showing all 'Spatial Structures' together with another column dedicated to the containing module.

rwmcintosh:
I implemented this default module feature similar to how new building blocks are added to blank projects. In this screenshot I was adding 2 PSV and 1 MSV just to get the folder logic correct.

image

We get the context menus, but some entries will have to be changed around. I'm going to take that task on next with #823 .

image

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -14,7 +14,7 @@ namespace MoBi.Assets
 {
    public static class AppConstants
    {
-      public static readonly int LayoutVersion = 24;
+      public static readonly int LayoutVersion = 25;
       public static readonly string NotMatch = "not tagged with";
       public static readonly string Match = "tagged with";
       public static readonly string MatchAll = "in all containers";
@@ -88,6 +88,7 @@ namespace MoBi.Assets
       public static class DefaultNames
       {
          public static readonly string MoleculeBuildingBlock = "Molecules";
+         public static readonly string Module1 = "Module1";
          public static readonly string ReactionBuildingBlock = "Reaction";
          public static readonly string SpatialStructure = "Organism";
          public static readonly string PassiveTransportBuildingBlock = "Passive Transports";
@@ -997,6 +998,7 @@ namespace MoBi.Assets
          public static readonly string ExportHistory = Captions.ExportHistory;
          public static readonly string StartPopulationSimulation = "Send Simulation to PK-Sim for Population Simulation...";
          public static readonly string BuildingBlockExplorer = "Building Blocks";
+         public static readonly string ModuleExplorer = "Modules";
          public static readonly string SimulationExplorer = "Simulations";
          public static readonly string New = "New";
          public static readonly string NewMolecule = AddNew(ObjectTypes.Molecule);
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -88,7 +88,7 @@ namespace MoBi.Assets
       public static class DefaultNames
       {
          public static readonly string MoleculeBuildingBlock = "Molecules";
-         public static readonly string Module1 = "Module1";
+         public static readonly string Module = "Module";
          public static readonly string ReactionBuildingBlock = "Reaction";
          public static readonly string SpatialStructure = "Organism";
          public static readonly string PassiveTransportBuildingBlock = "Passive Transports";
--- a/src/MoBi.Presentation/Presenter/Main/MenuAndToolBarPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/Main/MenuAndToolBarPresenter.cs
@@ -173,7 +173,7 @@ namespace MoBi.Presentation.Presenter.Main
          _menuBarItemRepository[MenuBarItemIds.NotificationView].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.ComparisonView].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.OpenSimulation].Enabled = true;
-         _menuBarItemRepository[MenuBarItemIds.BuildingBlockExplorerView].Enabled = true;
+         _menuBarItemRepository[MenuBarItemIds.ModuleExplorerView].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.SimulationExplorerView].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.GarbageCollection].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.Exit].Enabled = true;
--- a/src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
@@ -1,0 +1,168 @@
+using System.Collections.Generic;
+using MoBi.Core.Domain.Model;
+using MoBi.Presentation.DTO;
+using MoBi.Presentation.Nodes;
+using OSPSuite.Assets;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Presentation.Core;
+using OSPSuite.Presentation.Nodes;
+using OSPSuite.Presentation.Presenters;
+using OSPSuite.Presentation.Presenters.Classifications;
+using OSPSuite.Presentation.Presenters.ContextMenus;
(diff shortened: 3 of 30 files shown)


src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
```diff
@@ -0,0 +1,168 @@
+using System.Collections.Generic;
```
rwmcintosh, line 1: The module view and presenter is largely copied from BB view and presenter. Refactoring common code didn't make sense because I think the BB explorer will disappear. I opted to keep it for now so I can observe behaviours as I implement the context menu commands.

src/MoBi.Presentation/UICommand/ShowModuleExplorerCommand.cs
```diff
@@ -3,17 +3,20 @@
...
-      public ShowBuildingBlockExplorerCommand(IBuildingBlockExplorerPresenter buildingBlockExplorerPresenter)
+      public ShowModuleExplorerCommand(IBuildingBlockExplorerPresenter buildingBlockExplorerPresenter, IModuleExplorerPresenter moduleExplorerPresenter)
       {
          _buildingBlockExplorerPresenter = buildingBlockExplorerPresenter;
+         _moduleExplorerPresenter = moduleExplorerPresenter;
       }
 
       public void Execute()
       {
+         _moduleExplorerPresenter.ToggleVisibility();
```
rwmcintosh, line 19: I'm just paralleling these toggles for now. Eventually BB explorer disappears.

src/MoBi.Assets/AppConstants.cs
```diff
@@ -14,7 +14,7 @@ namespace MoBi.Assets
 {
    public static class AppConstants
    {
-      public static readonly int LayoutVersion = 24;
+      public static readonly int LayoutVersion = 25;
```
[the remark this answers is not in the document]
msevestre: oh yeah otherwuse it won't work for other people either

src/MoBi.Assets/AppConstants.cs
```diff
@@ -88,6 +88,7 @@ public static class Organs
       public static class DefaultNames
       {
          public static readonly string MoleculeBuildingBlock = "Molecules";
+         public static readonly string Module1 = "Module1";
```
msevestre, line 91: module1?

msevestre: The variable should at least be called Module not Module1. Thjis is not the winform designer here

src/MoBi.Presentation/Presenter/Main/MenuAndToolBarPresenter.cs
```diff
@@ -173,7 +173,7 @@ private void enableDefaultItems()
          _menuBarItemRepository[MenuBarItemIds.NotificationView].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.ComparisonView].Enabled = true;
          _menuBarItemRepository[MenuBarItemIds.OpenSimulation].Enabled = true;
-         _menuBarItemRepository[MenuBarItemIds.BuildingBlockExplorerView].Enabled = true;
```
msevestre, line 176: Are you actually removing the building block explorer? Mght be good to keep it around for now so that we can see what happens when loading older projects

src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
```diff
@@ -0,0 +1,168 @@
...
+namespace MoBi.Presentation.Presenter.Main
+{
+   public interface IModuleExplorerPresenter : IExplorerPresenter, IPresenter<IModuleExplorerView>
+   {
+      bool ShouldSort(ITreeNode node);
+   }
+
+   public class ModuleExplorerPresenter : ExplorerPresenter<IModuleExplorerView, IModuleExplorerPresenter>, IModuleExplorerPresenter
+   {
+      private readonly IObservedDataInExplorerPresenter _observedDataInExplorerPresenter;
```
msevestre, line 31: Observed data part of Module? This is a question

src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
```diff
@@ -0,0 +1,168 @@
...
+      protected override void AddProjectToTree(IMoBiProject project)
+      {
+         using (new BatchUpdate(_view))
+         {
+            _view.DestroyNodes();
+
+            project.Modules.Each(addModule);
+
+            _view.AddNode(_treeNodeFactory.CreateFor(MoBiRootNodeTypes.ExpressionProfilesFolder));
+            _view.AddNode(_treeNodeFactory.CreateFor(MoBiRootNodeTypes.IndividualsFolder));
```
msevestre, line 131: why that?  Shouldn't that be outside of the module explorer?

src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
```diff
@@ -0,0 +1,168 @@
...
+         module.ParameterStartValuesCollection.Each(bb => addBuildingBlockUnderNode(bb, parameterStartValuesCollectionNode));
+      }
+
+      private ITreeNode collectionNodeFor<T>(IReadOnlyList<IStartValuesBuildingBlock<T>> startValueBlockCollection, RootNodeType rootNodeType, ITreeNode moduleNode) where T : class, IStartValue
+      {
+         return startValueBlockCollection.Count > 1 ? _view.AddNode(_treeNodeFactory.CreateFor(rootNodeType).Under(moduleNode)) : moduleNode;
+      }
+   }
+
+   public interface IModuleExplorerView : IExplorerView, IView<IModuleExplorerPresenter>
```
msevestre, line 165: let's m ove this in the view namspace

src/MoBi.Presentation/Tasks/ProjectTask.cs
```diff
@@ -275,33 +276,47 @@ private bool loadProjectIntoContext(string fileName)
...
 
       private void generateDefaultsInCurrentProject()
       {
-         addDefault<IMoleculeBuildingBlock>(AppConstants.DefaultNames.MoleculeBuildingBlock);
-         addDefault(AppConstants.DefaultNames.ReactionBuildingBlock, () => _reactionBuildingBlockFactory.Create());
-         addDefault(AppConstants.DefaultNames.SpatialStructure, () => _spatialStructureFactory.CreateDefault(AppConstants.DefaultNames.SpatialStructure));
-         addDefault<IPassiveTransportBuildingBlock>(AppConstants.DefaultNames.PassiveTransportBuildingBlock);
-         addDefault<IEventGroupBuildingBlock>(AppConstants.DefaultNames.EventBuildingBlock);
-         addDefault<IObserverBuildingBlock>(AppConstants.DefaultNames.ObserverBuildingBlock);
+         var module = _context.Create<Module>();
```
msevestre, line 287: Why is the default with everythung in it?

msevestre (changes requested):
I think this looks very good already but I am wondering about some of the layout decisions here

src/MoBi.Presentation/Presenter/Main/MenuAndToolBarPresenter.cs
rwmcintosh: I have them both still, but I expect that before release, we will remove the bb explorer.

src/MoBi.Presentation/Presenter/Main/ModuleExplorerPresenter.cs
rwmcintosh: Do we have a separate explorer view for ObservedData, Individuals and Expressions? I didn't think so. They appear in this view, but are not part of a module domain object.

That was my understanding.

src/MoBi.Presentation/Tasks/ProjectTask.cs
rwmcintosh: I think the workflow here is the same as 'New Project' with Building Blocks. You need at least one of each type to make anything right?

That's the flow I tried to preserve.

rwmcintosh:
Issues opened to address rework based on feedback in meetings today.

PR #883, opened 2023-02-13
Some user interface changes based on demo

### LINKED ISSUE #817: New Module...
When creating a new module, there must be a way to kick off the creation. Maybe in the ribbon somewhere...

A window where the user can select options should be shown.

- At most one bb per type.

image
That's not bad, but do not require users to select 'No Entry' all the time

We have this concept of in/out selection like population analyses
image

### LINKED ISSUE #879: Create new project should not include a default module
Instead we can just populate RootNodes for PKSim Modules and Extension Modules where a user can get access to a context menu to create or import a new module.

### LINKED ISSUE #880: Root nodes for extension and pk-sim modules
Instead of modules having a top-level node for each module, we need to group them by PKSim or Extension types.

Create root nodes for these and list each module under this folder

rwmcintosh:
Fixes #880 Root nodes for extension and pk-sim modules
Fixes #879 Create new project should not include a default module 
Fixes #817 New Module with building blocks

The context menu creation only applies to the Extension modules because PKSim Modules can't be added this way.

image

image


rwmcintosh:
Some of this context menu implementation will come later. Eg. you cannot 'Load' yet, there is no serialization for module.

image

src/MoBi.Assets/AppConstants.cs
```diff
@@ -1074,7 +1078,9 @@ public static class MenuNames
 
          public static string AddExistingFromTemplate(string objectTypeName) => $"Load {objectTypeName} from Template...";
 
-       
+         public static string AddEmptyModule() => $"Create Empty Module...";
```
msevestre, line 1081: Why is this a method?

src/MoBi.Assets/AppConstants.cs
```diff
@@ -1074,7 +1078,9 @@ public static class MenuNames
...
+         public static string AddModuleWithBuildingBlocks() => $"Create Module with Building Blocks...";
+
```
msevestre, line 1083: Same question

src/MoBi.Core/Commands/AddModuleCommand.cs
```diff
@@ -0,0 +1,60 @@
...
+         ModuleId = module.Id;
+         Description = AppConstants.Commands.AddToProjectDescription(ObjectType, module.Name);
+         Silent = false;
+      }
+
+      protected override void ExecuteWith(IMoBiContext context)
+      {
+         var project = context.CurrentProject;
+         context.Register(_module);
+         _module.AllBuildingBlocks().Each(context.Register);
```
msevestre, line 32: I think registering a module should register the building block (visitor implementation). Thoughts?

src/MoBi.Core/Commands/RemoveModuleCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+         Description = AppConstants.Commands.RemoveFromProjectDescription(ObjectType, module.Name);
+         _module = module;
+      }
+
+      protected override void ExecuteWith(IMoBiContext context)
+      {
+         var project = context.CurrentProject;
+         removeFromProject(project);
+         context.Unregister(_module);
+         _module.AllBuildingBlocks().Each(context.Unregister);
```
msevestre, line 30: Same here. This should happen automatically

src/MoBi.Core/Domain/Extensions/ModuleExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using System.Collections.Generic;
+using System.Linq;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+
+namespace MoBi.Core.Domain.Extensions
+{
+   public static class ModuleExtensions
+   {
+      public static IReadOnlyList<IBuildingBlock> AllBuildingBlocks(this Module module)
```
msevestre, line 10: Why not a method of the module?

src/MoBi.Presentation/DTO/CreateModuleDTO.cs
```diff
@@ -0,0 +1,57 @@
...
+      {
+         if (_prohibitedNames == null)
+            return true;
+
+         return !_prohibitedNames.Contains(newName.Trim());
+      }
+      
+      private static class ModuleCreationRules
+      {
+         public static IBusinessRule NotEmptyName { get; } = CreateRule.For<CreateModuleDTO>()
```
msevestre, line 46: That rule already exists a a generic rule.

src/MoBi.Presentation/DTO/CreateModuleDTO.cs
```diff
@@ -0,0 +1,57 @@
...
+            .Property(item => item.Name)
+            .WithRule((dto, newName) => !string.IsNullOrEmpty(newName))
+            .WithError((dto, newName) => AppConstants.Validation.ModuleNameCannotBeEmpty);
+         
+         public static IBusinessRule UniqueName { get; } = CreateRule.For<CreateModuleDTO>()
+            .Property(item => item.Name)
```
msevestre, line 52: This one too probably

msevestre (changes requested):
Still working on it but some feedback already

src/MoBi.Presentation/DTO/CreateModuleDTO.cs
rwmcintosh: I found one for non-blank name, but I don't see this one as a generic rule anywhere. Do you know where it might be?

msevestre: @rwmcintosh Hum not sure. Maybe it's in MoBi directly. We always have this concept of unique names etc.. Is there a BuidlgnBlockDTO. I think it's ok to duplicate but it's probably already there 5 times

rwmcintosh: There are a couple of sort of generic ones. Like if you have an objectBaseDTO you can use the generic ones for ObjectBaseDTO. But this DTO doesn't have anything in common with ObjectBaseDTO. It's just a name and some bool properties. I'll put a generic rule in Core with the next commit.

rwmcintosh:
@msevestre I'm going to merge it. If you take another look and see something let me know. I can work it into a later commit or open another issue.

PR #886, opened 2023-02-14
Fixes #885 save modules to project

### LINKED ISSUE #885: Save Modules with Project

rwmcintosh:


src/MoBi.UI/MoBi.UI.csproj
```diff
@@ -47,7 +47,25 @@
   </ItemGroup>
 
   <ItemGroup>
+    <Reference Include="Accessibility">
```
msevestre, line 50: Why do we need those DLLs?

msevestre: If we do, we need to create a new package for our devepxress nuget stuff

src/MoBi.UI/MoBi.UI.csproj
```diff
@@ -47,7 +47,25 @@
...
+      <HintPath>..\..\..\..\..\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\System.Data.Linq\v4.0_4.0.0.0__b77a5c561934e089\System.Data.Linq.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Deployment">
+      <HintPath>..\..\..\..\..\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\System.Deployment\v4.0_4.0.0.0__b03f5f7f11d50a3a\System.Deployment.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Runtime.Serialization.Formatters.Soap">
+      <HintPath>..\..\..\..\..\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\System.Runtime.Serialization.Formatters.Soap\v4.0_4.0.0.0__b03f5f7f11d50a3a\System.Runtime.Serialization.Formatters.Soap.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Security">
+      <HintPath>..\..\..\..\..\WINDOWS\Microsoft.Net\assembly\GAC_MSIL\System.Security\v4.0_4.0.0.0__b03f5f7f11d50a3a\System.Security.dll</HintPath>
```
msevestre, line 67: This is all very weird. Why are those DLLs added

rwmcintosh: oh geez. I did not look carefully at the csproj. I'll back that out.

msevestre (changes requested):
I am worried about all those dependencies added. Can you explain why we need those coming from GAC and not from the framework? Feels like other deps are linked differently

rwmcintosh:
I don't know how it got added, but I didn't check the csproj files before committing (or even after). This is carelessness.

rwmcintosh:
Fixed

PR #894, opened 2023-02-17
Fixes #893 add in parent condition

### LINKED ISSUE #893: Update MoBi to use updated version of core with InParent condition

msevestre:


src/MoBi.Core/MoBiConfiguration.cs
```diff
@@ -26,7 +26,7 @@ public class MoBiConfiguration : OSPSuiteConfiguration, IMoBiConfiguration
       public override Origin Product { get; } = Origins.MoBi;
       public override string ProductNameWithTrademark { get; } = AppConstants.PRODUCT_NAME_WITH_TRADEMARK;
       public override string IconName { get; } = ApplicationIcons.MoBi.IconName;
-      protected override string[] LatestVersionWithOtherMajor { get; } = {"7.4", "6.3", "3.6" };
+      protected override string[] LatestVersionWithOtherMajor { get; } = {"11.1", "11.0", "7.4", "6.3", "3.6" };
```
Yuri05, line 29: I forgot: what was this const meant for?
Seems we did not update it for a while 🤔

Yuri05 (approved):

src/MoBi.Core/MoBiConfiguration.cs
msevestre: where to look for to get the settings. If you don't update this, it resets all default
So 12 should try to load settings from 11.1 for instance

Yuri05: ok

PR #903, opened 2023-02-17
Fixes #896 add parent property

### LINKED ISSUE #896: Spatial Structure: Add Parent property to the container
This is, by default set to the actual parent in the hierarchy (read-only). 
This is an empty entry for the root node of the spatial structure (e.g. Organism)

msevestre:


src/MoBi.UI/Extensions/LayoutItemExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using DevExpress.LookAndFeel;
+using DevExpress.Skins;
+using DevExpress.XtraLayout;
+
+namespace MoBi.UI.Extensions
+{
+   public static class LayoutItemExtensions
+   {
+      public static void InitializeDisabledColors(this LayoutControl layoutControl)
```
msevestre, line 9: @rwmcintosh This is copied from PKSim. We wanted to remove it but in fact we should move it to core so that we can have a disable text that we can actually read

src/MoBi.Presentation/Presenter/SelectContainerPresenter.cs
```diff
@@ -0,0 +1,102 @@
...
+using OSPSuite.Utility.Extensions;
+
+namespace MoBi.Presentation.Presenter
+{
+   public interface ISelectContainerPresenter : ISelectObjectPathPresenter
+   {
+      ObjectPath Select();
+   }
+
+   public class SelectContainerPresenter : AbstractDisposablePresenter<ISelectObjectPathView, ISelectObjectPathPresenter>, ISelectContainerPresenter
```
msevestre, line 20: @rwmcintosh @Yuri05 I copied the code from the SelectEvent and made it so that we can select a container. I do not really understand the original code. Thsi one is pretty straightforward as i do not add any dummy container

src/MoBi.Presentation/Mappers/ContainerToContainerDTOMapper.cs
```diff
@@ -11,16 +11,35 @@ public interface IContainerToContainerDTOMapper : IMapper<IContainer, ContainerD
...
 
       public virtual ContainerDTO MapFrom(IContainer container)
       {
-         return MapContainer(container, new ContainerDTO());
+         var dto = MapContainer(container, new ContainerDTO());
+         UpdateParentPath(container, dto);
+         return dto;
+      }
+
+      protected void UpdateParentPath(IContainer container, ContainerDTO containerDTO)
```
msevestre, line 30: This is the meat of the change here. If there is a parent container, I set the ParentPath from the parent otherwise I allow edit and set it to the ParentPath

src/MoBi.UI/Views/EditContainerView.cs
```diff
@@ -70,13 +80,20 @@ public override void InitializeResources()
          btName.ToolTip = ToolTips.Container.ContainerName;
          tabProperties.InitWith(AppConstants.Captions.Properties, ApplicationIcons.Properties);
          tabParameters.InitWith(AppConstants.Captions.Parameters, ApplicationIcons.Parameter);
+         layoutItemParentPath.Text = AppConstants.Captions.ParentPath.FormatForLabel();
+         layoutControl.InitializeDisabledColors(_lookAndFeel);
```
msevestre, line 84: This lines ensure that disable fields TEXT is not grayed out butg readble

msevestre (commented):
@Yuri05 @rwmcintosh 
Lots of noise in this commit due to removal of IObjectPathDTO interface. The changes are quite trivial

src/MoBi.UI/Views/EditContainerView.cs
Yuri05: finally :+1:

Yuri05 (approved):

PR #912, opened 2023-03-01
Fixes #892 distinct message when PKSim is present, but is not compatible

### LINKED ISSUE #892: When there's a version mismatch between MoBi and PKSim the starter message should indicate that
Right now it says that PKSim is not installed, but actually it's the wrong version

image

rwmcintosh:


rwmcintosh:
FYI @PavelBal 

Also, I broke the string into separate lines. The width made it difficult to read. I did the same for the previous string 

image

image

msevestre (approved):
you can replace " \n" with "\n" (the space before the \n)

PR #913, opened 2023-03-02
839 populate expression profile from pk sim db

rwmcintosh:


src/MoBi.Core/Services/IPKSimStarter.cs
```diff
@@ -8,5 +10,6 @@ public interface IPKSimStarter
       void StartWithWorkingJournalFile(string journalFilePath);
       IBuildingBlock CreateProfileExpression(ExpressionType expressionType);
       IBuildingBlock CreateIndividual();
+      List<ExpressionParameterValueUpdate> UpdateExpressionProfileFromDatabase(ExpressionProfileBuildingBlock expressionProfile);
```
msevestre, line 13: IReadonLy?

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForExpressionProfileBuildingBlock.cs
```diff
@@ -28,6 +36,35 @@ public IReadOnlyList<ExpressionProfileBuildingBlock> LoadFromPKML()
...
+         macroCommand.AddRange(expressionProfileUpdate.Select(parameter => updateCommandFor(buildingBlock, parameter.Path, parameter.UpdatedValue)));
+
+         return macroCommand.Run(Context);
+      }
+
+      private static ICommand updateCommandFor(ExpressionProfileBuildingBlock buildingBlock, ObjectPath path, double? value)
+      {
+         var parameterToUpdate = buildingBlock.FirstOrDefault(x => Equals(x.Path, path));
+         
+         if (parameterToUpdate == null)
```
msevestre, line 62: can you verify in the history that you do not get empty command int he parent macro command?

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForExpressionProfileBuildingBlock.cs
```diff
@@ -28,6 +36,35 @@ public IReadOnlyList<ExpressionProfileBuildingBlock> LoadFromPKML()
...
+         };
+
+         macroCommand.AddRange(expressionProfileUpdate.Select(parameter => updateCommandFor(buildingBlock, parameter.Path, parameter.UpdatedValue)));
+
+         return macroCommand.Run(Context);
+      }
+
+      private static ICommand updateCommandFor(ExpressionProfileBuildingBlock buildingBlock, ObjectPath path, double? value)
+      {
+         var parameterToUpdate = buildingBlock.FirstOrDefault(x => Equals(x.Path, path));
```
msevestre, line 60: Don;t we have a wa to search for parameter by path that's not o(n)? if feels like we will loop over this for each parameter of the expression profile.

src/MoBi.Presentation/Tasks/PKSimStarter.cs
```diff
@@ -80,9 +83,15 @@ public IBuildingBlock CreateIndividual()
...
+      public List<ExpressionParameterValueUpdate> UpdateExpressionProfileFromDatabase(ExpressionProfileBuildingBlock expressionProfile)
+      {
+         loadPKSimAssembly();
+         return executeMethod(getMethod(PKSIM_UI_STARTER_EXPRESSION_PROFILE_CREATOR, GET_EXPRESSION_DATABASE_QUERY), new object[] { expressionProfile }) as List<ExpressionParameterValueUpdate>;
+      }
+
+      private object executeMethod(MethodInfo method, object[] parameters = null)
       {
-         return method.Invoke(null, null);
+         return method.Invoke(null, parameters);
```
msevestre, line 94: lol

msevestre (changes requested):
looks good to me. Just a few suggestions

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForExpressionProfileBuildingBlock.cs
rwmcintosh: Yes, I verified empty commands don't appear in the history.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForExpressionProfileBuildingBlock.cs
rwmcintosh:
```
var parameterToUpdate = buildingBlock[path];
```
?

src/MoBi.Presentation/Tasks/PKSimStarter.cs
rwmcintosh: Prior to this feature, we haven't had to send anything over the fence to MoBi. We used to send IShell, but we found out that wasn't necessary.

msevestre (approved):

PR #922, opened 2023-03-09
Fixes #898 spatial structure allow creation of neighborhood

### LINKED ISSUE #898: Spatial Structure: Allow creation of neighborhood via UI (not drag and drop)
It should be possible to create neighborhoods by other means than drag-and-drop. E.g., a UI "Create neighborhood", where the user selects (or types in the full path to) the neighbors. In case of a tumor module, the user would select "PARENT|Tumor|Plasma" as one neighbor, and type in "Organism|VenousBlood|Plasma" as the second neighbor. the "PARENT" keyword is necessary as the path within the module is not complete. "PARENT" will be replaced by the container that is defined in the "Parent container" property.

image

msevestre:


msevestre:
@rwmcintosh This is a draft PR but you can start reviewing my changes to see if I am missing something obvious

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuFactoryForNeighborhoods.cs
```diff
@@ -0,0 +1,57 @@
...
+
+      public override IContextMenu CreateFor(ObjectBaseDTO objectBaseDTO, IPresenterWithContextMenu<IViewItem> presenter)
+      {
+         var contextMenu = _context.Resolve<ContextMenuForNeighborhoods>();
+         return contextMenu.InitializeWith(objectBaseDTO, presenter);
+      }
+
+      protected override bool IsSatisfiedBy(IEntity entity, IPresenterWithContextMenu<IViewItem> presenter)
+      {
+         return presenter.IsAnImplementationOf<IHierarchicalSpatialStructurePresenter>()
```
msevestre, line 31: don;t really lile this, Maybe we can just use the fact that it;s a container named neigthborhood

src/MoBi.Presentation/Presenter/SelectContainerPresenter.cs
```diff
@@ -19,84 +15,40 @@ public interface ISelectContainerPresenter : ISelectObjectPathPresenter
...
    public class SelectContainerPresenter : AbstractDisposablePresenter<ISelectObjectPathView, ISelectObjectPathPresenter>, ISelectContainerPresenter
    {
-      private readonly IContainerToContainerDTOMapper _containerDTOMapper;
-      private readonly IObjectPathFactory _objectPathFactory;
       private readonly IMoBiContext _context;
+      private readonly ISelectContainerInTreePresenter _selectContainerInTreePresenter;
+      private readonly IObjectPathFactory _objectPathFactory;
+      private readonly ISpatialStructureToSpatialStructureDTOMapper _spatialStructureDTOMapper;
 
       public SelectContainerPresenter(
```
msevestre, line 23: I creatred a component out of this so that I can use it multiple times

src/MoBi.Presentation/Presenter/SelectObjectInTreePresenter.cs
```diff
@@ -0,0 +1,52 @@
...
+      Func<ObjectBaseDTO, IReadOnlyList<ObjectBaseDTO>> GetChildren { get; set; }
+      bool IsValidSelection(ObjectBaseDTO selectedDTO);
+      void InitTreeStructure(IReadOnlyList<ObjectBaseDTO> entityDTOs);
+      IEntity SelectedEntity { get; }
+      ObjectBaseDTO SelectedDTO { get; }
+      ITreeNode TreeNodeFor(ObjectBaseDTO dto);
+      IObjectBase ObjectFrom(ObjectBaseDTO parentDTO);
+   }
+
+   public class SelectEntityInTreePresenter : AbstractPresenter<ISelectEntityInTreeView, ISelectEntityInTreePresenter>, ISelectEntityInTreePresenter
```
msevestre, line 23: that componet is superdumb
does nto even know how to iterate through children but the prop is passed as  afunc

src/MoBi.UI/Views/EditNeighborhoodBuilderView.cs
```diff
@@ -0,0 +1,66 @@
...
+      }
+
+      public void Activate()
+      {
+         ActiveControl = tbName;
+      }
+
+      public void AddFirstNeighborView(IView view)
+      {
+         panelFirstNeighbor.FillWith(view);
```
msevestre, line 36: need to react to selection of those so that the view is enabled or disabled. Not sure how to do this yet
Can't remember how we are doing it normally

msevestre:
@rwmcintosh @Yuri05 This is ready to review
The implementation to create the link in the builder is not implemented yet as we need to work now with Path and not container. That's why there is a TODO. Changes required in core. But I'd like this to be reviewed already

Here is the UI
image

src/MoBi.Presentation/Presenter/EditNeighborhoodBuilderPresenter.cs
```diff
@@ -0,0 +1,70 @@
...
+         _view.AddFirstNeighborView(_firstNeighborPresenter.BaseView);
+         _view.AddSecondNeighborView(_secondNeighborPresenter.BaseView);
+
+         _firstNeighborPresenter.StatusChanged += (o, e) => updateFirstNeighbor();
+         _secondNeighborPresenter.StatusChanged += (o, e) => updateSecondNeighbor();
+      }
+
+      private void updateSecondNeighbor()
+      {
+//TODO         _neighborhoodBuilder.SecondNeighbor =  _secondNeighborPresenter.NeighborPath;
```
msevestre, line 44: This will need to be done later once core is updated

src/MoBi.Presentation/Presenter/SelectNeighborPathPresenter.cs
```diff
@@ -0,0 +1,64 @@
...
+            return;
+
+         _selectedPathDTO.Path = containerObjectPath.PathAsString;
+         ViewChanged();
+      }
+
+      public void Init(ISpatialStructure spatialStructure, string label)
+      {
+         _view.Label = label;
+         var organism = spatialStructure.TopContainers.Find(x => x.ContainerType == ContainerType.Organism) ?? spatialStructure.TopContainers.FirstOrDefault();
```
msevestre, line 53: I am not so sure about this. Can we have two organisms in a spatial structure?

msevestre:
@rwmcintosh @Yuri05 I need to move with this so I am merging this. Please review and feedback will be performed in another PR

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuFactoryForContainer.cs
```diff
@@ -20,15 +18,17 @@ public ContextMenuFactoryForContainer(IMoBiContext context) : base(context)
 
       public override IContextMenu CreateFor(ObjectBaseDTO objectBaseDTO, IPresenterWithContextMenu<IViewItem> presenter)
       {
-         var contextMenu = IoC.Resolve<IContextMenuForContainer>();
```
rwmcintosh, line 23: An item we should always be on the lookout for from now on.

```IoC.Resolve```

File: src/MoBi.Presentation/Tasks/Edit/EditTasksForNeighborhoodBuilder.cs [hunk 8]
```diff
@@ -28,15 +29,27 @@ private IMoBiSpatialStructure getSpatialStructure()
          return _interactionTaskContext.Active<IMoBiSpatialStructure>();
       }
 
-      public override bool EditEntityModal(INeighborhoodBuilder entity, IEnumerable<IObjectBase> existingObjectsInParent, ICommandCollector commandCollector, IBuildingBlock buildingBlock)
+      public override bool EditEntityModal(INeighborhoodBuilder neighborhood, IEnumerable<IObjectBase> existingObjectsInParent, ICommandCollector commandCollector, IBuildingBlock buildingBlock)
       {
-         var name = _interactionTaskContext.DialogCreator.AskForInput(AppConstants.Dialog.AskForNewNeighborhoodBuilderName(entity.FirstNeighbor.Name, entity.SecondNeighbor.Name),
-            AppConstants.Captions.NewWindow(ObjectName), String.Empty,
-            GetForbiddenNamesWithoutSelf(entity, existingObjectsInParent));
+         // Either we are creating a neighborhood from the diagram to which case, to which case the neighborhood is connected
```
rwmcintosh, line 34: I don't understand this comment.

rwmcintosh: specifically - 'to which case, to which case'.

rwmcintosh: should it be 'in which case'

Yuri05:
Thinking about the use case "Adding an organ" (e.g. "Adding a tumor"): we have discussed something like the "outer neighborhoods". 
Tumor would be defined in a separate extension module and `Tumor/Plasma` would be connected to some Plasma compartments (`Venous/Plasma`, `Arterial/Plasma`) which are defined in **another** module.

This would require, that we define those nbhoods in the tumor module by setting:
* 1st neighbor path = path to a container **in the tumor module**
* 2nd neighbor path = path to a container **outside of the tumor module**

This means, that the selection of the 2nd neighbor in the dialog above must be **optional**.

I guess that the paths of both neighbors of a neighborhood will be editable, so that the neighbors can be easily modified by just editing the neighbor paths.

Thus for the "Tumor use case" the user would:
* select the 1st neighbor in the Dialog above
* keep the 2nd neighbor empty => this will create an **empty** 2nd neighbor path
* Once the neighborhood was created: edit the 2nd neighbor path (and e.g. enter `Organism|Venous Blood|Plasma`)

Thoughts?

@PavelBal

Yuri05:
Alternatively, the dialog above would allow to select the 2nd neighbor not only from the "own" spatial structure, but also from the spatial structure of another modules. Makes the implementation more complictaed ofc.

PavelBal:
> keep the 2nd neighbor empty => this will create an empty 2nd neighbor path
> Once the neighborhood was created: edit the 2nd neighbor path (and e.g. enter Organism|Venous Blood|Plasma)

Or directly enter the path for the second neighbor?

msevestre:
At the moment, both entries can be edited. I am not sure why we should start to restrict usage of one and not the other as they really are exactly the same. I'd like to see it in practice (.e.g.should be testable soon) and then we can revisit

Yuri05:
> At the moment, both entries can be edited. I am not sure why we should start to restrict usage of one and not the other as they really are exactly the same. I'd like to see it in practice (.e.g.should be testable soon) and then we can revisit

ok, I didn't get that both paths are editable - thought they could be only set via container selection.
All good then.

PR #923, opened 2023-03-11
Update to latest dll

msevestre:


src/MoBi.Presentation/Presenter/SpaceDiagram/SpatialStructureDiagramPresenter.cs
```diff
@@ -1,44 +1,50 @@
...
+         IMoBiConfiguration configuration,
+         IDiagramTask diagramTask,
+         IStartOptions runOptions,
+         IDiagramModelFactory diagramModelFactory)
+         : base(view, layouter, dialogCreator, diagramModelFactory, userSettings, context, diagramTask, runOptions)
       {
          _configuration = configuration;
-         _diagramPopupMenu = new PopupMenuSpaceDiagram(this, runOptions, container);
-         _containerPopupMenu = new PopupMenuFullContainerWithParametersNode(this, _context, runOptions, container);
-         _neighborhoodPopupMenu = new PopupMenuFullEntityNode<INeighborhoodBuilder>(this, _context, runOptions, container);
```
msevestre, line 41: @rwmcintosh I am removing those container that you added to each popup menu to satisfy the requirement of not using IoC. The container was already in the Context (not exposed) but I am exposing it now so that I don't have to have those weird `OSPSuite.Utility.Container.IContainer container` interfaces everywhere

msevestre (commented):
merging

PR #925, opened 2023-03-11
Fixes #924 update neighbors path in builder

### LINKED ISSUE #924: Update neighbors path in neighborhood builder when renaming a container

msevestre:


src/MoBi.Core/Commands/RenameContainerCommand.cs
```diff
@@ -0,0 +1,75 @@
...
+using OSPSuite.Assets;
+using OSPSuite.Core.Commands.Core;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Events;
+using OSPSuite.Utility.Extensions;
+
+namespace MoBi.Core.Commands
+{
+   public class RenameContainerCommand : BuildingBlockChangeCommandBase<ISpatialStructure>
```
msevestre, line 12: @Yuri05 Special command to rename containers in spatial structure. In this case, we need to also check if they are used in neighborhood and update the path

src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
```diff
@@ -1,19 +1,54 @@
...
+      /// </summary>
+      IReadOnlyList<NeighborhoodBuilder> GetConnectingNeighborhoods(IReadOnlyList<IContainer> neighbors, IObjectPathFactory objectPathFactory);
    }
 
    public class MoBiSpatialStructure : SpatialStructure, IMoBiSpatialStructure
    {
       public IDiagramModel DiagramModel { get; set; }
       public IDiagramManager<IMoBiSpatialStructure> DiagramManager { get; set; }
 
+      public IReadOnlyList<NeighborhoodBuilder> GetConnectingNeighborhoods(IReadOnlyList<IContainer> neighbors, IObjectPathFactory objectPathFactory)
```
msevestre, line 24: THis code was duplicated. It's probably wrong with the new concept but I did not want to have to update it all the time

src/MoBi.Presentation/Mappers/ObjectBaseToObjectBaseDTOMapper.cs
```diff
@@ -20,11 +22,24 @@ protected T Map<T>(IObjectBase objectBase) where T : ObjectBaseDTO, new()
...
+      {
+         //special case for neighborhood where we can create a dynamic description if not there
+         if (!string.IsNullOrEmpty(objectBase.Description))
+            return objectBase.Description;
+
+
+         if (!(objectBase is NeighborhoodBuilder neighborhood))
+            return string.Empty;
+
+         return ToolTips.Neighborhood.Between(neighborhood);
```
msevestre, line 41: I was fed up of not having tooltip. for neighborhoods in the tree!

src/MoBi.Presentation/Tasks/Edit/EditTaskFor.cs
```diff
@@ -87,28 +87,28 @@ private IMoBiCommand rename(T objectBase, IEnumerable<string> forbiddenNames, IB
       {
          var newName = NewNameFor(objectBase, forbiddenNames.ToList());
 
-         var objectName = _interactionTaskContext.GetTypeFor(objectBase);
```
msevestre, line 90: so confusing...

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
```diff
@@ -69,32 +69,20 @@ public IMoBiCommand SetContainerMode(IBuildingBlock buildingBlock, IContainer co
...
+         if (buildingBlock.IsAnImplementationOf<ISpatialStructure>() && containerCanBePartOfNeighborhoodPath(container))
+            return new RenameContainerCommand(container, newName, buildingBlock.DowncastTo<ISpatialStructure>());
 
-            if (firstFound && secondFound) neighborhoods.Add(neighborhood);
-         }
+         return base.GetRenameCommandFor(container, buildingBlock, newName, objectType);
+      }
 
-         return neighborhoods;
+      private bool containerCanBePartOfNeighborhoodPath(IContainer container)
```
msevestre, line 81: @Yuri05 I am not sure when I should use the special command. I said for now all container except the one called MoleculeProperties, the parameter (DistributedParameter) and the NeighborhoodBuilderthemselves (also container)

tests/MoBi.Tests/Core/Commands/RenameContainerCommandSpecs.cs
```diff
@@ -0,0 +1,107 @@
...
+   {
+      protected override void Because()
+      {
+         sut = new RenameContainerCommand(_container, "NEW_NAME", _spatialStructure).Run(_context);
+      }
+
+      [Observation]
+      public void should_rename_the_path_in_the_neighborhoods_not_referencing_this_container()
+      {
+         _neighborhood1.FirstNeighborPath.PathAsString.ShouldBeEqualTo("A|NEW_NAME|A");
```
msevestre, line 96: Funny I wrote this test and did not check the run but it fails. And of course, my implementation is wrong. I just cannot check for the exact path but somehow it should just be subpart lol. I'll fix it tomorrow

msevestre:
@Yuri05 I don't see any other way that scanning all neighborhoods everytime a container is renamed, as I just cannot simply see if the container is connected or not. I may have simply renamed the parent?

src/MoBi.Core/Commands/RenameContainerCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+         var objectPathFactory = context.ObjectPathFactory;
+         var containerPath = objectPathFactory.CreateAbsoluteObjectPath(_container);
+
+         //perform the rename
+         _oldName = _container.Name;
+         _container.Name = _newName;
+
+         //the container has been renamed, we need to update the path of all neighborhoods connected to it
+         //no need to update referenced in the neighborhood as they have not changed
+         var updateNeighborhoods = updateNeighborhoodsDef(containerPath, objectPathFactory.CreateAbsoluteObjectPath(_container));
```
msevestre, line 55: @Yuri05 @rwmcintosh @georgeDaskalakis This is a new try for me in .NET to create curried function that capture a context so that you don't have to repeat yourself with parameter who do not change.
See now the next call
instead of having to do 
```
updateNeighborhoods(containerPath, newContainerPath, x => x.FirstNeighborPath)
updateNeighborhoods(containerPath, newContainerPath, x => x.SecondNeighborPath)
```

I can simply do 
```
updateNeighborhoods(x => x.FirstNeighborPath)
updateNeighborhoods(x => x.SecondNeighborPath)
```

and do not even need to create a local variable newContainerPath

Thoughts on this syntax? It's very common to do this in functional languages such as javascript and I think it can simply the code a bit (here maybe not but I just want to get the hang of it)

src/MoBi.Core/Commands/RenameContainerCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+         updateNeighborhoods(x => x.FirstNeighborPath);
+         updateNeighborhoods(x => x.SecondNeighborPath);
+
+         context.PublishEvent(new RenamedEvent(_container));
+      }
+
+      private Action<Func<NeighborhoodBuilder, ObjectPath>> updateNeighborhoodsDef(string oldContainerPath, string newContainerPath) =>
+         (neighborPathFunc) =>
+         {
+            _buildingBlock.Neighborhoods.Select(x => new
```
msevestre, line 65: @Yuri05 Not sure of a better way to do this
the path could be something like A|B|C|D
we rename C to NEW
I need to replace THE C => A|B|NEW|D

src/MoBi.Core/Commands/RenameContainerCommand.cs
```diff
@@ -0,0 +1,80 @@
...
+               {
+                  NeihgborPath = neighborPathFunc(x),
+                  NeighborPathString = neighborPathFunc(x).ToString()
+               })
+               .Where(x => x.NeighborPathString.StartsWith(oldContainerPath))
```
Yuri05, line 70: I see a problem here.
E.g. imagine in your example above with the renaming C==>NEW you would have a nbhood with the neighbor path
A|B|CCC|D. C is not on this path, but it would become renamed to A|B|NEWCC|D which is wrong.

So actually it should be something like 
`NeighborPathString.Equals(oldContainerPath) || NeighborPathString.StartsWith($"{oldContainerPath}|")`

msevestre: yep good catch.

msevestre: @Yuri05 Updated with a test

tests/MoBi.Tests/Core/Commands/RenameContainerCommandSpecs.cs
```diff
@@ -0,0 +1,150 @@
...
+   {
+      protected override void Because()
+      {
+         sut = new RenameContainerCommand(_similarContainer, "NEW_NAME", _spatialStructure).Run(_context);
+      }
+      
+      [Observation]
+      public void should_not_rename_the_path_in_the_neighborhoods_not_referencing_this_container()
+      {
+         _neighborhood1.FirstNeighborPath.PathAsString.ShouldBeEqualTo("A|AAA|A");
```
msevestre, line 120: @Yuri05 This is your case. Well done

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
Yuri05: Is ok I think.

Yuri05 (approved):

PR #928, opened 2023-03-15
Start refactoring of ObjectBaseDTO

msevestre:


msevestre:
all: Starting a refactoring so that we can eventually stop retrieving object by Id all the time and simply use the underlying object saved in the DTO.

msevestre (commented):
Big change here
DTO now have a ref to the underlying objec.t Not more _context.Get<dto.id> BS required. I did not remove the usage everywhere however but something we should do when we see it (all context menus etc)
DTO as an ApplicationIcons and not an icon as string
fixed 1M typos
Actually implemented the task

src/MoBi.UI/MoBi.UI.csproj
```diff
@@ -57,6 +57,9 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Resources.resx</DependentUpon>
     </Compile>
+    <Compile Update="Views\EditNeighborhoodBuilderView.cs">
```
rwmcintosh, line 60: Why did this get added? I didn't think it was necessary.

tests/MoBi.Tests/Presentation/ParameterDTOSpecs.cs
```diff
@@ -18,13 +18,14 @@ protected override void Context()
       }
    }
 
-   public class When_creating_a_parameter_dto_for_a_constant_aprameter_with_an_undefined_value : concern_for_ParameterDTO
+   public class When_creating_a_parameter_dto_for_a_constant_a_parameter_with_an_undefined_value : concern_for_ParameterDTO
```
rwmcintosh, line 21: When_creating_a_parameter_dto_for_a_constant_a_parameter_with_an_undefined_value
When_creating_a_parameter_dto_for_a_constant_parameter_with_an_undefined_value

is it supposed to be constant_parameter instead of constant_a_parameter?

PR #948, opened 2023-04-04
Fixes #941 Import and export users default simulation settings

### LINKED ISSUE #941: Import and export users default simulation settings
That would allow users to easily change their default simulation settings for all projects without having to manually change each setting by hand.

Can this be serialized and deserialized with user settings. It seems to fit.

rwmcintosh:


rwmcintosh:
image

Should I rename this grouping? I'm not sure it's easy enough to understand.

I think the group name should be "Simulation Settings" and first button renamed to "Edit"

rwmcintosh:
Cut off with an ellipsis, but that's like the Workflows group caption.

image

src/MoBi.Assets/AppConstants.cs
```diff
@@ -806,7 +806,7 @@ public static class BarNames
          public static readonly string Views = "Views";
          public static readonly string Add = "Add";
          public static readonly string Edit = "Edit";
-         public static readonly string Project = "Project";
+         public static readonly string ProjectSimulationSettings = "Project Simulation Settings";
```
msevestre, line 809: From the screenshot, it feels too long
In which tab is this located in the ribbon?

src/MoBi.Presentation/Presenter/Main/MenuAndToolBarPresenter.cs
```diff
@@ -80,7 +80,7 @@ protected override void AddRibbonPages()
          _view.AddApplicationMenu(_buttonGroupRepository.Find(ButtonGroupIds.File));
          _view.AddPageGroupToPage(_buttonGroupRepository.Find(ButtonGroupIds.BuildingBlocks), AppConstants.RibbonPages.Modeling);
          _view.AddPageGroupToPage(_buttonGroupRepository.Find(ButtonGroupIds.Workflows), AppConstants.RibbonPages.Modeling);
-         _view.AddPageGroupToPage(_buttonGroupRepository.Find(ButtonGroupIds.Project), AppConstants.RibbonPages.Modeling);
+         _view.AddPageGroupToPage(_buttonGroupRepository.Find(ButtonGroupIds.ProjectSimulationSettings), AppConstants.RibbonPages.Modeling);
```
msevestre, line 83: Ok in Modeling. Do we have one for project

rwmcintosh: No, we don't, but that's the full bar for 'Modeling'. Maybe this could be named 'Project'.

image

rwmcintosh: Then rename our new grouping 'Simulation Settings'
The 'Create' group will be reduced to Module, Individual, Expression, Simulation. All these are added to the 'Project' so I think it makes some sense.

msevestre: I would not rename modeling to project. Let's simply use simulation settings for now

Issue #946, opened 2023-04-03
Create Simulation Dialog changes


rwmcintosh:
This dialog needs to be revised to allow users to select
- multiple modules
- multiple Expressions
- One individual
- optional PV and IC

PR #956, opened 2023-04-14
#946 create simulation dialog changes

rwmcintosh:


src/MoBi.Core/Serialization/Xml/Serializer/MoBiProjectXmlSerializer.cs
```diff
@@ -11,6 +11,7 @@ public override void PerformMapping()
       {
          base.PerformMapping();
          Map(x => x.ReactionDimensionMode);
+         Map(x => x.SimulationSettings);
```
rwmcintosh, line 14: This was missed in another issue. MoBi project was created with simulation settings, but wasn't saved to the project and therefore, was missing after loading a project

rwmcintosh:
The create is working, but still todo
- tests
- configure simulation (just needs a command)

rwmcintosh:
image

image

image

rwmcintosh:
you drag/drop to reorder modules and expressions

src/MoBi.Presentation/Nodes/ModuleNode.cs
```diff
@@ -0,0 +1,17 @@
...
+using OSPSuite.Presentation.Nodes;
+using OSPSuite.Utility;
+
+namespace MoBi.Presentation.Nodes
+{
+   public class ModuleNode : AbstractNode<Module>
+   {
+      public ModuleNode(Module module) : base(module)
+      {
+         Id = ShortGuid.NewGuid();
```
msevestre, line 11: Why not using the Id of the module? Is this what we do normally?

src/MoBi.Presentation/DTO/ModuleConfigurationDTO.cs
```diff
@@ -0,0 +1,43 @@
...
+         _moleculeStartValuesCollection.AddRange(moduleConfiguration.Module.MoleculeStartValuesCollection);
+         _parameterStartValuesCollection.AddRange(moduleConfiguration.Module.ParameterStartValuesCollection);
+      }
+
+      public ParameterStartValuesBuildingBlock SelectedParameterStartValues { get; set; }
+      public MoleculeStartValuesBuildingBlock SelectedMoleculeStartValues { get; set; }
+      public IReadOnlyList<MoleculeStartValuesBuildingBlock> MoleculeStartValuesCollection => _moleculeStartValuesCollection;
+      public IReadOnlyList<ParameterStartValuesBuildingBlock> ParameterStartValuesCollection => _parameterStartValuesCollection;
+
+      public static MoleculeStartValuesBuildingBlock NullMoleculeStartValues { get; } = new MoleculeStartValuesBuildingBlock().WithName(AppConstants.Captions.NoMoleculeStartValues);
```
msevestre, line 28: I find it very hard to reason about to have static members between various class members. Can you move them higher up and separate them clearly. Also you are accessing them before they are declared. This can create issues sometimes

src/MoBi.Presentation/DTO/IndividualSelectionDTO.cs
```diff
@@ -0,0 +1,26 @@
...
+namespace MoBi.Presentation.DTO
+{
+   public class IndividualSelectionDTO
+   {
+      public IndividualSelectionDTO(IReadOnlyList<IndividualBuildingBlock> allIndividuals)
+      {
+         _allIndividuals.AddRange(allIndividuals);
+      }
+      
+      public static IndividualBuildingBlock NullIndividualBuildingBlock = new IndividualBuildingBlock().WithName(AppConstants.Captions.IndividualNotSelected);
```
msevestre, line 15: Similarily, I don't think this object belongs in a DTO. I would suggest to create a class and put all those statis members there). This class can also be static that way, there is no ambiguity

src/MoBi.Presentation/Mappers/SelectedIndividualToIndividualSelectionDTOMapper.cs
```diff
@@ -0,0 +1,31 @@
...
+      public SelectedIndividualToIndividualSelectionDTOMapper(IMoBiContext moBiContext)
+      {
+         _moBiContext = moBiContext;
+      }
+      
+      public IndividualSelectionDTO MapFrom(IndividualBuildingBlock individualBuildingBlock)
+      {
+         var individualSelectionDTO = new IndividualSelectionDTO(_moBiContext.CurrentProject.IndividualsCollection);
+         if(individualBuildingBlock != null)
+            individualSelectionDTO.SelectedIndividualBuildingBlock = individualBuildingBlock;
```
msevestre, line 26: this is weird. Why are you doing this if you have the Null individual Building block? the behavior of setting null could be in the setter of the DTO. (tell don't ask)

src/MoBi.Presentation/Nodes/TreeNodeFactory.cs
```diff
@@ -107,6 +123,38 @@ public ITreeNode CreateFor(SimulationConfiguration simulationConfiguration)
...
+      {
+         var moduleConfigurationNode = new ModuleConfigurationNode(moduleConfiguration).WithIcon(ApplicationIcons.Module);
+         var module = moduleConfiguration.Module;
+
+         addModuleBuildingBlocks(moduleConfigurationNode, module);
+         
+         if(moduleConfiguration.SelectedMoleculeStartValues != null)
+            createAndAddNodeUnder(moduleConfigurationNode, moduleConfiguration.SelectedMoleculeStartValues);
+         if(moduleConfiguration.SelectedParameterStartValues != null)
+            createAndAddNodeUnder(moduleConfigurationNode, moduleConfiguration.SelectedParameterStartValues);
```
msevestre, line 136: I feel that resharper needs to be run in this file. Stuff seems to be a bit out of space alignment

src/MoBi.Presentation/Nodes/TreeNodeFactory.cs
```diff
@@ -107,6 +123,38 @@ public ITreeNode CreateFor(SimulationConfiguration simulationConfiguration)
...
 
+      public ITreeNode CreateFor(ModuleConfiguration moduleConfiguration)
+      {
+         var moduleConfigurationNode = new ModuleConfigurationNode(moduleConfiguration).WithIcon(ApplicationIcons.Module);
+         var module = moduleConfiguration.Module;
+
+         addModuleBuildingBlocks(moduleConfigurationNode, module);
+         
+         if(moduleConfiguration.SelectedMoleculeStartValues != null)
+            createAndAddNodeUnder(moduleConfigurationNode, moduleConfiguration.SelectedMoleculeStartValues);
```
msevestre, line 134: whata about having the `createAndAddNodeUnder` method check that the building block is not null  instead of checking all the time?  you are doing it for `addBuildingBlockNodeUnder` so it would be more cohesive

src/MoBi.Presentation/Presenter/CreateSimulationPresenter.cs
```diff
@@ -117,13 +117,7 @@ private void validateDimensions(IModel model)
...
       private CreationResult createModel()
       {
-         //Create the model using a build configuration referencing the templates building block so that references to template builders are defined properly 
-         //we override the _buildConfiguration so that reference to builders are saved
-
-         // TODO SIMULATION_CONFIGURATION
-         // _simulationConfiguration = _buildConfigurationFactory.CreateFromReferencesUsedIn(_simulationConfiguration);
-         
-         _simulationConfiguration.SimulationSettings = _cloneManager.CloneBuildingBlock(_context.CurrentProject.SimulationSettings);
+         _simulationConfiguration.SimulationSettings = _cloneManager.CloneBuildingBlock(_context.CurrentProject.SimulationSettings ?? new SimulationSettings { Solver = new SolverSettings() });
```
msevestre, line 120: this is weird here.
Can this be null? If yes, why? 
If not, then not required.
Also don't we have a SimulationSettingsFactory that would set all the default for us? Presenter creating a bunch of domain objects that have nothing to do with it's purpose feels weird. In fact the whole retrieval of simulation settings belongs probably outside (if we want to separate concerns)

src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -0,0 +1,149 @@
...
+      public int CompareSelectedNodes(ITreeNode node1, ITreeNode node2)
+      {
+         var expression1 = node1.TagAsObject as ExpressionProfileBuildingBlock;
+         var expression2 = node2.TagAsObject as ExpressionProfileBuildingBlock;
+
+         return _selectedExpressions.IndexOf(expression1) - _selectedExpressions.IndexOf(expression2);
+      }
+
+      public IndividualBuildingBlock SelectedIndividual => _individualSelectionDTO.IsNull() ? null : _individualSelectionDTO.SelectedIndividualBuildingBlock;
+      public IReadOnlyList<ExpressionProfileBuildingBlock> ExpressionProfiles => _selectedExpressions;
```
msevestre, line 115: run reshaper in this file too. Space is missing

src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -0,0 +1,149 @@
...
+      public bool CanDrop(ITreeNode dragNode, ITreeNode targetNode)
+      {
+         return targetNode?.TagAsObject is ExpressionProfileBuildingBlock;
+      }
+
+      public void MoveNode(ITreeNode dragNode, ITreeNode targetNode)
+      {
+         var movingExpression = dragNode.TagAsObject as ExpressionProfileBuildingBlock;
+         var targetExpression = targetNode.TagAsObject as ExpressionProfileBuildingBlock;
+
```
msevestre, line 131: need to check that they are not null? not sure

src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -0,0 +1,149 @@
...
+         _selectedExpressions.Remove(movingExpression);
+         _selectedExpressions.Insert(_selectedExpressions.IndexOf(targetExpression), movingExpression);
+         _view.SortSelectedExpressions();
+      }
+   }
+
+   public interface IEditIndividualAndExpressionConfigurationsView : IView<IEditIndividualAndExpressionConfigurationsPresenter>
```
msevestre, line 138: move in the view file please

src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -0,0 +1,149 @@
...
+   {
+      void BindTo(IndividualSelectionDTO individualSelectionDTO);
+      void AddUnusedExpression(ITreeNode treeNodeToAdd);
```
msevestre, line 141: Why is this a treeNode when the method is called AddUnusedExpression?

src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -0,0 +1,149 @@
...
+      void AddUsedExpression(ITreeNode treeNodeToAdd);
+      bool EnableAdd { get; set; }
+      bool EnableRemove { get; set; }
+      void RemoveUsedExpression(ITreeNode selectedNode);
```
msevestre, line 145: above is treenodeToAdd but here it's selectedNode

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -0,0 +1,203 @@
...
+      }
+
+      private void addUnusedModulesToSelectionView()
+      {
+         _context.CurrentProject.Modules.Where(module => !moduleInUse(module)).Each(addModuleToSelectionView);
+      }
+
+      private bool moduleInUse(Module module)
+      {
+         return _moduleConfigurations.Any(x => x.Uses(module));
```
msevestre, line 52: is this right? Isn't a module conifugiration using a clone of the module?

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -0,0 +1,203 @@
...
+
+      private ModuleConfigurationDTO dtoFor(ModuleConfiguration moduleConfiguration)
+      {
+         return _moduleConfigurations.FirstOrDefault(x => x.Uses(moduleConfiguration));
+      }
+
+
+      private void addModuleToSelectionView(Module module)
+      {
+         _view.AddModuleNode(_treeNodeFactory.CreateFor(module));
```
msevestre, line 63: here it's called AddModuleNode. Above it should be AddxxExpressionNode and that would make sense to me

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -0,0 +1,203 @@
...
+      }
+
+      private ModuleConfiguration moduleConfigurationFor(ITreeNode treeNode)
+      {
+         return treeNode?.TagAsObject as ModuleConfiguration;
+      }
+      
+      private ModuleConfigurationDTO moduleConfigurationDTOFor(ITreeNode treeNode)
+      {
+         var moduleConfiguration = moduleConfigurationFor(treeNode);
```
msevestre, line 130: It feels like we have 3 layers of indirection
The object, the node and the DTO. Why? (typically, a node is somekind of DTO)
it's hard to follow why goes on. What surprises me is that you are trying to retrieve. a DTO from an object. we use it sometimes but not so often

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -0,0 +1,203 @@
...
+         return dtoFor(moduleConfiguration);
+      }
+
+      public IReadOnlyList<MoleculeStartValuesBuildingBlock> MoleculeStartValuesCollectionFor(ITreeNode selectedNode)
+      {
+         if (selectedNode == null)
+            return new List<MoleculeStartValuesBuildingBlock>();
+
+         var dto = moduleConfigurationDTOFor(selectedNode);
+         return dto.MoleculeStartValuesCollection;
```
msevestre, line 140: another example of the 3 levels of indirection. Node=>dto=>domain

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -0,0 +1,203 @@
...
+         var dto1 = moduleConfigurationDTOFor(node1);
+         var dto2 = moduleConfigurationDTOFor(node2);
+
+         return _moduleConfigurations.IndexOf(dto1) - _moduleConfigurations.IndexOf(dto2);
+      }
+
+      public IReadOnlyList<ModuleConfiguration> ModuleConfigurations => _moduleConfigurations.Select(x => x.ModuleConfiguration).ToList();
+
+      public bool CanDrag(ITreeNode node)
+      {
```
msevestre, line 163: any reason why line 160 is on one line and lie 162 not? I'd suggest to always shorten one liner in one line 

public bool CanDrag(ITreeNode node)=>nodeIsModuleConfiguration(node);

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -0,0 +1,203 @@
...
+         
+         _moduleConfigurations.Remove(movingConfiguration);
+
+         var targetNodeIndex = _moduleConfigurations.IndexOf(targetConfiguration);
+         _moduleConfigurations.Insert(targetNodeIndex, movingConfiguration);
+         _view.SortSelectedModules();
+      }
+   }
+
+   public interface IEditModuleConfigurationsPresenter : ISimulationConfigurationItemPresenter, ICanDragDropPresenter
```
msevestre, line 192: move up by convention. Interface always first in presentation file.

src/MoBi.UI/Views/EditIndividualAndExpressionConfigurationsView.cs
```diff
@@ -0,0 +1,147 @@
...
+      public EditIndividualAndExpressionConfigurationsView(IImageListRetriever imageListRetriever)
+      {
+         InitializeComponent();
+         _imageListRetriever = imageListRetriever;
+         _treeNodeExplorerViewDragDropBinder = new TreeNodeExplorerViewDragDropBinder(simulationExpressionsTree);
+      }
+
+      public override void InitializeResources()
+      {
+         projectExpressionsTree.TreeView.StateImageList = _imageListRetriever.AllImages16x16;
```
msevestre, line 34: call base. We never know when we'll do sthg in the base class

msevestre (changes requested):
Looks very good. I have not started the UI and yet I have a good feel of what it does
I am just a bit surprised by the need of so many DTOs. Can't we just use Node as DTO (being defacto our view representation already?)

src/MoBi.Presentation/Presenter/CreateSimulationPresenter.cs
rwmcintosh: It cannot be null. It was a patch I made to cover for a previous bug, which I then fixed, but then didn't remove the patch.

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
rwmcintosh: The only thing that prevents me from getting rid of this DTO is that during configuration, I don't want to track changes using the domain object. The user could cancel, and I want to just get rid of everything. With the DTO I can keep track of selected start values easily. I think that's the only part where I need a temporary object besides the node. Although, the node also has child nodes with these objects as tags. I can try that.

msevestre: I can't tell. This is just hard to follow because of this level of indirection. In PKSim, we clone the whole config before editing. If the user cancels, we just discard. Otherwise we use as is. No need to synchronize anything

rwmcintosh: A clone here will make everything easier.

PR #957, opened 2023-04-18
#946 Some small changes to layout and mapping a new module configuration

rwmcintosh:
I was going to build all this into my tests and configure PR, but there's one issue that would stop you from really using this UI meanwhile so it needs to be included in OSMOSES as I work on the configure feature.

### DIFF
--- a/src/MoBi.Core/Commands/UpdateSimulationConfigurationCommand.cs
+++ b/src/MoBi.Core/Commands/UpdateSimulationConfigurationCommand.cs
@@ -1,0 +1,34 @@
+using MoBi.Assets;
+using MoBi.Core.Domain.Model;
+using OSPSuite.Assets;
+using OSPSuite.Core.Commands;
+using OSPSuite.Core.Domain.Builder;
+
+namespace MoBi.Core.Commands
+{
+   public class UpdateSimulationConfigurationCommand : OSPSuiteCommand<IMoBiContext>
+   {
+      private SimulationConfiguration _newConfiguration;
+      private IMoBiSimulation _simulation;
+
+      public UpdateSimulationConfigurationCommand(IMoBiSimulation simulation, SimulationConfiguration newConfiguration)
+      {
+         ObjectType = ObjectTypes.Simulation;
+         CommandType = AppConstants.Commands.EditCommand;
+         Description = AppConstants.Commands.ChangeSimulationConfiguration(simulation.Name);
+         _newConfiguration = newConfiguration;
+         _simulation = simulation;
+      }
+
+      protected override void ExecuteWith(IMoBiContext context)
+      {
+         _simulation.Configuration = _newConfiguration;
+      }
+
+      protected override void ClearReferences()
+      {
+         _simulation = null;
+         _newConfiguration = null;
+      }
+   }
+}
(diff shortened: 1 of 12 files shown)


src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -154,7 +154,22 @@ public int CompareSelectedNodes(ITreeNode node1, ITreeNode node2)
          return _moduleConfigurationsDTOs.IndexOf(dto1) - _moduleConfigurationsDTOs.IndexOf(dto2);
       }
 
-      public IReadOnlyList<ModuleConfiguration> ModuleConfigurations => _moduleConfigurationsDTOs.Select(x => x.ModuleConfiguration).ToList();
+      public IReadOnlyList<ModuleConfiguration> ModuleConfigurations => _moduleConfigurationsDTOs.Select(mapNewModuleConfiguration).ToList();
+
+      private ModuleConfiguration mapNewModuleConfiguration(ModuleConfigurationDTO dto)
```
rwmcintosh, line 159: Somehow I lost the generating of a new ModuleConfiguration from dtos during refactoring.

src/MoBi.Presentation/Nodes/TreeNodeFactory.cs
```diff
@@ -126,8 +126,10 @@ public ITreeNode CreateFor(ModuleConfigurationDTO moduleConfiguration)
 
          addModuleBuildingBlocks(moduleConfigurationNode, module);
 
-         createAndAddNodeUnder(moduleConfigurationNode, moduleConfiguration.SelectedMoleculeStartValues);
-         createAndAddNodeUnder(moduleConfigurationNode, moduleConfiguration.SelectedParameterStartValues);
+         if(moduleConfiguration.HasMoleculeStartValues())
```
rwmcintosh, line 129: sometimes the selected start values are null, other times, they are static Null objects

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -184,9 +199,9 @@ private void removeNode(ITreeNode parentNode, ITreeNode nodeToRemove)
 
       public bool CanDrag(ITreeNode node) => nodeIsModuleConfiguration(node);
 
-      private bool nodeIsModuleConfiguration(ITreeNode node) => node.TagAsObject is ModuleConfigurationDTO;
+      private bool nodeIsModuleConfiguration(ITreeNode node) => node?.TagAsObject is ModuleConfigurationDTO;
```
rwmcintosh, line 202: drag a node where there isn't already a node, get an exception.

rwmcintosh:
Some spacing issues addressed.
image

src/MoBi.Presentation/Nodes/TreeNodeFactory.cs
rwmcintosh: It tests for the special null object

src/MoBi.Core/Commands/UpdateSimulationConfigurationCommand.cs
```diff
@@ -0,0 +1,34 @@
+using MoBi.Assets;
+using MoBi.Core.Domain.Model;
+using OSPSuite.Assets;
+using OSPSuite.Core.Commands;
+using OSPSuite.Core.Domain.Builder;
+
+namespace MoBi.Core.Commands
+{
+   public class UpdateSimulationConfigurationCommand : OSPSuiteCommand<IMoBiContext>
```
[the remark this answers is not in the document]
rwmcintosh: We can change it no problem. Should be just serialize/deserialize to restore.

PR #967, opened 2023-04-18
966 add calculation methods to simulation configuration

rwmcintosh:


src/MoBi.Presentation/Presenter/ConfigureSimulationPresenter.cs
```diff
@@ -1,7 +1,6 @@
 using MoBi.Assets;
```
rwmcintosh, line 1: Changes in this file are still WIP

src/MoBi.Core/Domain/Services/SimulationFactory.cs
```diff
@@ -28,13 +28,16 @@ public class SimulationFactory : ISimulationFactory
       private readonly ICreationMetaDataFactory _creationMetaDataFactory;
       private readonly ISimulationParameterOriginIdUpdater _simulationParameterOriginIdUpdater;
       private readonly IDiagramManagerFactory _diagramManagerFactory;
+      private readonly ISimulationConfigurationFactory _simulationConfigurationFactory;
 
-      public SimulationFactory(IIdGenerator idGenerator, ICreationMetaDataFactory creationMetaDataFactory, ISimulationParameterOriginIdUpdater simulationParameterOriginIdUpdater, IDiagramManagerFactory diagramManagerFactory)
+      public SimulationFactory(IIdGenerator idGenerator, ICreationMetaDataFactory creationMetaDataFactory, ISimulationParameterOriginIdUpdater simulationParameterOriginIdUpdater,
```
msevestre, line 33: all on one line I would say

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+         _calculationMethodRepository = calculationMethodRepository;
+         _cloneManager = cloneManager;
+         _context = context;
+      }
+
+      public SimulationConfiguration Create()
+      {
+         var simulationConfiguration = new SimulationConfiguration
+         {
+            SimulationSettings = _cloneManager.Clone(_context.CurrentProject.SimulationSettings)
```
msevestre, line 31: Can those be null? They should not right?

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
```diff
@@ -0,0 +1,40 @@
...
+         };
+
+         _calculationMethodRepository.All().Each(
+            cm => simulationConfiguration.AddCalculationMethod(_cloneManager.Clone(cm, new FormulaCache())));
```
msevestre, line 35: what's with the new formula cache here? This looks suspicious. Aren't they also buildingBlocks?

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
rwmcintosh: you mean ```_context.CurrentProject.SimulationSettings```. No that should not be null. If that happens to you again then try to write the steps.

src/MoBi.Presentation/Nodes/ExpressionProfilesNode.cs
```diff
@@ -0,0 +1,19 @@
+﻿using MoBi.Assets;
+using OSPSuite.Assets;
+using OSPSuite.Presentation.Nodes;
+using OSPSuite.Utility;
+
+namespace MoBi.Presentation.Nodes
+{
+   public class ExpressionProfilesNode : AbstractNode
```
msevestre, line 8: Can we call it Folder =>  ExpressionProfileFolderNode. I think this is what we use for Folder and it's clearer than seeing the s

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
rwmcintosh: That's a copy/paste from BuildConfigurationFactory in develop branch. I wonder if that was a compensation that isn't needed now. I'll try it and update if it works.

src/MoBi.Core/Domain/Services/SimulationFactory.cs
rwmcintosh: Are you sure? That's almost 300 characters long.

msevestre: Sorry what I mean is 
ONE PARAMETER PER LINE
So split at ,

PR #970, opened 2023-04-20
#946 configure simulation

rwmcintosh:


src/MoBi.Core/Domain/Services/SimulationFactory.cs
```diff
@@ -20,6 +24,10 @@ public interface ISimulationFactory
       ///    Creates and returns a new <see cref="IMoBiSimulation" />
       /// </summary>
       IMoBiSimulation Create();
+
+      IMoBiSimulation CreateSimulationAndValidate(SimulationConfiguration configuration, string name);
```
msevestre, line 28: change name=>simulationName

src/MoBi.Core/Domain/Services/SimulationFactory.cs
```diff
@@ -20,6 +24,10 @@ public interface ISimulationFactory
...
+
+      IModel CreateModelAndValidate(string modelName, SimulationConfiguration simulationConfiguration);
```
msevestre, line 30: follow same order as method before

src/MoBi.Core/Domain/Services/SimulationFactory.cs
```diff
@@ -64,5 +84,62 @@ public IMoBiSimulation Create()
...
+
+         if (results == null || results.IsInvalid)
+            throw new MoBiException(AppConstants.Exceptions.CouldNotCreateSimulation);
+
+         validateDimensions(results.Model, results.SimulationBuilder);
+
+         return results.Model;
+      }
+
+      public IMoBiSimulation CreateSimulationAndValidate(SimulationConfiguration configuration, string name)
```
msevestre, line 106: wondering why CreateSImulationAndValidate does not use CreateModelAndValidate. Almost the same aside from the heavyWorkManager

src/MoBi.Presentation/Mappers/ModuleConfigurationToModuleConfigurationDTOMapper.cs
```diff
@@ -8,12 +10,32 @@ public interface IModuleConfigurationToModuleConfigurationDTOMapper : IMapper<Mo
...
       public ModuleConfigurationDTO MapFrom(ModuleConfiguration moduleConfiguration)
       {
-         return new ModuleConfigurationDTO(moduleConfiguration);
+         return new ModuleConfigurationDTO(createProjectModuleConfiguration(moduleConfiguration));
+      }
+
+      private ModuleConfiguration createProjectModuleConfiguration(ModuleConfiguration moduleConfiguration)
+      {
+         // If the module configuration already contains project building blocks then use it
+         var module = _context.CurrentProject.Modules.First(x => Equals(x.Name, moduleConfiguration.Module.Name));
```
msevestre, line 31: FindByName
Also using first means it will cras if it's not found. On purpose?

src/MoBi.Presentation/Mappers/ModuleConfigurationToModuleConfigurationDTOMapper.cs
```diff
@@ -8,12 +10,32 @@ public interface IModuleConfigurationToModuleConfigurationDTOMapper : IMapper<Mo
...
+         if (module.Equals(moduleConfiguration.Module))
+            return moduleConfiguration;
+
+         // Otherwise create a new module configuration with the same name and the building blocks from the project
+         return new ModuleConfiguration(module,
+            moduleConfiguration.SelectedMoleculeStartValues == null ? null : module.MoleculeStartValuesCollection.FindByName(moduleConfiguration.SelectedMoleculeStartValues?.Name),
```
msevestre, line 37: ? not needed at the end

msevestre: I would rewrite this to be more concise
module.MoleculeStartValuesCollection.FindByName(moduleConfiguration.SelectedMoleculeStartValues?.Name)??null

src/MoBi.Presentation/Mappers/SelectedIndividualToIndividualSelectionDTOMapper.cs
```diff
@@ -18,13 +18,12 @@ public SelectedIndividualToIndividualSelectionDTOMapper(IMoBiContext moBiContext
...
       }
-      
+
       public IndividualSelectionDTO MapFrom(IndividualBuildingBlock individualBuildingBlock)
       {
-         var individualSelectionDTO = new IndividualSelectionDTO(_moBiContext.CurrentProject.IndividualsCollection)
-         {
-            SelectedIndividualBuildingBlock = individualBuildingBlock
-         };
+         var individualSelectionDTO = new IndividualSelectionDTO(_moBiContext.CurrentProject.IndividualsCollection);
```
msevestre, line 24: we have an interface to retrieve the project. It's almost the same but abstract the fact that you are using the context all the time. Should we use this instead or it does nto make sense? IProjectRetriever or sthg like this. The context is the heavy duty object that does it all

src/MoBi.Presentation/Mappers/SelectedIndividualToIndividualSelectionDTOMapper.cs
```diff
@@ -18,13 +18,12 @@ public SelectedIndividualToIndividualSelectionDTOMapper(IMoBiContext moBiContext
...
+         if (individualBuildingBlock != null)
+            individualSelectionDTO.SelectedIndividualBuildingBlock = _moBiContext.CurrentProject.IndividualsCollection.FindByName(individualBuildingBlock.Name);
```
msevestre, line 26: maybe we need some acessor method . We are always doing so many A.B.C.D . I am not a fan

src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -55,7 +54,12 @@ public void Edit(SimulationConfiguration simulationConfiguration)
 
       private void addUsedExpressionsToSelectedView(SimulationConfiguration simulationConfiguration)
       {
-         simulationConfiguration.ExpressionProfiles.Each(addUsedExpressionToSelectedView);
+         simulationConfiguration.ExpressionProfiles.Each(profile =>
+         {
+            var projectProfile = _context.CurrentProject.ExpressionProfileCollection.FindByName(profile.Name);
```
msevestre, line 59: same here. We need ExpressionByName etc..or maybe this is overkill.

src/MoBi.Presentation/Presenter/EditModuleConfigurationsPresenter.cs
```diff
@@ -98,6 +103,7 @@ public void RemoveModuleConfiguration(ITreeNode selectedTreeNode)
          _view.RemoveNodeFromSelectedView(selectedTreeNode);
 
          addModuleToSelectionView(moduleConfigurationDTOFor(selectedTreeNode).Module);
+         ViewChanged();
```
msevestre, line 106: explain why this is required. We never put comment there but I think this is helpful

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForSimulation.cs
```diff
@@ -87,15 +94,30 @@ public IMoBiCommand AddToProject(IMoBiSimulation simulation)
...
-         using (var presenter = ApplicationController.Start<ICreateSimulationPresenter>())
+         using (var presenter = ApplicationController.Start<ICreateSimulationConfigurationPresenter>())
          {
-            return presenter.Create();
+            IMoBiSimulation simulation = null;
+            var moBiSimulation = _simulationFactory.Create();
+            initializeDefaultIndividual(moBiSimulation);
+            var simulationConfiguration = presenter.CreateBasedOn(moBiSimulation);
+
+            if (simulationConfiguration != null)
```
msevestre, line 104: Would it be a bit easier to read if we return early? You would not have to initialize sim line 99 
Soif null (cancel) return null

src/MoBi.Presentation/Mappers/ModuleConfigurationToModuleConfigurationDTOMapper.cs
rwmcintosh: Yeah, this should always be able to find the like-named module. Although, at the moment, I'm not sure that if you rename a project module, that it would be renamed in simulation, but that's the idea right?
And you can't remove a used module from the project?

src/MoBi.Presentation/Mappers/ModuleConfigurationToModuleConfigurationDTOMapper.cs
```diff
@@ -28,14 +27,14 @@ public ModuleConfigurationDTO MapFrom(ModuleConfiguration moduleConfiguration)
       private ModuleConfiguration createProjectModuleConfiguration(ModuleConfiguration moduleConfiguration)
       {
          // If the module configuration already contains project building blocks then use it
-         var module = _context.CurrentProject.Modules.First(x => Equals(x.Name, moduleConfiguration.Module.Name));
+         var module = _projectRetriever.Current.Modules.First(x => Equals(x.Name, moduleConfiguration.Module.Name));
```
msevestre, line 30: HERE

msevestre (approved):

PR #971, opened 2023-04-20
Fixes #963 simulation settings are null when loadinga simulation

### LINKED ISSUE #963: SImulation Settings project are null when loading  a simulation from file
Load simulation from previous issues
Select module, individual, no expression profile
At one point, clone of simulation settings in project crashed because they were null

msevestre:


tests/MoBi.Tests/Presentation/ProjectTaskSpecs.cs
```diff
@@ -676,13 +671,6 @@ protected override void Because()
          sut.New(ReactionDimensionMode.AmountBased);
       }
 
-      [Observation]
-      public void there_should_be_a_simulation_settings_created_by_the_simulation_settings_factory()
-      {
-         _context.CurrentProject.SimulationSettings.ShouldNotBeNull();
```
msevestre, line 682: here the test is simpler by testing that settings is the object you expect

PR #976, opened 2023-04-24
Fixes #974 Module Context Menu => Simply

### LINKED ISSUE #974: Module Context Menu => Simply
Instead of Create Empty and Create with BB, only one action
Create Module... which opens the Create with bb. The NAME should be selected as soon as the view opens so that one can enter the name directly and OK. Create Empty needs to go

rwmcintosh:


src/MoBi.Presentation/Mappers/CreateModuleDTOToModuleMapper.cs
```diff
@@ -55,32 +47,39 @@ public class CreateModuleDTOToModuleMapper : ModuleDTOMapper, ICreateModuleDTOTo
...
+            return;
+         
+         module.Add(buildingBlock);
+      }
+
       public Module MapFrom(CreateModuleDTO createModuleDTO)
       {
          var module = _context.Create<Module>().WithIcon(ApplicationIcons.Module.IconName).WithName(createModuleDTO.Name);
 
-         module.Add(conditionalCreate(createModuleDTO.WithMolecule,
```
rwmcintosh, line 62: the conditionalCreate used to assign to the member, but now calls the Add method. We just need to not add null bb's now

src/MoBi.UI/Views/CreateModuleView.cs
```diff
@@ -36,6 +36,7 @@ public override void InitializeResources()
          cbMoleculeStartValues.Text = MoleculeStartValues;
          cbParameterStartValues.Text = ParameterStartValues;
          createBuildingBlocksGroup.Text = CreateBuildingBlocks;
+         tbModuleName.Select();
```
rwmcintosh, line 39: Making sure the name is focused and has the cursor when the view is shown

src/MoBi.Assets/AppConstants.cs
```diff
@@ -1095,9 +1095,8 @@ public static class MenuNames
          public static string AddExisting(string objectTypeName) => $"Load {objectTypeName}...";
 
          public static string AddExistingFromTemplate(string objectTypeName) => $"Load {objectTypeName} from Template...";
-
-         public static string AddEmptyModule => $"Create Empty Module...";
-         public static string AddModuleWithBuildingBlocks => $"Create Module with Building Blocks...";
+         
+         public static string AddModuleWithBuildingBlocks => $"Create Module";
```
rwmcintosh, line 1099: I'm just using the extension method 'WithEllipsis' to add the '...'

src/MoBi.UI/Views/CreateModuleView.cs
msevestre: I think there is a method to override in the view. This sometime does not work
 protected virtual void SetActiveControl()
      {
      }

```
  protected override void SetActiveControl()
      {
         ActiveControl = tbName;
      }
```

msevestre (changes requested):

PR #977, opened 2023-04-24
Fixes #965 Creating Module should not be case sensitivie

### LINKED ISSUE #965: Creating Module should not be case sensitivie
I can create module S and s. This should not be possible

rwmcintosh:


src/MoBi.Presentation/DTO/CreateModuleDTO.cs
```diff
@@ -17,7 +17,7 @@ public CreateModuleDTO()
 
       public void AddForbiddenNames(IReadOnlyList<string> prohibitedNames)
       {
-         _prohibitedNames = prohibitedNames;
+         _prohibitedNames = prohibitedNames.Select(x => x.Trim().ToLower()).ToList();
```
msevestre, line 20: Trim().ToLower(). Create a method so that you don't repeat this logic twice. But I am sure it's already done

msevestre (changes requested):
I feel we are using the wrong base class

src/MoBi.Presentation/DTO/CreateModuleDTO.cs
rwmcintosh: ObjectBaseDTO

PR #984, opened 2023-04-28
973 model creation reference templates

rwmcintosh:


src/MoBi.Core/Domain/Services/SimulationFactory.cs
```diff
@@ -105,10 +108,15 @@ public IModel CreateModelAndValidate(SimulationConfiguration simulationConfigura
          return results.Model;
       }
 
-      public IMoBiSimulation CreateSimulationAndValidate(SimulationConfiguration configuration, string simulationName)
+      public IMoBiSimulation CreateSimulationAndValidate(SimulationConfiguration configurationReferencingBuildingBlocks, string simulationName)
       {
-         var model = CreateModelAndValidate(configuration, simulationName, AppConstants.Captions.CreatingSimulation);
-         return createSimulation(model, configuration, simulationName);
+         var model = CreateModelAndValidate(configurationReferencingBuildingBlocks, simulationName, AppConstants.Captions.CreatingSimulation);
+         return CreateFrom(cloneOf(configurationReferencingBuildingBlocks), model).WithName(simulationName);
```
rwmcintosh, line 114: Creating the clone after the model is validated

src/MoBi.Presentation/Presenter/CreateSimulationConfigurationPresenter.cs
```diff
@@ -100,10 +100,20 @@ private void updateSimulationConfiguration(SimulationConfiguration simulationCon
          var selectedExpressions = individualAndExpressionPresenter.ExpressionProfiles;
          var selectedIndividual = individualAndExpressionPresenter.SelectedIndividual;
 
-         _simulationConfigurationTask.UpdateFrom(simulationConfiguration, moduleConfigurations, selectedIndividual, selectedExpressions);
+         updateFrom(simulationConfiguration, moduleConfigurations, selectedIndividual, selectedExpressions);
```
msevestre, line 103: Personally, i find this code less readable. I do not. know of this is a copy or a clone happening, , the method is only used once and 4 parameters are passed. 4 lines of method declarations . Can't say that I appreciate the benefits over the costs.

src/MoBi.Core/Domain/Services/SimulationFactory.cs
```diff
@@ -105,10 +108,15 @@ public IModel CreateModelAndValidate(SimulationConfiguration simulationConfigura
...
+      }
+
+      private SimulationConfiguration cloneOf(SimulationConfiguration configurationReferencingBuildingBlocks)
+      {
+         return _cloneManager.Clone(configurationReferencingBuildingBlocks);
```
msevestre, line 119: simlarily here. I don't think having a method that calls the method of one member makes it more readable, especially because used once. Personal preference, but the cloneManage.clone is as readable of cloneOf. Adding a method is just a level of indirection

msevestre (approved):

PR #988, opened 2023-05-03
Fixes #987 Import Expression as PSV is on the wrong node

### LINKED ISSUE #987: Import Expression as PSV is on the wrong node
That last context menu should be at the PSV folder level

image

rwmcintosh:


src/MoBi.Presentation/DTO/ObjectBaseDTO.cs
```diff
@@ -158,4 +159,14 @@ public ModuleViewItem(Module module)
          Module = module;
       }
    }
+
+   public class ModuleParameterStartValuesCollectionViewItem : ObjectBaseDTO
```
msevestre, line 163: I don't understand this object. It's the same as the ModuleViewItem above? 🤔

msevestre (changes requested):
I do not think that we need two objects that are virtually identical just of the sake of context menu. Something is a bit off. The IsSatisfy can use another condition maybe?

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForParameterStartValuesFolder.cs
```diff
@@ -0,0 +1,72 @@
...
+         var contextMenu = new ContextMenuForParameterStartValuesFolder(_container);
+         return contextMenu.InitializeWith(viewItem.DowncastTo<ModuleViewItem>(), presenter);
+      }
+
+      public bool IsSatisfiedBy(IViewItem viewItem, IPresenterWithContextMenu<IViewItem> presenter)
+      {
+         if (!(viewItem is ModuleViewItem moduleViewItem))
+            return false;
+
+         return moduleViewItem.TargetAsObject.IsAnImplementationOf<IEnumerable<ParameterStartValuesBuildingBlock>>();
```
rwmcintosh, line 69: Like this?

PR #1005, opened 2023-05-15
972 mobi does not show the right buidling block types even when available

rwmcintosh:


src/MoBi.Core/Domain/Repository/BuildingBlockRepository.cs
```diff
@@ -1,43 +1,83 @@
...
+      IReadOnlyList<MoBiSpatialStructure> SpatialStructureCollection { get; }
+      IReadOnlyList<ObserverBuildingBlock> ObserverBlockCollection { get; }
+      IReadOnlyList<EventGroupBuildingBlock> EventBlockCollection { get; }
+      IReadOnlyList<InitialConditionsBuildingBlock> InitialConditionBlockCollection { get; }
+      IReadOnlyList<ParameterValuesBuildingBlock> ParametersValueBlockCollection { get; }
+      IndividualBuildingBlock IndividualByName(string buildingBlockName);
+      ExpressionProfileBuildingBlock ExpressionProfileByName(string buildingBlockName);
    }
 
    public class BuildingBlockRepository : IBuildingBlockRepository
```
rwmcintosh, line 27: The new functionality of finding a project's building blocks is concentrated here. Most tasks formerly used IMoBiContext to get the project then just used ```AllBuildingBlocks```

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -45,12 +47,14 @@ public class ForbiddenNamesRetriever : IForbiddenNamesRetriever,
       IVisitor<IModelCoreSimulation>
 
    {
-      private readonly IMoBiContext _context;
+      private readonly IBuildingBlockRepository _buildingBlockRepository;
+      private readonly ISimulationRepository _simulationRepository;
       private HashSet<string> _nameHashForVisitor;
 
-      public ForbiddenNamesRetriever(IMoBiContext context)
+      public ForbiddenNamesRetriever(IBuildingBlockRepository buildingBlockRepository, ISimulationRepository simulationRepository)
```
rwmcintosh, line 54: There are a lot of changes like this - removing IMoBiContext, and using IBuildingBlockRepository instead.

Sometimes IMoBiContext was used for another purpose, so couldn't be removed

src/MoBi.Core/Serialization/ORM/Mappers/ProjectMetaDataToProjectMapper.cs
```diff
@@ -61,11 +63,15 @@ public MoBiProject MapFrom(ProjectMetaData projectMetaData)
                }
 
                //then load the rest of the entities
+               var nonProjectBuildingBlocks = new List<IBuildingBlock>();
```
rwmcintosh, line 66: Some new functionality here to implement a kind of project conversion.
If there's only one building block of each type, you get one module, otherwise we group by name and add them by name.

src/MoBi.Core/Serialization/ORM/Mappers/ProjectMetaDataToProjectMapper.cs
```diff
@@ -119,11 +180,40 @@ private void addEntityToProject(EntityMetaData entity, SerializationContext seri
          else if (deserializedEntity.IsAnImplementationOf<Module>())
             _project.AddModule(deserializedEntity as Module);
 
-         else if (deserializedEntity.IsAnImplementationOf<IWithId>())
-            _project.AddBuildingBlock(deserializedEntity as IBuildingBlock);
+         else if (deserializedEntity.IsAnImplementationOf<IndividualBuildingBlock>())
+            _project.AddIndividualBuildingBlock(deserializedEntity as IndividualBuildingBlock);
 
-         else
+         else if (deserializedEntity.IsAnImplementationOf<ExpressionProfileBuildingBlock>())
```
msevestre, line 186: looks like a good use case for pattern matching in C#. This was not available when the code was written but is definitely something wes should use now

msevestre (changes requested):

msevestre:
Looks good in general. One question and maybe one bug

src/MoBi.Core/Domain/Repository/BuildingBlockRepository.cs
```diff
@@ -1,43 +1,83 @@
...
+
+      public IReadOnlyList<ParameterValuesBuildingBlock> ParametersValueBlockCollection => get<ParameterValuesBuildingBlock>();
+
+      public IndividualBuildingBlock IndividualByName(string buildingBlockName)
+      {
+         return IndividualsCollection.FindByName(buildingBlockName);
       }
 
-      private static IEnumerable<TBuildingBlock> moduleBuildingBlocks<TBuildingBlock>(MoBiProject currentProject) where TBuildingBlock : IBuildingBlock
+      public ExpressionProfileBuildingBlock ExpressionProfileByName(string buildingBlockName)
```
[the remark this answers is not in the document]
rwmcintosh: We are using the ByName when configuring a simulation. In that case we have some clones of these building blocks and we want to go back to the like-named building block from the project if the user reconfigures... at least for now.

We do the same for Modules, but that code lives in the MoBiProject because there's no 'ModuleRepository' in the same way as a BuildingBlockRepository.

Due to you pointing this code out, I found that these are not being used, and the same code was still in the project. I am removing it from the project and using this - it was my original intention.

msevestre (approved):
Looks good to me

PR #1008, opened 2023-05-17
Fixes #996 Message when PK-Sim version in not compatible

### LINKED ISSUE #996: Expression profile: Database Query - Message when PK-Sim version in not compatible
Now the message says "you can specify PK-Sim version under Utilities -> Options". PLease add "... Utilities -> Options, tab "Application".

Also include a hint that the application (MoBi) must be restarted in order the changes to take place.

rwmcintosh:
image


rwmcintosh:
That's just a message change for when the correct PKSim.UI.Starter.dll is not found in the PKSim path from MoBi.

PR #1021, opened 2023-06-02
Fixes #855 Include Initial Conditions of the associated protein in Ex…

### LINKED ISSUE #855: Include Initial Conditions of the associated protein in Expression Profile

rwmcintosh:
This won't build without a CI core build, but you can get the jist


src/MoBi.Presentation/Presenter/BuildingBlockWithInitialConditionsPresenter.cs
```diff
@@ -0,0 +1,129 @@
+using System;
```
rwmcintosh, line 1: This is a new base class that has no new code, only pulled up from the former InitialConditionsTask

src/MoBi.Presentation/Presenter/ExpressionProfileInitialConditionsPresenter.cs
```diff
@@ -0,0 +1,31 @@
+using MoBi.Core.Domain.Model;
```
rwmcintosh, line 1: BuildingBlockWithInitialConditionsPresenter subclasses don't have any content

src/MoBi.Core/Commands/AddPathAndValueEntityToBuildingBlockCommand.cs
```diff
@@ -65,7 +65,7 @@ public AddParameterValueToBuildingBlockCommand(PathAndValueEntityBuildingBlock<P
 
    public class AddInitialConditionToBuildingBlockCommand : AddPathAndValueEntityToBuildingBlockCommand<InitialCondition>
    {
-      public AddInitialConditionToBuildingBlockCommand(PathAndValueEntityBuildingBlock<InitialCondition> initialConditionsBuildingBlock, InitialCondition pathAndValueEntity)
+      public AddInitialConditionToBuildingBlockCommand(IBuildingBlock<InitialCondition> initialConditionsBuildingBlock, InitialCondition pathAndValueEntity)
```
rwmcintosh, line 68: This is what saved a bunch of refactoring of commands and tasks. In a lot of places we didn't need the specification of ```PathAndValueEntityBuildingBlock<InitialCondition>```, we really just needed ```IBuildingBlock<InitialCondition>```

src/MoBi.Core/Commands/ChangePathAndValueEntityNameCommand.cs
```diff
@@ -22,7 +23,7 @@ protected ChangePathAndValueEntityNameCommand(TBuildingBlock buildingBlock, Obje
          : base(buildingBlock)
       {
          _newValue = newValue;
-         _originalEntity = buildingBlock[path];
+         _originalEntity = buildingBlock.Single(x => Equals(x.Path, path));
```
rwmcintosh, line 26: Sometimes we don't get to use the cache lookup which was nice.

msevestre: I don't get this change here. Why can't it behave the same as before?
Plus not sure if this is equivalent. a cache may return null if this is not found. This will throw. 
And Maybe this is an actual new interface. ILookupBuildingBlock or sthg. We cannot lose suddenly access by path. Or am I missing the problem?

src/MoBi.Core/Commands/RemovePathAndValueEntityFromBuildingBlockCommand.cs
```diff
@@ -1,23 +1,24 @@
...
       private readonly T _originalEntity;
 
-      public RemovePathAndValueEntityFromBuildingBlockCommand(PathAndValueEntityBuildingBlock<T> parent, ObjectPath path) : base(parent)
+      public RemovePathAndValueEntityFromBuildingBlockCommand(IBuildingBlock<T> parent, ObjectPath path) : base(parent)
       {
          CommandType = AppConstants.Commands.DeleteCommand;
 
          ObjectType = new ObjectTypeResolver().TypeFor<T>();
-         _originalEntity = _buildingBlock[path];
+         _originalEntity = _buildingBlock.Single(x => Equals(x.Path, path));
```
msevestre, line 21: I think we need to do something about this. An abstraction is missing

src/MoBi.Core/Commands/RenameExpressionProfileBuildingBlockCommand.cs
```diff
@@ -39,7 +39,7 @@ protected override void RenameObjectBase(IMoBiContext context)
          // We need to modify the path of all builders when the building block name changes
          // The building block name reflects the molecule name so when that changes, the path of all builders needs to be updated
          // to the new molecule name
-         _expressionProfileBuildingBlock.Each(x =>
+         _expressionProfileBuildingBlock.Each<ExpressionParameter>(x =>
```
msevestre, line 42: We need to add an ExpresionParameters also to the building block in this case.

src/MoBi.Core/Commands/UpdateInitialConditionInBuildingBlockCommand.cs
```diff
@@ -32,8 +33,10 @@ public class UpdateInitialConditionInBuildingBlockCommand : BuildingBlockChangeC
          _scaleDivisor = scaleDivisor;
          _negativeValuesAllowed = negativeValuesAllowed;
 
-         var initialCondition = _buildingBlock[_path];
-         if (initialCondition == null) return;
```
msevestre, line 36: ok clearly the code above is not equivalent if this returns null when not found.

src/MoBi.Presentation/Tasks/Interaction/BaseInitialConditionsTask.cs
```diff
@@ -0,0 +1,228 @@
...
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Formulas;
+using OSPSuite.Core.Domain.Services;
+using OSPSuite.Core.Domain.UnitSystem;
+using OSPSuite.Utility.Collections;
+using OSPSuite.Utility.Extensions;
+
+namespace MoBi.Presentation.Tasks.Interaction
+{
+   public abstract class BaseInitialConditionsTask<TBuildingBlock> : StartValuesTask<TBuildingBlock, InitialCondition> where TBuildingBlock : class, IBuildingBlock<InitialCondition>
```
msevestre, line 25: A bit confused with this base task .We already have base presenter . Canot the concreted presenter instantiate the class with the right generics? It feels like we are duplicating the hierarchy twice

src/MoBi.UI/Views/BaseInitialConditionsView.cs
```diff
@@ -0,0 +1,118 @@
...
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.UnitSystem;
+using OSPSuite.Presentation.Views;
+using OSPSuite.UI.Binders;
+using OSPSuite.UI.Controls;
+using OSPSuite.UI.Extensions;
+
+namespace MoBi.UI.Views
+{
+   public abstract partial class BaseInitialConditionsView<TPresenter> : BasePathAndValueEntityView<InitialConditionDTO, InitialCondition> where TPresenter : IBuildingBlockWithInitialConditionsPresenter
```
msevestre, line 21: Abtract views are nasty. Why do we need this?

msevestre (commented):
Lots of code for me to review at once but
* looks like the expected code change
* I did not expect to see some Abstract view and different implementation. Compositio of view is much more efficient. 
* Presenters with Base..why not. But why also have the same structure for the task . Task could be a generic class IInitialCodition<T> and the presenter would instantiate them (since they know what T is). Maybe I went to quick over but not ideal to duplicate the same structure all the time .I would expect composition to simplify some of this as well
*

src/MoBi.Core/Commands/ChangePathAndValueEntityNameCommand.cs
rwmcintosh: I could have used singleOrDefault instead. But let me see about this lookup by path for both types

msevestre: Sure. Then at least it's equivalent. But I think we are loosing some good UX for development

msevestre (commented):
looks good

src/MoBi.Presentation/DTO/PathAndValueEntityDTO.cs
```diff
@@ -28,7 +28,7 @@ protected PathAndValueEntityDTO(T underlyingObject) : base(underlyingObject)
 
       public string Name
       {
-         get => PathWithValueObject.Name;
+         get => ((IWithName)PathWithValueObject).Name;
```
msevestre, line 31: meh

src/MoBi.Presentation/Tasks/Interaction/ParameterValuesTask.cs
```diff
@@ -164,27 +125,26 @@ public override IMoBiCommand RefreshPathAndValueEntitiesFromBuildingBlocks(Param
...
-
-            if (!HasEquivalentPathAndValueEntity(pathAndValueEntity, parameter))
-               macroCommand.Add(SetDisplayValueWithUnit(pathAndValueEntity, parameter.ConvertToDisplayUnit(parameter.Value), parameter.DisplayUnit, buildingBlock));
-
-            // Evaluating the pathAndValueEntity before the formula is important if the pathAndValueEntity is a constant and the original building block uses a constant formula
-            if (!HasEquivalentFormula(pathAndValueEntity, parameter.Formula))
-               macroCommand.Add(ChangeValueFormulaCommand(buildingBlock, pathAndValueEntity, parameter.Formula.IsConstant() ? null : _cloneManagerForBuildingBlock.Clone(parameter.Formula, buildingBlock.FormulaCache)));
-         });
+         // pathAndValueEntitiesToRefresh.Each(pathAndValueEntity =>
+         // {
```
msevestre, line 129: let's remove this?

msevestre: or why do we need this?

msevestre (approved):

msevestre (approved):

src/MoBi.Presentation/Tasks/Interaction/ParameterValuesTask.cs
rwmcintosh: I didn't want to remove code for refresh until we figure out how it's going to work. 

We just need to provide a spatial structure and molecule building block and all this code will work again.

I think that is somewhat the plan right now. Refresh would still work, but you have to pick the building blocks to refresh from.

PR #1028, opened 2023-06-09
update core to 12.0.191.

rwmcintosh:


src/MoBi.Core/Domain/UnitSystem/MoBiDimensionFactory.cs
```diff
@@ -8,14 +8,14 @@ namespace MoBi.Core.Domain.UnitSystem
 {
    public interface IMoBiDimensionFactory : IDimensionFactory
    {
-      IDimension TryGetDimensionCaseInsensitive(string dimensionName);
```
rwmcintosh, line 11: The old signature threw me a curve ball. In the implementation at least the parameter is named appropriately

Issue #1017, opened 2023-05-31
Dialog for editing parent container: show module name


PavelBal:
![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/205403c7-9229-4b79-bf3a-4b967f8aa36e)

Add name of the module to the name of the BB for clarity.

PR #1029, opened 2023-06-12
1016 show the name of the module in the tab of a bb

rwmcintosh:


src/MoBi.Presentation/Presenter/EditEventBuilderPresenter.cs
```diff
@@ -112,7 +112,7 @@ public void Edit(EventBuilder eventBuilder, IReadOnlyList<IObjectBase> existingO
 
       public void Edit(EventBuilder eventBuilder)
       {
-         Edit(eventBuilder, eventBuilder.ParentContainer.Children);
+         Edit(eventBuilder, eventBuilder.ParentContainer?.Children);
```
rwmcintosh, line 115: We had a couple others like this from another refactoring. I found them while opening editors for each type to check the captions.

rwmcintosh:
If you have a BB with module parent you get the 'ModuleName - BBName' format
image

Otherwise, you just get the BBName
image

msevestre (commented):
see comment

rwmcintosh:
See PR in core - https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/2033

src/MoBi.Presentation/Mappers/SpatialStructureToSpatialStructureDTOMapper.cs
```diff
@@ -24,6 +25,7 @@ public SpatialStructureDTO MapFrom(SpatialStructure spatialStructure)
          dto.TopContainer = spatialStructure.TopContainers.MapAllUsing(_containerToDTOContainerMapper);
          dto.Neighborhoods = _containerToDTOContainerMapper.MapFrom(spatialStructure.NeighborhoodsContainer);
          dto.MoleculeProperties = _containerToDTOContainerMapper.MapFrom(spatialStructure.GlobalMoleculeDependentProperties);
+         dto.Name = spatialStructure.DisplayName;
```
rwmcintosh, line 28: This is the change for #1017. The ```dto.Name``` usually takes the ```objectBase.Name``` during mapping, but we are overriding that default in this mapper. This mapper is used whenever you need to select a spatial structure in a tree view from among all spatial structures in the building block repository.

Issue #1015, opened 2023-05-31
Add context menu "new building block" to IC and PV BBs


PavelBal:
I would expect to have "right click -> add new"

![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/1cb4eca6-620e-43b6-a464-2dd7a9359173)

PR #1041, opened 2023-06-19
Add context menu "new building block" to IC and PV BBs #1015

rwmcintosh:
Re-using the standard "AddBuildingBlocks" dialog to add either Parameter Values or Initial Conditions building blocks when a user right clicks on the folder node.

Also, double click expands/contracts the folder node similar to other collector/folder nodes

### DIFF
--- a/src/MoBi.UI/Views/AddBuildingBlocksToModuleView.cs
+++ b/src/MoBi.UI/Views/AddBuildingBlocksToModuleView.cs
@@ -1,10 +1,13 @@
-﻿using DevExpress.XtraLayout;
+﻿using System.Collections.Generic;
+using System.Drawing;
+using DevExpress.XtraLayout;
 using DevExpress.XtraLayout.Utils;
 using MoBi.Presentation.DTO;
 using MoBi.Presentation.Presenter;
 using MoBi.Presentation.Views;
 using OSPSuite.DataBinding;
 using OSPSuite.DataBinding.DevExpress;
+using OSPSuite.Utility.Extensions;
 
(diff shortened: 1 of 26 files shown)


src/MoBi.UI/Views/AddBuildingBlocksToModuleView.cs
```diff
@@ -32,13 +44,46 @@ protected override void StartValueCheckChanged(bool enabled, LayoutControlItem n
...
+      }
+
+      private void adjustHeight(int heightAdjustment)
+      {
+         Size = new Size(Width, Height + heightAdjustment);
+      }
+
+      private void hideAllItemsExcept(List<BaseLayoutItem> itemsToShow)
+      {
+         var initialGroupHeight = createBuildingBlocksGroup.Size.Height;
```
msevestre, line 73: what's with this height here? I don't quite get it

msevestre (commented):
I am not sure I get the initial height stuff. 
Also to check on different monitors. Manual resizing =>always issues

src/MoBi.UI/Views/AddBuildingBlocksToModuleView.cs
rwmcintosh: That's the height of the layout group before hiding many items.

Later we take the difference and then apply that difference to the dialog height to adjust.

rwmcintosh:
I can't try different physical monitors, but adjusting the resolution and scaling on my monitor indicates that this works at different scales and resolutions.

PR #1044, opened 2023-06-29
Fixes #1037 "Extend" of IC

### LINKED ISSUE #1037: "Extend" of IC
- Load “simple.pkml” simulation
- Create new module “BrainModule” with “Spatial Structure” BB
- Delete the top container “Organism”
- Add a new top container “Organ -> Physical container” “Brain” to the module
- Set “Parent path” of container “Brain” in module “BrainModule” to “Organism|Liver”

- Create a new IC BB in the module "BrainModule"
- Click "Extend"

## Current behavior
- Nothing happens

## Expected behavior

- The user select one Spatial Structure - BB (from any module) and one Molecules BB (from any module). IC entries for this combination are created. By default, all present (?), initial values are taken from the molecules BB.

rwmcintosh:
If you extend in either of the two use cases and there is only one matching bb of each required type, then you don't get any prompt, the IC's are just extended using the only possible choices.

Otherwise...

When you extend an expression profile:
image

When you extend an IC bb:
image

This view and presenter were mainly used as 'Create IC' view before. We no longer select spatial structures or molecules when creating them now, but the concept is reused.


msevestre:
can we make thos Views not has tall? This looks ugly :)

msevestre:
Also I would add a description field explaining why the spt stru would be used

src/MoBi.Presentation/DTO/SelectSpatialStructureAndMoleculesDTO.cs
```diff
@@ -0,0 +1,38 @@
...
+{
+   public class SelectSpatialStructureAndMoleculesDTO : DxValidatableDTO
+   {
+      public MoleculeBuildingBlock Molecules { get; set; }
+      public MoBiSpatialStructure SpatialStructure { get; set; }
+
+      public SelectSpatialStructureAndMoleculesDTO(bool moleculeRequired = true)
+      {
+         MoleculeRequired = moleculeRequired;
+         Rules.Add(AllRules.MoleculeSelected);
```
msevestre, line 17: should we only add the rule if molecule is required and not even saved this parameter

src/MoBi.Presentation/Presenter/SelectBuildingBlocksForExtendPresenter.cs
```diff
@@ -0,0 +1,74 @@
...
+         var dto = new SelectSpatialStructureAndMoleculesDTO(moleculeRequired)
+         {
+            Molecules = moleculeBuildingBlock,
+            SpatialStructure = spatialStructure
+         };
+         return dto;
+      }
+
+      public MoBiSpatialStructure SelectedSpatialStructure => _dto.SpatialStructure;
+      public MoleculeBuildingBlock SelectedMoleculeBuildingBlock => _dto.Molecules;
```
msevestre, line 62: from a cohesion stand point, we have two different ways of implementing the same thing. one is => the other one is 
{

xx

}}

Let's make ti consistent by inlining  public IEnumerable<MoleculeBuildingBlock> GetMolecules()

and in fact, this should be a get as well. Why is this not a get like SelectedSpatialStructure?

msevestre: and also the other one

msevestre, line 62: spacing missing here.

src/MoBi.Presentation/Presenter/SelectBuildingBlocksForExtendPresenter.cs
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      private void setViewCaption(bool moleculeRequired)
+      {
+         _view.Caption = moleculeRequired ? AppConstants.Captions.SelectSpatialStructureAndMolecules : AppConstants.Captions.SelectSpatialStructure;
+      }
+
+      private SelectSpatialStructureAndMoleculesDTO createDTO(MoleculeBuildingBlock moleculeBuildingBlock, MoBiSpatialStructure spatialStructure, bool moleculeRequired)
+      {
+         var dto = new SelectSpatialStructureAndMoleculesDTO(moleculeRequired)
```
msevestre, line 53: no need for local dto. Return right away

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -130,6 +135,31 @@ public IMoBiCommand UpdateInitialConditionScaleDivisor(TBuildingBlock buildingBl
...
+      public void ExtendExpressionProfileInitialConditions(ExpressionProfileBuildingBlock buildingBlock)
+      {
+         var (spatialStructure, _) = SelectBuildingBlocksForExtend(moleculeRequired: false);
+         if (spatialStructure == null)
+            return;
+
+         var moleculeBuilder = _moleculeBuilderTask.CreateDefault(buildingBlock.MoleculeName);
+         var startFormula = defaultStartFormulaFrom(moleculeBuilder, buildingBlock);
+
+         var newStartValues = spatialStructure.PhysicalContainers.Select(container => _initialConditionsCreator.CreateInitialCondition(container, moleculeBuilder, startFormula)).ToList();
```
msevestre, line 147: rename startValues. We do not have this anymore

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -130,6 +135,31 @@ public IMoBiCommand UpdateInitialConditionScaleDivisor(TBuildingBlock buildingBl
          return new UpdateInitialConditionScaleDivisorCommand(buildingBlock, initialCondition, newScaleDivisor, oldScaleDivisor).Run(Context);
       }
 
+      public void ExtendExpressionProfileInitialConditions(ExpressionProfileBuildingBlock buildingBlock)
+      {
+         var (spatialStructure, _) = SelectBuildingBlocksForExtend(moleculeRequired: false);
+         if (spatialStructure == null)
+            return;
+
+         var moleculeBuilder = _moleculeBuilderTask.CreateDefault(buildingBlock.MoleculeName);
```
msevestre, line 144: maybe explain that this is a temp builder as comment?

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -130,6 +135,31 @@ public IMoBiCommand UpdateInitialConditionScaleDivisor(TBuildingBlock buildingBl
...
+      }
+
+      private IFormula defaultStartFormulaFrom(MoleculeBuilder moleculeBuilder, ExpressionProfileBuildingBlock buildingBlock)
+      {
+         return buildingBlock.InitialConditions.Any() ? mostUsedFormula(buildingBlock.InitialConditions) : moleculeBuilder.DefaultStartFormula;
+      }
+
+      private IFormula mostUsedFormula(IReadOnlyCollection<InitialCondition> initialConditions)
+      {
+         return initialConditions.GroupBy(x => x.Formula).OrderBy(x => x.Count()).First().Key;
```
msevestre, line 160: add a comment saying that you know there is one?

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -130,6 +135,31 @@ public IMoBiCommand UpdateInitialConditionScaleDivisor(TBuildingBlock buildingBl
...
+
+         var newStartValues = spatialStructure.PhysicalContainers.Select(container => _initialConditionsCreator.CreateInitialCondition(container, moleculeBuilder, startFormula)).ToList();
+
+         updateDefaultIsPresentToFalseForSpecificExtendedValues(newStartValues, buildingBlock.InitialConditions.ToList());
+         AddCommand(Extend(newStartValues, buildingBlock));
+      }
+
+      private IFormula defaultStartFormulaFrom(MoleculeBuilder moleculeBuilder, ExpressionProfileBuildingBlock buildingBlock)
+      {
+         return buildingBlock.InitialConditions.Any() ? mostUsedFormula(buildingBlock.InitialConditions) : moleculeBuilder.DefaultStartFormula;
```
msevestre, line 155: this is a bit weird. I would expect the DefaultStartFormula to be used if there is one and if not, we used the mostUsedOne. Here the code reads the opposite

tests/MoBi.Tests/Presentation/SelectSpatialStructureAndMoleculesPresenter.cs
```diff
@@ -0,0 +1,102 @@
...
+         sut = new SelectBuildingBlocksForExtendPresenter(_view, _buildingBlockRepository);
+      }
+   }
+
+   public class When_Selecting_building_blocks_and_the_view_is_canceled : concern_for_SelectSpatialStructureAndMoleculesPresenter
+   {
+      protected override void Context()
+      {
+         base.Context();
+         A.CallTo(() => _buildingBlockRepository.MoleculeBlockCollection).Returns(new List<MoleculeBuildingBlock> { new MoleculeBuildingBlock() });
```
msevestre, line 31: you can just use new [] 
less code, just better

src/MoBi.Presentation/Presenter/SelectBuildingBlocksForExtendPresenter.cs
```diff
@@ -0,0 +1,74 @@
...
+using MoBi.Presentation.DTO;
+using MoBi.Presentation.Views;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Presentation.Presenters;
+
+namespace MoBi.Presentation.Presenter
+{
+   public interface ISelectBuildingBlocksForExtendPresenter : IDisposablePresenter
+   {
+      IEnumerable<MoleculeBuildingBlock> GetMolecules();
```
msevestre, line 15: Is the underlying prop IEnumerable or IReadonly?

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
rwmcintosh: I think we decided on the opposite didn't we? 
That's what I understood yesterday. There will be a defaultStartFormula because that's the way the basic builder is configured a ConstantFormula with value 0.

src/MoBi.Presentation/Tasks/Interaction/StartValuesTask.cs
```diff
@@ -305,5 +278,26 @@ public override IMoBiCommand GetAddCommand(TBuildingBlock itemToAdd, Module pare
...
+
+         using (var selectorPresenter = Context.Resolve<ISelectBuildingBlocksForExtendPresenter>())
+         {
+            selectorPresenter.SelectBuildingBlocksForExtend(moleculeRequired);
+            if (selectorPresenter.SelectedSpatialStructure == null || (moleculeRequired && selectorPresenter.SelectedMoleculeBuildingBlock == null))
+            {
+               return (null, null);
+            }
+
+            return (selectorPresenter.SelectedSpatialStructure, selectorPresenter.SelectedMoleculeBuildingBlock);
```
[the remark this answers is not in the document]
rwmcintosh: That was in case a user selects one but not  the other. But that actually can't happen with the validation anyway. On view cancel, the selections are cleared.

tests/MoBi.Tests/Presentation/SelectSpatialStructureAndMoleculesPresenter.cs
rwmcintosh: Need to teach Copilot that one.

src/MoBi.Presentation/Presenter/SelectBuildingBlocksForExtendPresenter.cs
rwmcintosh: They were IEnumerable, but they will become IReadOnly

rwmcintosh:
> Also I would add a description field explaining why the spt stru would be used

Not sure what you mean here. Are you talking about adding a description on the view? There never was one before.

rwmcintosh:
image

msevestre (approved):
This makes a lot of sense to me with this change

PR #1049, opened 2023-07-14
Fixes #1014 It is possible to delete a module used by a simulation

### LINKED ISSUE #1014: It is possible to delete a module used by a simultion
- Load simulation "Aciclovir.pkml" from the r-package examples
- Open the building block "Spatial structure" from the created module
- Delete module

1) Simulation is still there - deleting a module that is used in a simulation should not be possible
2) The spatial structures BB is still open, while the module does not exist any more :) 

image

rwmcintosh:


src/MoBi.Presentation/Tasks/CheckNameVisitor.cs
```diff
@@ -193,7 +189,7 @@ public void Visit(IObjectBase objectBase)
 
          if (string.Equals(_oldName, objectBase.Name))
          {
-            _changes.Add(objectBase, _buildingBlock, new EditObjectBasePropertyInBuildingBlockCommand(_namePropertyName, _newName, _oldName, objectBase, _buildingBlock),
+            _changes.Add(objectBase, _buildingBlock, new RenameObjectBaseCommand(objectBase, _newName, _buildingBlock),
```
rwmcintosh, line 192: This was the line triggered when you had that little rename helper. Formerly, the name of the building block would change, but not the name in the simulation. You'd get an indication of the change in the simulation explorer.

Now when a building block shares names with other building blocks, and you chose to rename all of them together, that will trigger renaming the same building block in all simulations too.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForModule.cs
```diff
@@ -113,8 +113,11 @@ public void LoadBuildingBlocksFromTemplateToModule(Module module)
 
       public void RemoveModule(Module module)
       {
-         context.AddToHistory(new RemoveModuleCommand(module)
-            .Run(context));
+         var referringSimulations = Context.CurrentProject.SimulationsCreatedUsing(module);
```
rwmcintosh, line 116: There was no test for module in use before. But to get it working, we need to sync the names.
Syncing the names is the bulk of the work here.

src/MoBi.Core/Commands/RenameObjectBaseCommand.cs
```diff
@@ -48,13 +51,23 @@ protected virtual void RenameObjectBase(IMoBiContext context)
...
-            renameBuildingBlockTask.RenameInSimulationUsingTemplateBuildingBlock(_objectBase.DowncastTo<IBuildingBlock>());
+            doWithRenameInSimulationTask(task => task.RenameInSimulationUsingTemplateBuildingBlock(OldName, _objectBase.DowncastTo<IBuildingBlock>()), context);
+         }
+         else if (_objectBase.IsAnImplementationOf<Module>())
+         {
+            doWithRenameInSimulationTask(task => task.RenameInSimulationUsingTemplateModule(OldName, _objectBase.DowncastTo<Module>()), context);
          }
       }
 
+      private void doWithRenameInSimulationTask(Action<IRenameInSimulationTask> action, IMoBiContext context)
```
msevestre, line 65: I am not sure I get this. What about you instantiate this tast in the calling method and then just call the method instead of creating a method with a delegate. Can't see the benefit of this indirectionhere

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -20,19 +21,31 @@ public interface IMoBiSimulation : IWithDiagramFor<IMoBiSimulation>, ISimulation
       SimulationPredictedVsObservedChart PredictedVsObservedChart { get; set; }
       SimulationResidualVsTimeChart ResidualVsTimeChart { get; set; }
 
-      string ParameterIdentificationWorkingDirectory { get; set; }
```
msevestre, line 23: why is this gone? Not used?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -73,9 +86,38 @@ public void RemoveAllChartTemplates()
...
-         //TODO SIMULATION_CONFIGURATION
+         // We can consider the building block in-use if it belongs to a module that is in use.
+         if (templateBuildingBlock.Module != null)
+            return usesModuleBuildingBlock(templateBuildingBlock);
+
+         // Simple name match for building blocks that do not belong to a module
+         switch (templateBuildingBlock)
+         {
+            case IndividualBuildingBlock individualBuildingBlock:
+               return Equals(Configuration.Individual?.Name, individualBuildingBlock.Name);
```
msevestre, line 97: string.equals?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -73,9 +86,38 @@ public void RemoveAllChartTemplates()
...
          return false;
-         // return MoBiBuildConfiguration.BuildingInfoForTemplate(templateBuildingBlock) != null;
+      }
+
+      private bool usesModuleBuildingBlock(IBuildingBlock templateBuildingBlock)
+      {
+         return BuildingBlocks.Any(buildingBlock => buildingBlock.Module.IsNamed(templateBuildingBlock.Module.Name) && buildingBlock.IsTemplateMatchFor(templateBuildingBlock));
+      }
+
+      public IReadOnlyList<IBuildingBlock> BuildingBlocks
```
msevestre, line 110: That looks like a property but underneath the hood, there is a lot of stuff going on with select many, concat etc.. Wonder if we should make it a function instead so that it's clear that there is something going on? Thoughts?

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
```diff
@@ -0,0 +1,74 @@
...
+      private readonly IMoBiProjectRetriever _projectRetriever;
+
+      public RenameInSimulationTask(IMoBiProjectRetriever projectRetriever)
+      {
+         _projectRetriever = projectRetriever;
+      }
+
+      public void RenameInSimulationUsingTemplateBuildingBlock(string oldName, IBuildingBlock templateBuildingBlock)
+      {
+         getBuildingBlocksWithMatchingNameAndType(_projectRetriever.Current.Simulations, oldName, templateBuildingBlock).Each(simulationAndBuildingBlock =>
```
msevestre, line 39: not sure why you are passing the simulations as parameter to this method since it comes from member . 
Dont make the method static and you are good to go

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
```diff
@@ -0,0 +1,74 @@
...
+
+      private static IReadOnlyList<(IMoBiSimulation Simulation, IBuildingBlock BuildingBlock)> getBuildingBlocksWithMatchingNameAndType(IReadOnlyList<IMoBiSimulation> simulations, string nameToMatch, IBuildingBlock templBuildingBlock)
+      {
+         return simulations.SelectMany(simulation =>
+         {
+            // We cannot test for in use using the members of simulation. The standard in-use or created-by tests match based on names.
+            // Here, the names do not match because the building block has already been renamed
+            return simulation.BuildingBlocks.Where(buildingBlock => buildingBlock.IsTemplateMatchFor(templBuildingBlock, nameToMatch))
+               .Select(b => (simulation, b));
+         }).ToList();
```
msevestre, line 54: questionable if the tolist is rquired here for a private method used to iterate

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
```diff
@@ -0,0 +1,74 @@
...
+      }
+
+      public void RenameInSimulationUsingTemplateModule(string oldName, Module templateModule)
```
msevestre, line 57: rename oldName => oldModuleName

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
```diff
@@ -0,0 +1,74 @@
...
+      {
+         _projectRetriever.Current.Simulations.Each(simulation =>
+         {
+            // We cannot test for in use using the members of simulation. The standard in-use or created-by tests match based on names.
+            // Here, the names do not match because the module has already been renamed
+            simulation.Configuration.ModuleConfigurations.Select(x => x.Module).Where(module => module.IsNamed(oldName))
```
msevestre, line 63: feels like we need some accessor method. 
this Configuration.ModuleConfigurations.Select(x => x.Module) seems to be used multiple times. Should we add a Modules{get} in the simulation?

msevestre: also here it's a findByName kind of guy

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
```diff
@@ -0,0 +1,74 @@
...
+               .Each(module => renameModules(simulation, module, templateModule.Name));
```
msevestre, line 64: can it be null?

src/MoBi.Core/Commands/RenameObjectBaseCommand.cs
rwmcintosh: It's just about removing duplicates of context.Resolve. I don't want to resolve the task if it's not going to be used and I didn't want to just duplicate that line. But, it's probably a lot more straightforward with

```
if(something)
  resolveThisThing
else if (something else)
  resolveThisThing
```

src/MoBi.Presentation/DTO/ModuleConfigurationDTO.cs
```diff
@@ -1,44 +1,54 @@
...
       {
-         return Equals(_moduleConfiguration, moduleConfiguration);
+         return Equals(ModuleConfiguration, moduleConfiguration);
       }
 
       public bool HasInitialConditions => !NullPathAndValueEntityBuildingBlocks.NullInitialConditions.Equals(SelectedInitialConditions);
 
       public bool HasParameterValues => !NullPathAndValueEntityBuildingBlocks.NullParameterValues.Equals(SelectedParameterValues);
+
+      public event PropertyChangedEventHandler PropertyChanged;
```
msevestre, line 52: ?? why this weird implementation of PRopertyChanged
Event should always be initialized otherwise you have a crash
Implement the NotifyDTO that handles eferythinf for you anyways

msevestre (changes requested):

src/MoBi.Core/Domain/Model/MoBiProject.cs
```diff
@@ -147,5 +145,14 @@ public IEnumerable<IObjectBase> All()
       {
          return All<IObjectBase>().Union(Simulations);
       }
+
+      /// <summary>
+      /// Returns a list of simulations that have a module where the name matches <paramref name="module"/>
+      /// This indicates that the <paramref name="module"/> was used as a template for the simulation
+      /// </summary>
+      public IReadOnlyList<IMoBiSimulation> SimulationsCreatedUsing(Module module)
```
msevestre, line 153: SimulationUsing? 
Why created?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -20,19 +21,31 @@ public interface IMoBiSimulation : IWithDiagramFor<IMoBiSimulation>, ISimulation
...
+
       /// <summary>
-      ///    Returns true if the simulation as created using the <paramref name="templateBuildingBlock" /> otherwise false.
+      /// Returns true if the simulation uses <paramref name="templateBuildingBlock"/>.
+      /// For module building blocks, the test checks if the <paramref name="templateBuildingBlock"/> is a member of
+      /// a module that's in use <see cref="Uses"/>. If the <paramref name="templateBuildingBlock"/> is not a module building block
+      /// the test is based on building block name and type.
       /// </summary>
+      /// <returns>True if the simulation uses the <paramref name="templateBuildingBlock"/>, otherwise false</returns>
       bool IsCreatedBy(IBuildingBlock templateBuildingBlock);
```
msevestre, line 36: shouldn't we continues with Uses?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
rwmcintosh: yes, not used

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -73,9 +86,38 @@ public void RemoveAllChartTemplates()
...
+         if (templateBuildingBlock.Module != null)
+            return usesModuleBuildingBlock(templateBuildingBlock);
+
+         // Simple name match for building blocks that do not belong to a module
+         switch (templateBuildingBlock)
+         {
+            case IndividualBuildingBlock individualBuildingBlock:
+               return Equals(Configuration.Individual?.Name, individualBuildingBlock.Name);
+            case ExpressionProfileBuildingBlock expressionProfileBuildingBlock:
+               return Configuration.ExpressionProfiles.Any(x => Equals(x.Name, expressionProfileBuildingBlock.Name));
```
msevestre, line 99: ExistByName

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -73,9 +86,38 @@ public void RemoveAllChartTemplates()
...
+      private bool usesModuleBuildingBlock(IBuildingBlock templateBuildingBlock)
+      {
+         return BuildingBlocks.Any(buildingBlock => buildingBlock.Module.IsNamed(templateBuildingBlock.Module.Name) && buildingBlock.IsTemplateMatchFor(templateBuildingBlock));
+      }
+
+      public IReadOnlyList<IBuildingBlock> BuildingBlocks
+      {
+         get
+         {
+            var buildingBlocks = Configuration.ModuleConfigurations.SelectMany(moduleConfiguration => moduleConfiguration.Module.BuildingBlocks).Concat(Configuration.ExpressionProfiles).ToList();
```
msevestre, line 114: maybe we should cache some of this shit (e.g. all building blocs used in a simulation)(. This never changes once the simulation is created and could be saved in the config or simulations. THis is repeated just too often I find

msevestre (changes requested):

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
rwmcintosh: can what be null?

src/MoBi.Core/Commands/RenameObjectBaseCommand.cs
msevestre: sure. But the task is the same so put it before the if and it's a clear call then
Plus use switch on type so that you don't downcast etc

src/MoBi.Core/Domain/Services/RenameInSimulationTask.cs
msevestre: oh well I assumed it was a find byname because you can;t havetwo modules with the same name right? You can only have one at most but we are looping. Unexpected

src/MoBi.Core/Domain/Model/MoBiProject.cs
rwmcintosh: That just matches the existing overload, but they can both be renamed

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
rwmcintosh: I'm with you on making an accessor that gives me all the modules like ```simulation.Modules {get}```. But caching them? I don't think it takes too long to calculate the modules for a simulation, and from that all the building blocks for the modules.

This way it won't be repeated.

msevestre: not the modules but the building blocks. THis never changes once it's created. Probalby not required unless this access is done over all parameters withing a loop etc.. . If we cached, we could go back to get get method and not a function

msevestre (approved):

PR #1055, opened 2023-07-25
Fixes #1051 Expression doubled when configuring a simulation

### LINKED ISSUE #1051: Expression doubled when configuring a simulation
When you create a simulation, the dialog works correctly (left/right panes show only one expression).

After creating the simulation, if you configure it, then you'll see the expression appear both in the left and right pane.

image

rwmcintosh:
When configured an existing simulation, the node for the selected expression appeared in the unselected expression pane.

This was just an order of operations problem.

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
@@ -45,21 +45,21 @@ namespace MoBi.Presentation.Presenter
       public void Edit(SimulationConfiguration simulationConfiguration)
       {
          _individualSelectionDTO = _selectedIndividualDTOMapper.MapFrom(simulationConfiguration.Individual);
-
+         initializeSelectedExpressions(simulationConfiguration);
          addUnusedExpressionsToSelectionView();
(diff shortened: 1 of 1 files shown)


src/MoBi.Presentation/Presenter/EditIndividualAndExpressionConfigurationsPresenter.cs
```diff
@@ -45,21 +45,21 @@ public class EditIndividualAndExpressionConfigurationsPresenter : AbstractSubPre
...
-         addUsedExpressionsToSelectedView(simulationConfiguration);
+         addUsedExpressionsToSelectedView();
 
          _view.BindTo(_individualSelectionDTO);
       }
 
-      private void addUsedExpressionsToSelectedView(SimulationConfiguration simulationConfiguration)
+      private void initializeSelectedExpressions(SimulationConfiguration simulationConfiguration)
+      {
+         _buildingBlockRepository.ExpressionProfileCollection.Where(x => x.NameIsOneOf(simulationConfiguration.ExpressionProfiles.AllNames())).Each(_selectedExpressions.Add);
```
rwmcintosh, line 57: This new step initializes this ```_selectedExpressions``` before either panel in the view is initialized.

Yuri05 (approved):

PR #1057, opened 2023-07-28
Fixes #1052 Update/Commit - sync red/green checks with building block versions

### LINKED ISSUE #1052: Update/Commit - sync red/green checks with building block versions
See https://github.com/Open-Systems-Pharmacology/MoBi/issues/1050#issue-1804975654

rwmcintosh:
This is just the start of working on update/commit.

For this first step, I just want to get the green/red checks working when the user updates the template building block.

In that case it's just a simple check of BuildingBlock version.

But you do not have the direct link between simulation building blocks and template building blocks that you had in version 11. Instead, we have to find the matching template based on parent module, building block name and type. Once found, the versions can be compared.

### DIFF
--- a/src/MoBi.Core/Services/SimulationConfigurationFactory.cs
+++ b/src/MoBi.Core/Services/SimulationConfigurationFactory.cs
@@ -30,8 +30,7 @@ namespace MoBi.Core.Services
             SimulationSettings = _cloneManager.Clone(_projectRetriever.Current.SimulationSettings)
          };
 
-         _calculationMethodRepository.All().Each(
-            cm => simulationConfiguration.AddCalculationMethod(_cloneManager.Clone(cm)));
+         _calculationMethodRepository.All().Each(simulationConfiguration.AddCalculationMethod);
 
          return simulationConfiguration;
       }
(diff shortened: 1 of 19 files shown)


src/MoBi.Core/Services/SimulationConfigurationFactory.cs
```diff
@@ -30,8 +30,7 @@ public SimulationConfiguration Create()
             SimulationSettings = _cloneManager.Clone(_projectRetriever.Current.SimulationSettings)
          };
 
-         _calculationMethodRepository.All().Each(
-            cm => simulationConfiguration.AddCalculationMethod(_cloneManager.Clone(cm)));
+         _calculationMethodRepository.All().Each(simulationConfiguration.AddCalculationMethod);
```
rwmcintosh, line 33: @Yuri05 that's just a continuation of the earlier PR where we don't want to clone calculation methods anymore.

src/MoBi.Presentation/Nodes/TreeNodeFactory.cs
```diff
@@ -174,13 +175,7 @@ private void createAndAddNodeUnder(ITreeNode rootNode, IBuildingBlock buildingBl
...
       {
-         var statusIcon = ApplicationIcons.GreenOverlayFor(buildingBlock.Icon);
-         // var statusIcon = buildingBlockInfo.BuildingBlockChanged
-         //    ? ApplicationIcons.RedOverlayFor(buildingBlock.Icon)
-         //    : ApplicationIcons.GreenOverlayFor(buildingBlock.Icon);
-
-         return CreateFor(buildingBlock)
-            .WithIcon(statusIcon);
+         return CreateFor(buildingBlock);
       }
```
Yuri05, line 179: why not just call `CreateFor(buildingBlock)` instead of `createWithIcon(buildingBlock)`?
Then we can just remove this function

Yuri05 (approved):

PR #1058, opened 2023-08-03
#875 time column showing in chart although it should not

### LINKED ISSUE #875: Chart Editor: Link Data to Simulation also adds "Time"
When I select a simulation output with the option "Link Data to Simulations", the "Time" repository is also added:

image

georgeDaskalakis:
closes #875

### DIFF
--- a/tests/MoBi.Tests/Presentation/SimulationChartPresenterSpecs.cs
+++ b/tests/MoBi.Tests/Presentation/SimulationChartPresenterSpecs.cs
@@ -136,11 +136,26 @@ namespace MoBi.Presentation
       {
          _observedDataAddedEventCounter.ShouldBeEqualTo(1);
       }
+   }
+
+   public class When_adding_observed_data_to_the_char : concern_for_SimulationChartPresenter
+   {
+      protected override void Context()
+      {
+         base.Context();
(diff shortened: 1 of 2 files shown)


tests/MoBi.Tests/Presentation/SimulationChartPresenterSpecs.cs
```diff
@@ -143,4 +137,25 @@ public void the_event_must_have_been_called_once()
          _observedDataAddedEventCounter.ShouldBeEqualTo(1);
       }
    }
+
+   public class When_adding_observed_data_to_the_char : concern_for_SimulationChartPresenter
```
Yuri05, line 141: typo: char=>chart

georgeDaskalakis: corrected

Yuri05:
> @Yuri05 if everything else seems OK, should we merge?

I think so. I cannot judge if the change does not have any side effects, though.

georgeDaskalakis:
> > @Yuri05 if everything else seems OK, should we merge?
> 
> I think so. I cannot judge if the change does not have any side effects, though.

Hmm OK, let's discuss briefly next week with @msevestre . Tbh, I cannot think of a scenario where we really need to explicitly show the time column.

PR #1061, opened 2023-08-09
Fixes #1053 Add ParameterValueTracker to simulation and use it to sto…

### LINKED ISSUE #1053: Update/Commit - add ParameterValueTracker to simulation and use it to store original values
See https://github.com/Open-Systems-Pharmacology/MoBi/issues/1050#issue-1804975654

This is a list of path and values. Each item is used to keep track of the original parameter value when an update happens on a simulation parameter.

rwmcintosh:


src/MoBi.Core/Services/QuantityToParameterValueMapper.cs
```diff
@@ -0,0 +1,36 @@
...
+      public QuantityToParameterValueMapper(IEntityPathResolver entityPathResolver)
+      {
+         _entityPathResolver = entityPathResolver;
+      }
+
+      public ParameterValue MapFrom(IQuantity input)
+      {
+         var parameterValue = new ParameterValue
+         {
+            Path = _entityPathResolver.ObjectPathFor(input),
```
rwmcintosh, line 25: These are the properties I could imagine changing in the context of simulation. I don't even think you can change Dimension or DisplayUnit

src/MoBi.Core/Commands/SetQuantityPropertyInSimulationCommandBase.cs
```diff
@@ -20,6 +21,11 @@ protected SetQuantityPropertyInSimulationCommandBase(TQuantity quantity, IMoBiSi
          CommandType = AppConstants.Commands.EditCommand;
       }
 
+      protected override void ExecuteWith(IMoBiContext context)
+      {
+         context.Resolve<IQuantityValueInSimulationChangeTracker>().TrackChanges(_quantity, _simulation, () => base.ExecuteWith(context));
```
msevestre, line 26: ok that's funky. Calling the base in an action? I have never seen this in my life.

rwmcintosh: The purpose of the action is just so I can wrap it with a step to create the original value before the action if one doesn't exist, then a step to check if the result of the action makes the value equal to the original value.

So, what do you suggest? I only need to do these two steps, and I can definitely unwrap this call and do two separate ones.

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -122,6 +133,24 @@ public IReadOnlyList<IBuildingBlock> BuildingBlocks()
          return buildingBlocks;
       }
 
+      public IReadOnlyList<ParameterValue> OriginalQuantityValues => _quantityValueCache.ToList();
```
msevestre, line 136: I'd suggest to returnd IReadOnlyDictionary or another type in this case which is IEnumerable too so that you don't have to ToList all the time.

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -122,6 +133,24 @@ public IReadOnlyList<IBuildingBlock> BuildingBlocks()
...
+
+      public void AddOriginalQuantityValue(ParameterValue parameterValue)
+      {
+         if (_quantityValueCache[parameterValue.Path] == null)
```
msevestre, line 140: why the check here?. Shoudn't you always overwrite?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -122,6 +133,24 @@ public IReadOnlyList<IBuildingBlock> BuildingBlocks()
...
+            _quantityValueCache[parameterValue.Path] = parameterValue;
+      }
+
+      public void RemoveOriginalQuantityValue(ObjectPath objectPath)
+      {
+         _quantityValueCache.Remove(objectPath);
+      }
+
+      public ParameterValue OriginalQuantityValueFor(ObjectPath objectPath)
+      {
```
msevestre, line 150: what about you inline some the calll in this  file as we have a bunch of => already so that it's consistent

src/MoBi.Core/Serialization/Xml/Serializer/MoBiSimulationXmlSerializer.cs
```diff
@@ -23,6 +23,7 @@ public override void PerformMapping()
          Map(x => x.ParameterIdentificationWorkingDirectory);
          Map(x => x.HasUpToDateResults);
          Map(x => x.OutputMappings);
+         MapEnumerable(x => x.OriginalQuantityValues, simulation => simulation.AddOriginalQuantityValue);
```
msevestre, line 26: why x and then simulation? Let's use x :)

src/MoBi.Core/Services/QuantityToParameterValueMapper.cs
```diff
@@ -0,0 +1,36 @@
...
+   public class QuantityToParameterValueMapper : IQuantityToParameterValueMapper
+   {
+      private readonly IEntityPathResolver _entityPathResolver;
+
+      public QuantityToParameterValueMapper(IEntityPathResolver entityPathResolver)
+      {
+         _entityPathResolver = entityPathResolver;
+      }
+
+      public ParameterValue MapFrom(IQuantity input)
```
msevestre, line 21: rename input

src/MoBi.Core/Services/QuantityValueInSimulationChangeTracker.cs
```diff
@@ -0,0 +1,71 @@
...
+                newParameterValue.ValueOrigin.Equals(oldParameterValue.ValueOrigin);
+      }
+
+      /// <summary>
+      /// If the result of the tracked change is that the new quantity becomes the same as the original quantity,
+      /// then we remove the original quantity value from the simulation.
+      /// </summary>
+      private void checkForOriginalValueRestored(IQuantity quantity, IMoBiSimulation simulation)
+      {
+         if (quantity == null || simulation == null)
```
msevestre, line 57: how is this even possible on an internal method?

src/MoBi.Core/Services/QuantityValueInSimulationChangeTracker.cs
```diff
@@ -0,0 +1,71 @@
...
+
+         changeAction();
+
+         checkForOriginalValueRestored(quantity, simulation);
+      }
+
+      private void storeOriginalQuantityValue(IQuantity quantity, IMoBiSimulation simulation)
+      {
+         simulation.AddOriginalQuantityValue(_quantityToParameterValueMapper.MapFrom(quantity));
+         _eventPublisher.PublishEvent(new SimulationStatusChangedEvent(simulation));
```
msevestre, line 38: this will be fired even if there are no changes . is this right?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
msevestre: Ok I see. You only want to add it once. This needs to be explained here that we want to track the very first entry.

src/MoBi.Core/Commands/SetQuantityPropertyInSimulationCommandBase.cs
msevestre: Ok I understand what this does after reading the code.
Can you create a local variable maybe for the context.Resolve. 
e.g. changeTracker = xxx
changeTracker.TrackChanges()

src/MoBi.Core/Services/QuantityValueInSimulationChangeTracker.cs
```diff
@@ -0,0 +1,71 @@
...
+         storeOriginalQuantityValue(quantity, simulation);
+
+         changeAction();
+
+         checkForOriginalValueRestored(quantity, simulation);
+      }
+
+      private void storeOriginalQuantityValue(IQuantity quantity, IMoBiSimulation simulation)
+      {
+         simulation.AddOriginalQuantityValue(_quantityToParameterValueMapper.MapFrom(quantity));
```
msevestre, line 37: shoy;d thid method take a PV instead of the quantity?
That way you don't have to create a new PV twice every single time (in the store and in the check)

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
rwmcintosh: IReadOnlyCollection

src/MoBi.Core/Services/QuantityValueInSimulationChangeTracker.cs
rwmcintosh: yes, that's true. We could count the number of PV's before and after if you think it's worth while to cut down

src/MoBi.Core/Services/QuantityValueInSimulationChangeTracker.cs
```diff
@@ -0,0 +1,71 @@
...
+         _eventPublisher.PublishEvent(new SimulationStatusChangedEvent(simulation));
+      }
+
+      private bool areEquivalent(ParameterValue newParameterValue, ParameterValue oldParameterValue)
+      {
+         return newParameterValue.Value == oldParameterValue.Value &&
```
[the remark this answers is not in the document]
rwmcintosh: The tracker is only setting and checking some of the fields, so I don't know... I'd probably have to expand the number of fields that are being set and checked to move this outside the context of this tracker don't you think?

msevestre (approved):
looks good to me once we add a small change to the tracker to make it clear that the action is modifying (at least in theory) the quantity

msevestre (approved):

PR #1069, opened 2023-08-16
1060 update/commit commit changes from simulation to a building block

rwmcintosh:


src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+         _templateResolverTask = templateResolverTask;
+         _entitiesInSimulationRetriever = entitiesInSimulationRetriever;
+         _initialConditionsCreator = initialConditionsCreator;
+         _parameterValuesCreator = parameterValuesCreator;
+         _nameCorrector = nameCorrector;
+      }
+
+      public IMoBiCommand CommitSimulationChanges(IMoBiSimulation simulationWithChanges)
+      {
+         var lastModuleConfiguration = simulationWithChanges.Configuration.ModuleConfigurations.Last();
```
msevestre, line 47: Can we add a comment in the code explaining why this is last? This is a decision we made but this will be forgotteh

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+         var templateModule = _templateResolverTask.TemplateModuleFor(lastModuleConfiguration.Module);
+         
+         var moBiMacroCommand = new MoBiMacroCommand
```
msevestre, line 50: let's call it a macroCommand. Mobi in the name feels weird to me

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+         {
+            CommandType = CommitCommand,
+            Description = CommitCommandDescription(templateModule, simulationWithChanges),
+            ObjectType = new ObjectTypeResolver().TypeFor<Module>()
```
msevestre, line 54: why are you instantiating the objec tnew instead of having it as a dependency?

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+         moBiMacroCommand.Add(lastModuleConfiguration.SelectedParameterValues == null ? 
+            addNewParameterValuesFromSimulationChanges(simulationWithChanges, templateModule) : 
+            updateParameterValuesFromSimulationChanges(simulationWithChanges, lastModuleConfiguration.SelectedParameterValues));
+
+         return moBiMacroCommand.Run(_context);
+      }
+
+      private ICommand updateParameterValuesFromSimulationChanges(IMoBiSimulation simulationWithChanges, ParameterValuesBuildingBlock selectedParameterValues)
+      {
+         return new MoBiEmptyCommand();
```
msevestre, line 69: Throw a not implemented yet exception instead ?

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+      }
+
+      private ICommand updateInitialConditionsFromSimulationChanges(IMoBiSimulation simulationWithChanges, InitialConditionsBuildingBlock selectedInitialConditions)
+      {
+         return new MoBiEmptyCommand();
```
msevestre, line 74: Throw a not implemented yet exception instead ?

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+      }
+
+      /// <summary>
+      ///    Returns path and quantity tuple for all changes in the <paramref name="simulationWithChanges" /> of type
+      ///    <typeparamref name="TQuantity" />
+      /// </summary>
+      private IEnumerable<(ObjectPath objectPath, TQuantity quantity)> changesFrom<TQuantity>(IMoBiSimulation simulationWithChanges) where TQuantity : Quantity
+      {
+         var quantities = _entitiesInSimulationRetriever.EntitiesFrom<TQuantity>(simulationWithChanges);
+         return simulationWithChanges.OriginalQuantityValues.Where(x => quantities.Contains(x.Path)).Select(x => (objectPath: x.Path, quantity: quantities[x.Path]));
```
msevestre, line 96: hmm.. I like the map and reduce pattern. I am not sure what bests but I am surprised when I read this to see contains
I would have expected
simulationWithChanges.OriginalQuantityValues.Select(x=>{path: x.Path, quantities: quantities[x.Path]}). Where(x=>x.quantities!=null)

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+      {
+         var initialConditionsToAdd = changesFrom<MoleculeAmount>(simulationWithChanges).Select(x => createInitialCondition(x.quantity, x.objectPath)).ToList();
+         return createAddBuildingBlockCommand<InitialConditionsBuildingBlock, InitialCondition>(simulationWithChanges.Name, lastModule.InitialConditionsCollection.AllNames(), lastModule, initialConditionsToAdd);
+      }
+
+      /// <summary>
+      ///    Returns path and quantity tuple for all changes in the <paramref name="simulationWithChanges" /> of type
+      ///    <typeparamref name="TQuantity" />
+      /// </summary>
+      private IEnumerable<(ObjectPath objectPath, TQuantity quantity)> changesFrom<TQuantity>(IMoBiSimulation simulationWithChanges) where TQuantity : Quantity
```
msevestre, line 93: rename objectPath => quantityPath. Otherwise it's not clear that the belong together

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+      }
+
+      private IMoBiCommand createAddBuildingBlockCommand<TBuildingBlock, TPathAndValueEntity>(string buildingBlockName, IEnumerable<string> usedNames, Module moduleToAddTo, IReadOnlyList<TPathAndValueEntity> entitiesToAdd)
+         where TBuildingBlock : PathAndValueEntityBuildingBlock<TPathAndValueEntity>, IBuildingBlock where TPathAndValueEntity : PathAndValueEntity
+      {
+         if (!entitiesToAdd.Any())
+            return new MoBiEmptyCommand();
+
+         var buildingBlock = _context.Create<TBuildingBlock>().WithName(buildingBlockName);
+         entitiesToAdd.Each(buildingBlock.Add);
```
msevestre, line 106: Can't you do AddChildren or sthg like this?

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+      ///    Returns path and quantity tuple for all changes in the <paramref name="simulationWithChanges" /> of type
+      ///    <typeparamref name="TQuantity" />
+      /// </summary>
+      private IEnumerable<(ObjectPath objectPath, TQuantity quantity)> changesFrom<TQuantity>(IMoBiSimulation simulationWithChanges) where TQuantity : Quantity
+      {
+         var quantities = _entitiesInSimulationRetriever.EntitiesFrom<TQuantity>(simulationWithChanges);
+         return simulationWithChanges.OriginalQuantityValues.Where(x => quantities.Contains(x.Path)).Select(x => (objectPath: x.Path, quantity: quantities[x.Path]));
+      }
+
+      private IMoBiCommand createAddBuildingBlockCommand<TBuildingBlock, TPathAndValueEntity>(string buildingBlockName, IEnumerable<string> usedNames, Module moduleToAddTo, IReadOnlyList<TPathAndValueEntity> entitiesToAdd)
```
msevestre, line 99: can we retrieve  IEnumerable<string> usedNames from the module 
e.g. modules.All<TBuildingBlock>.AllName(). Just trying to reduce the number of parameters if possible

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+         if (!entitiesToAdd.Any())
+            return new MoBiEmptyCommand();
+
+         var buildingBlock = _context.Create<TBuildingBlock>().WithName(buildingBlockName);
+         entitiesToAdd.Each(buildingBlock.Add);
+         _nameCorrector.AutoCorrectName(usedNames, buildingBlock);
+         return new AddBuildingBlockToModuleCommand<TBuildingBlock>(buildingBlock, moduleToAddTo);
+      }
+
+      private InitialCondition createInitialCondition(MoleculeAmount moleculeAmount, ObjectPath path)
```
msevestre, line 111: hard to say what paht this is here.

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -0,0 +1,126 @@
...
+      {
+         var newPath = path.Clone<ObjectPath>();
+         // Remove the name from the path before creating the initial condition with separate arguments for name and path
+         newPath.RemoveAt(newPath.Count - 1);
+         var initialCondition = _initialConditionsCreator.CreateInitialCondition(newPath, path.Last(), moleculeAmount.Dimension, moleculeAmount.DisplayUnit, moleculeAmount.ValueOrigin);
```
msevestre, line 116: feels like the interface here is missing a property lol. CreateIniticalCOndiition(path, moleculeAmouhnt) like we have for parameter. This is not symmetrical and it's ugly :)

msevestre (changes requested):
Looks very good and easy to read. I just made a few basic suggestions
I am not sure if we need to propaget the name simulationWichChanges all the time. In the context of the COmmitSimulationChanges, passing simulation to private methods could be simplified. I like long names but it's sometimes a bit hard to read. On the other hand, parameter path is not enough info for me,

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
rwmcintosh: That's going to be implemented tomorrow

rwmcintosh: and if you throw an exception, you can't test anything.

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
rwmcintosh: I'll make a comment at the public interface level describing how commits work

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
rwmcintosh: I can make an extension method that just does this. There's nothing already existing.

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
rwmcintosh: If you have all building blocks except PV's and they are all named like the simulation, then you wouldn't be able to create the PV building block with the same name as all the other building blocks.

I think the expected behavior of adding the first PV or IC from a commit would be to have the same name as the simulation, which is allowed.

rwmcintosh: ```module.BuildingBlocks.OfType<TBuildingBlock>().AllNames();```
That should work

PR #1070, opened 2023-08-18
Fixes #1065 Commit to a module that has an existing IC/PV

### LINKED ISSUE #1065: Update/Commit - Commit to a module that has an existing IC/PV
See #1050 
This is for when you commit, and there's an IC/PV in the module already. You need to either add or update values

rwmcintosh:


src/MoBi.Core/Services/QuantityTask.cs
```diff
@@ -213,22 +211,17 @@ private ICommand setParameterDefaultStateInSimulation(IParameter parameter, bool
...
 
-         //add one before setting the value in the simulation to enable correct undo
-         macroCommand.Add(_quantitySynchronizer.Synchronize(quantity, simulation));
-
          IOSPSuiteCommand executedCommand = simulationCommandToBeRun.AsHidden().Run(_context);
 
          if (shouldUpdateValueOriginAndState)
             executedCommand = withUpdatedDefaultStateAndValueOrigin(executedCommand, quantity, simulation);
 
          macroCommand.Add(executedCommand);
```
msevestre, line 223: I don't really get why we have a macro command now. It looks like we have one command that we keep updating but I am not sure. For me, the macro command is just complexity for nothing

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForSimulation.cs
```diff
@@ -41,11 +46,16 @@ public IContextMenu InitializeWith(ObjectBaseDTO dto, IPresenter presenter)
...
             createRenameItem(simulation),
+            createConfigure(simulation).AsGroupStarter()
+         };
 
-            createConfigure(simulation).AsGroupStarter(),
-            createUpdate(simulation),
-            createCommit(simulation),
+         if (simulation.BuildingBlocks().Any(hasChanges))
+            _allMenuItems.Add(createUpdate(simulation));
+         if (simulation.OriginalQuantityValues.Any())
```
msevestre, line 54: run resharper in this file

msevestre (commented):
Looks good to me. Just this one question regarding usage of MacroComamnd required before but probably not now

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForSimulation.cs
rwmcintosh: It did not make any changes for these if statements FYI. I just added a line.

src/MoBi.Core/Services/QuantityTask.cs
rwmcintosh: removed it.

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForSimulation.cs
msevestre: Sure. Still run resharoer because the file is now off lol.

msevestre (approved):

PR #1080, opened 2023-08-24
Fixes #1066 update the simulationconfiguration after commit

### LINKED ISSUE #1066: Update/Commit - Update the SimulationConfiguration after commit
Once you have committed values back to a module, you need to update the simulationConfiguration of the simulation that was committed because the building blocks contained there will now be out of date.

rwmcintosh:
I'm actually doing this during the commit.

In the process of creating PV's and IC's for the template building block, the commit task also performs the same tasks on the simulation building block.


src/MoBi.Assets/AppConstants.cs
```diff
@@ -29,6 +29,7 @@ public static class AppConstants
       public static readonly string ParameterType = "Type";
       public static readonly string NaN = "<NaN>";
       public static readonly string Project = "Project";
+      public static readonly string Simulation = "Simulation";
```
msevestre, line 32: We have to have this constant somewhere lol

src/MoBi.Assets/AppConstants.cs
```diff
@@ -495,9 +496,14 @@ public static string RemoveValuePointDescription(double xDisplayValue, double yD
...
-         public static string CommitCommandDescription(Module module, IModelCoreSimulation simulation)
+         public static string CommitCommandDescription(IModelCoreSimulation simulation)
          {
-            return $"Commit changes made in Simulation: '{simulation.Name}' to module: '{module.Name}'";
+            return $"Commit changes made in Simulation: '{simulation.Name}'";
+         }
+
+         public static string CommitToModuleCommandDescription(Module module, IModelCoreSimulation simulation, string moduleParent)
+         {
+            return $"Commit changes made in Simulation: '{simulation.Name}' to {moduleParent} module: '{module.Name}'";
```
msevestre, line 506: I don't understand the parameter module parent. What does this mean?

src/MoBi.Core/Commands/RemoveSelectedBuildingBlockFromLastModuleConfigurationCommand.cs
```diff
@@ -0,0 +1,56 @@
...
+         _simulationId = simulation.Id;
+      }
+
+      protected override void DoExecute(IMoBiContext context)
+      {
+         base.DoExecute(context);
+         switch (_buildingBlock)
+         {
+            case InitialConditionsBuildingBlock _:
+               _simulation.Configuration.ModuleConfigurations.Last().SelectedInitialConditions = null;
```
msevestre, line 26: Create a local variable for the last module to make the code easier to read?

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
```diff
@@ -57,7 +57,8 @@ public SimulationConfiguration CreateFromProjectTemplatesBasedOn(SimulationConfi
 
          configuration.ModuleConfigurations.Each(moduleConfiguration => { simulationConfiguration.AddModuleConfiguration(templateModuleConfigurationFor(moduleConfiguration)); });
 
-         simulationConfiguration.Individual = templateBuildingBlockFor(configuration.Individual);
+         if (configuration.Individual != null)
+            simulationConfiguration.Individual = templateBuildingBlockFor(configuration.Individual);
```
msevestre, line 61: Why are you saving the template and not the clone?

msevestre (commented):
I need to review more .

src/MoBi.Assets/AppConstants.cs
rwmcintosh: moduleParent is going to be either "Simulation" or "Project" to highlight where the module lives.

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
rwmcintosh: That's so the linking all works when you create the model.

This method creates the configuration with the template bb's. Then the model, then they are cloned

rwmcintosh: Oh, that's a bug I found along the way BTW. this crashed if you had a simulation without an individual

src/MoBi.Assets/AppConstants.cs
```diff
@@ -495,9 +496,14 @@ public static string RemoveValuePointDescription(double xDisplayValue, double yD
             return string.Format("Remove {3} X: '{0} {5}' Y: '{1} {6}' from {4} '{2}'", xDisplayValue, yDisplayValue, formulaName, ObjectTypes.ValuePoint, ObjectTypes.TableFormula, xDisplayUnit.Name, yDisplayUnit.Name);
          }
 
-         public static string CommitCommandDescription(Module module, IModelCoreSimulation simulation)
+         public static string CommitCommandDescription(IModelCoreSimulation simulation)
          {
-            return $"Commit changes made in Simulation: '{simulation.Name}' to module: '{module.Name}'";
+            return $"Commit changes made in Simulation: '{simulation.Name}'";
```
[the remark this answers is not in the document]
rwmcintosh: Here's an example where I expand the top level. Maybe it makes sense to put the module name anyway at the top one because that's the primary task of the commit.

Or do you think some other way of describing it?

image

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
msevestre: Ok I was wondering why this change was related to commit. This got ne confused

msevestre (approved):
I think the code is working but the nested macro command structure is just over the top for me.  Some of the helper methods are just in the way of understanding .I don't need to see a method called createEmptyCommitMacroCommand that just does nothing tieh coming but sets command and description. Ultimately, I would not create so many macro command. All those actions are atomic. They belong in one that can be reversed and that's all. I also believe that the subcommand should be hidden because no one will care in the history (at least the one coming from the simulation)

rwmcintosh:
> I think the code is working but the nested macro command structure is just over the top for me. Some of the helper methods are just in the way of understanding .I don't need to see a method called createEmptyCommitMacroCommand that just does nothing tieh coming but sets command and description. Ultimately, I would not create so many macro command. All those actions are atomic. They belong in one that can be reversed and that's all. I also believe that the subcommand should be hidden because no one will care in the history (at least the one coming from the simulation)

Relatively easy to change that. Only one macro command is created now and all the subcommand are added. Commands that used are used to manipulate the simulation building blocks are hidden.

PR #1089, opened 2023-08-29
1067 simulation status when a module is updated

rwmcintosh:
If you add a building block to a module simulations using that module should be marked as out of sync.

Special attention for initial conditions and parameter values, because the module can have multiple instances, but only one can be in use in a simulation so adding those does not change the sync status.

Remove is similar, except it is also not possible to remove a building block from a module if the building block is in use. Again, only IC's and PV's can be removed from a module that's in use, and only if they are not selected.

### DIFF
--- a/src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
+++ b/src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
@@ -29,12 +29,7 @@ namespace MoBi.Core.Commands
       {
          var projectRetriever = context.Resolve<IMoBiProjectRetriever>();
          var affectedSimulations = projectRetriever.Current.SimulationsUsing(changedModule);
-         affectedSimulations.Each(x => refreshSimulation(x, context));
-      }
-
-      private void refreshSimulation(IMoBiSimulation simulation, IMoBiContext context)
-      {
-         context.PublishEvent(new SimulationStatusChangedEvent(simulation));
+         affectedSimulations.Each(x => context.PublishEvent(new SimulationStatusChangedEvent(x)));
       }
 
       protected override void ClearReferences()
--- a/src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
+++ b/src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
@@ -1,6 +1,5 @@
 using MoBi.Core.Domain.Model;
 using MoBi.Core.Events;
-using MoBi.Core.Services;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Utility.Extensions;
@@ -27,8 +26,7 @@ namespace MoBi.Core.Commands
 
       protected void PublishSimulationStatusChangedEvents(Module changedModule, IMoBiContext context)
       {
-         var projectRetriever = context.Resolve<IMoBiProjectRetriever>();
-         var affectedSimulations = projectRetriever.Current.SimulationsUsing(changedModule);
+         var affectedSimulations = context.CurrentProject.SimulationsUsing(changedModule);
          affectedSimulations.Each(x => context.PublishEvent(new SimulationStatusChangedEvent(x)));
       }
 
--- a/src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
+++ b/src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
@@ -1,0 +1,46 @@
+using MoBi.Core.Domain.Model;
+using MoBi.Core.Events;
+using MoBi.Core.Services;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Utility.Extensions;
+
+namespace MoBi.Core.Commands
+{
+   public abstract class ModuleContentChangedCommand<T> : MoBiReversibleCommand where T : class, IBuildingBlock
+   {
+      protected Module _existingModule;
+      private readonly string _existingModuleId;
+      protected T _buildingBlock;
+
+      protected ModuleContentChangedCommand(T buildingBlock, Module existingModule)
+      {
+         _existingModule = existingModule;
+         _existingModuleId = existingModule.Id;
+         _buildingBlock = buildingBlock;
+      }
+
+      public override void RestoreExecutionData(IMoBiContext context)
+      {
+         _existingModule = context.Get<Module>(_existingModuleId);
+      }
+
+      protected void PublishSimulationStatusChangedEvents(Module changedModule, IMoBiContext context)
+      {
+         var projectRetriever = context.Resolve<IMoBiProjectRetriever>();
+         var affectedSimulations = projectRetriever.Current.SimulationsUsing(changedModule);
+         affectedSimulations.Each(x => refreshSimulation(x, context));
+      }
+
+      private void refreshSimulation(IMoBiSimulation simulation, IMoBiContext context)
+      {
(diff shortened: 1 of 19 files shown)


src/MoBi.Presentation/Presenter/Main/SimulationExplorerPresenter.cs
```diff
@@ -188,21 +188,51 @@ private void refreshDisplayedSimulation(IMoBiSimulation simulation)
...
          
          var changedTemplateBuildingBlocks = findChangedTemplateBuildingBlocks(simulation).ToList();
+         var changedModules = findChangedTemplateModules(simulation).ToList();
 
          var isChangedSimulation = changedTemplateBuildingBlocks.Any() || simulation.OriginalQuantityValues.Any();
          simulationNode.Icon = isChangedSimulation ? ApplicationIcons.SimulationRed : ApplicationIcons.SimulationGreen;
 
+         updateModuleIcons(changedModules.Select(x => x.simulationModule), simulationNode.AllNodes.OfType<ModuleConfigurationNode>());
          updateBuildingBlockIcons(changedTemplateBuildingBlocks.Select(x => x.simulationBuildingBlock), simulationNode.AllNodes.OfType<BuildingBlockNode>());
       }
```
msevestre, line 199: Would it make sense to move all of this stuff outside of the presenter? WE have all those tasks already. Does it belong in InteractionTaskForSimulation . Most of those methods are private . It's rare to have so much logic in a presenter

src/MoBi.Core/Commands/ModuleContentChangedCommand.cs
```diff
@@ -0,0 +1,41 @@
...
+      }
+
+      public override void RestoreExecutionData(IMoBiContext context)
+      {
+         _existingModule = context.Get<Module>(_existingModuleId);
+      }
+
+      protected void PublishSimulationStatusChangedEvents(Module changedModule, IMoBiContext context)
+      {
+         var projectRetriever = context.Resolve<IMoBiProjectRetriever>();
```
msevestre, line 30: retrieve project from context

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForSimulation.cs
```diff
@@ -31,7 +32,9 @@ public interface IInteractionTasksForSimulation : IInteractionTasksForChildren<M
       /// <summary>
       ///    Returns the project building block that matches <paramref name="buildingBlock" />
       /// </summary>
-      IBuildingBlock TemplateBuildingBlockFor(IBuildingBlock buildingBlock);
+      TBuildingBlock TemplateBuildingBlockFor<TBuildingBlock>(TBuildingBlock buildingBlock) where TBuildingBlock : class, IBuildingBlock;
+
+      Module TemplateModuleFor(Module module);
```
msevestre, line 37: See here, we added some code to the interaction task for simulation (which is a task realted to simulation stuff). This is where I  would consider moving some of the code from above (which also called the InteractionTask for Simulation). Not sure if this makes sense

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForSimulation.cs
```diff
@@ -94,6 +97,11 @@ public IBuildingBlock TemplateBuildingBlockFor(IBuildingBlock buildingBlock)
          return _templateResolverTask.TemplateBuildingBlockFor(buildingBlock);
       }
 
+      public Module TemplateModuleFor(Module module)
```
msevestre, line 100: let's try to inline those now. I know it's a matter of preference ultimately but I'd like to simplify those simple one liner

msevestre (changes requested):

src/MoBi.Presentation/Presenter/Main/SimulationExplorerPresenter.cs
```diff
@@ -188,21 +188,51 @@ private void refreshDisplayedSimulation(IMoBiSimulation simulation)
...
+      {
+         return simulation.Configuration.ModuleConfigurations.Select(x => (templateModule: _interactionTasksForSimulation.TemplateModuleFor(x.Module), moduleConfiguration: x)).
+            Where(match => !versionMatch(match)).Select(x => (templateModule: x.templateModule, simulationModule: x.moduleConfiguration.Module));
+      }
+
+      private bool versionMatch((Module templateModule, ModuleConfiguration moduleConfiguration) match)
+      {
+         var moduleConfiguration = match.moduleConfiguration;
+         var simulationModule = moduleConfiguration.Module;
+         var templateInitialConditions = moduleConfiguration.SelectedInitialConditions == null ? null : _interactionTasksForSimulation.TemplateBuildingBlockFor(moduleConfiguration.SelectedInitialConditions);
```
msevestre, line 210: if TemplateBuildingBlockFor returns null if the passed parameter is null, then you don't nee dall those condition all the time. Woudl iut make sense?

src/MoBi.Presentation/Presenter/Main/SimulationExplorerPresenter.cs
```diff
@@ -188,21 +188,51 @@ private void refreshDisplayedSimulation(IMoBiSimulation simulation)
...
          updateBuildingBlockIcons(changedTemplateBuildingBlocks.Select(x => x.simulationBuildingBlock), simulationNode.AllNodes.OfType<BuildingBlockNode>());
       }
 
+      private IEnumerable<(Module templateModule, Module simulationModule)> findChangedTemplateModules(IMoBiSimulation simulation)
+      {
+         return simulation.Configuration.ModuleConfigurations.Select(x => (templateModule: _interactionTasksForSimulation.TemplateModuleFor(x.Module), moduleConfiguration: x)).
+            Where(match => !versionMatch(match)).Select(x => (templateModule: x.templateModule, simulationModule: x.moduleConfiguration.Module));
+      }
+
+      private bool versionMatch((Module templateModule, ModuleConfiguration moduleConfiguration) match)
```
msevestre, line 206: I would create two parameters instead of one here because we have match.xx everywhere. Or you deconstruct right away on first line. I am not sure there is a gain here for the one call that we have on line 203

src/MoBi.Presentation/Presenter/Main/SimulationExplorerPresenter.cs
```diff
@@ -188,21 +188,51 @@ private void refreshDisplayedSimulation(IMoBiSimulation simulation)
...
+
       private void updateBuildingBlockIcons(IEnumerable<IBuildingBlock> changedSimulationBuildingBlocks, IEnumerable<BuildingBlockNode> simulationNodeChildren)
       {
          simulationNodeChildren.Each(x => updateBuildingBlockIcon(x, changedSimulationBuildingBlocks));
       }
 
-      private static void updateBuildingBlockIcon(BuildingBlockNode x, IEnumerable<IBuildingBlock> changedSimulationBuildingBlocks)
+      private void updateModuleIcon(ModuleConfigurationNode moduleConfigurationNode, IEnumerable<Module> changedModules)
+      {
+         moduleConfigurationNode.Icon = changedModules.AllNames().Contains(moduleConfigurationNode.Text) ? ApplicationIcons.RedOverlayFor(moduleConfigurationNode.BaseIcon) : ApplicationIcons.GreenOverlayFor(moduleConfigurationNode.BaseIcon);
```
msevestre, line 230: ` changedModules.AllNames().Contains(moduleConfigurationNode.Text) ` looks weird to me
I a do not sure what were are using the text of the node here. This feels dangeraous

msevestre (changes requested):

PR #1092, opened 2023-09-06
Fixes #1090 SynchronizeParameterValueCommand does not undo the changes

### LINKED ISSUE #1090: SynchronizeParameterValueCommand does not undo the changes
I thought this command would work differently. When reverted, it does not undo its updates unless it is run after a change in the quantity it's based on.

rwmcintosh:
Non reversible command added at the end.

I verified that if I put this command in first, the whole macro is non-reversible. It doesn't matter.


msevestre:
looks good

PR #1093, opened 2023-09-06
Fixes #1088 IC and PV BB cannot have the same name.

### LINKED ISSUE #1088: IC and PV BB cannot have the same name.
1. New MoBi project
2. Create a new project with PV and IC
3. Clone the available IC BB and name it "IC_customName"
4. Try to clone the PV BB and name it "IC_customName":

image

However, it should be possible to have IC and PV BBs with the same name

rwmcintosh:


rwmcintosh:
I also removed some stuff that was not being used

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -101,6 +97,15 @@ public IMoBiCommand SetNegativeValuesAllowed(TBuildingBlock initialConditions, I
          return macroCommand;
       }
 
+      protected override bool CorrectName(TBuildingBlock buildingBlock, Module module)
+      {
+         if (buildingBlock is ExpressionProfileBuildingBlock)
```
msevestre, line 102: This needs a comment here. Why is there some special stuff done for NOT ExpressionProfle? I don't understand the code

msevestre (changes requested):

PR #1095, opened 2023-09-07
Fixes #1076 Expression profile: Initial conditions

### LINKED ISSUE #1076: Expression profile: Initial conditions
1) We cannot add or delete new entries in the "Parameters" tab, but can delete entries in the "Initial conditions" tab. I think the latter should also be not modifiable.

2) I would propose to hide the "Is Present" column
3) Also remove the "Value Origin" column
4) Rename column "Initial condition" to "Value"

rwmcintosh:
Also removed a bunch of code that's not used.


src/MoBi.Presentation/Presenter/ExpressionProfileInitialConditionsPresenter.cs
```diff
@@ -26,6 +26,11 @@ public class ExpressionProfileInitialConditionsPresenter : BuildingBlockWithInit
          IFormulaToValueFormulaDTOMapper formulaToValueFormulaDTOMapper,
          IDimensionFactory dimensionFactory) : base(view, startValueMapper, initialConditionsTask, msvCreator, context, deleteStartValuePresenter, formulaToValueFormulaDTOMapper, dimensionFactory, isPresentSelectionPresenter, negativeStartValuesAllowedSelectionPresenter)
       {
+         HideDeleteColumn();
+         HideIsPresentView();
+         HideDeleteView();
```
msevestre, line 31: I dont't quite get why you implement a mix and match of creating methods to encapsulate the view method or calling the view method directly. 
This is ok but surprising to be inconsistent withing the SAME PR

src/MoBi.Presentation/Presenter/PathAndValueBuildingBlockPresenter.cs
```diff
@@ -286,6 +266,11 @@ public void Handle(PathAndValueEntitiesBuildingBlockChangedEvent eventToHandle)
             HandleBuildingBlockEvent(eventToHandle);
       }
 
+      public void HideDeleteView()
```
msevestre, line 269: here method is created

msevestre (approved):

PR #1122, opened 2023-09-21
Fixes #1114 MoBi12: Missing value origins in Expression profiles and …

### LINKED ISSUE #1114: MoBi12: Missing value origins in Expression profiles and Individual BBs
There is no possibility to enter the value origin information here.
Value origin exported from PK-Sim is lost.

rwmcintosh:
…Individual BBs

### DIFF
--- a/src/MoBi.Presentation/Presenter/IPathAndValueBuildingBlockPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/IPathAndValueBuildingBlockPresenter.cs
@@ -6,14 +6,14 @@ using OSPSuite.Core.Domain.UnitSystem;
 
 namespace MoBi.Presentation.Presenter
 {
-   public interface IPathAndValueBuildingBlockPresenter<in TDTO> : IBreadCrumbsPresenter
+   public interface IPathAndValueBuildingBlockPresenter<in TParameterDTO> : IBreadCrumbsPresenter
    {
-      void SetParameterValue(TDTO expressionParameterDTO, double? newValue);
-      void SetUnit(TDTO expressionParameter, Unit unit);
-      void SetFormula(TDTO expressionParameterDTO, IFormula newValueFormula);
+      void SetParameterValue(TParameterDTO parameterDTO, double? newValue);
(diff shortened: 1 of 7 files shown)


src/MoBi.Presentation/Presenter/IPathAndValueBuildingBlockPresenter.cs
```diff
@@ -13,5 +14,6 @@ public interface IPathAndValueBuildingBlockPresenter<in TDTO> : IBreadCrumbsPres
       IEnumerable<ValueFormulaDTO> AllFormulas();
       void AddNewFormula(TDTO expressionParameterDTO);
       IEnumerable<IDimension> DimensionsSortedByName();
+      void SetValueOrigin(TDTO parameterDTO, ValueOrigin valueOrigin);
```
msevestre, line 17: I don't get any of the parameter namign here. TDTO? Why is line 15 about expressionParameter and like 17 about parameter. Should it be TParameterDTO maybe? and the like 15 renamed?

src/MoBi.Presentation/Presenter/PathWithValueBuildingBlockPresenter.cs
```diff
@@ -43,6 +44,11 @@ public IEnumerable<ValueFormulaDTO> AllFormulas()
          return allFormulas;
       }
 
+      public void SetValueOrigin(TBuilderDTO individualParameterDTO, ValueOrigin newValueOrigin)
```
msevestre, line 47: and here another name. WHAT?!?

msevestre (changes requested):
looks good. Just naming of parmaeters and generic variable is all over the place

src/MoBi.Presentation/Presenter/IPathAndValueBuildingBlockPresenter.cs
rwmcintosh: OMG.

msevestre (approved):

PR #1123, opened 2023-09-21
Closes #1121 parameter values and initial conditions folders need to be created when missing during tree update

### LINKED ISSUE #1121: Parameter Values and Initial Conditions folders need to be created when missing during tree update
In Module Explorer,
If a module has no PV building blocks, the empty folder is not shown in the tree.

If a user then adds those building blocks to a module without the folder, the bb is not shown in the correct location

rwmcintosh:


msevestre (approved):
looks good. Wondering if we can simplify

PR #1124, opened 2023-09-25
Fixes #1109 MoBi 12: application subcontainers not properly displayed

### LINKED ISSUE #1109: MoBi 12: application subcontainers not properly displayed in the Events-BB
E.g. for an application transport:

In MoBi 11 (correct representation)
grafik

In MoBi 12 (wrong)
grafik

The same happens in MoBi 12 with ALL application subcontainers. Somehow MoBi 12 interpretes all application subcontainers as a top application container (because the shown view belongs to the top application container)
grafik

rwmcintosh:


src/MoBi.Presentation/Presenter/EditEventGroupBuildingBlockPresenter.cs
```diff
@@ -130,9 +130,9 @@ private IEnumerable<T> allPresenterImplementing<T>()
          return AllSubPresenters.OfType<T>();
       }
 
-      protected override (bool canHandle, IContainer parentObject) SpecificCanHandle(IObjectBase selectedObject)
+      protected override (bool canHandle, IContainer containerObject) SpecificCanHandle(IObjectBase selectedObject)
       {
-         return (shouldShow(selectedObject), null);
+         return (shouldShow(selectedObject), selectedObject as IContainer);
```
rwmcintosh, line 135: That's the real change. In v11.2 the method was:

```
      protected override Tuple<bool, IObjectBase> SpecificCanHandle(IObjectBase selectedObject)
      {
         return new Tuple<bool, IObjectBase>(shouldShow(selectedObject), selectedObject);
      }
```
So rather than naming this parentObject, I changed it to containerObject. In this case, the selected object is a Container.

msevestre (approved):

PR #1126, opened 2023-09-26
Fixes #1112 MoBi 12: Hide some standard warnings?

### LINKED ISSUE #1112: MoBi 12: Hide some standard warnings?
When I create a simulation based on a **small molecules spatial structure** (which does not include **Endosome** compartments) - I get a warning for each organ and each protein, that the initial condition cannot be applied to the `<Organ>|Endosome` container.
The warning is correct; however it will be the case whenever I use expression profiles exported from PK-Sim in combination with small molecules spatial structure.
E.g. below for 5 Proteins I get 135 Warnings (27 Organs * 5).

Would be nice to hide those warnings per default.
grafik

rwmcintosh:


src/MoBi.Presentation/Presenter/Main/NotificationPresenter.cs
```diff
@@ -207,9 +191,24 @@ private void updateNotificationWith(IEnumerable<NotificationMessageDTO> notifica
...
+      private bool isUnresolvedEndosomeForInitialConditionMessage(NotificationMessageDTO message)
+      {
+         if (!(message.Object is InitialCondition initialCondition)) 
+            return false;
+
+         // Only applies to endosome containers. Don't ignore other unresolved containers
+         if (!string.Equals(initialCondition.ContainerPath.Last(), Endosome))
+            return false;
+
+         return message.NotificationMessage.Message.Equals(Validation.StartValueDefinedForContainerThatCannotBeResolved(message.ObjectName, initialCondition.ContainerPath));
```
rwmcintosh, line 209: Testing the text here to see if this is the kind of message we should ignore. I don't know if it makes sense to have this test. Presumably there could be some other warning message for this container that we would incorrectly ignore if we don't somehow test for something specific to this notification.

src/MoBi.Presentation/Presenter/Main/NotificationPresenter.cs
```diff
@@ -176,20 +169,11 @@ public void ExportToFile()
...
-      {
-         VisibleNotification = VisibleNotification ^ typeToToggle;
-      }
+      public void Toggle(NotificationType typeToToggle) => VisibleNotification ^= typeToToggle;
 
-      public void Handle(ShowValidationResultsEvent validationResultsEvent)
-      {
-         updateNotificationWith(validationResultsEvent.ValidationResult.Messages.Select(m => _notificationMessageMapper.MapFrom(m)));
-      }
+      public void Handle(ShowValidationResultsEvent validationResultsEvent) => updateNotificationWith(validationResultsEvent.ValidationResult.Messages.Select(m => _notificationMessageMapper.MapFrom(m)));
```
msevestre, line 176: I think this is where some { } are required , or at least breaking somewhere else. Right now the actual code is broken up in a weird place which is hard to read

src/MoBi.Presentation/Presenter/Main/NotificationPresenter.cs
```diff
@@ -207,9 +191,24 @@ private void updateNotificationWith(IEnumerable<NotificationMessageDTO> notifica
 
       private bool notificationIsVisible(NotificationMessageDTO message)
       {
+         if (!_userSettings.ShowUnresolvedEndosomeMessagesForInitialConditions && isUnresolvedEndosomeForInitialConditionMessage(message))
```
msevestre, line 194: Should we reorganize to be a bt more readable
if(isOsberverMessagxx)
return showPKSm

if(isEndosomieError)
return sxxx

return true

msevestre (approved):
looks good to me with one suggeston

msevestre (approved):

Issue #1153, opened 2023-10-04
Changing parent container path of a container should change neighborhoods paths


PavelBal:
Replace the full path to this container by the new full path,

PavelBal:
Also for references in the formulas.

PR #1156, opened 2023-10-04
WIP #1153 changing parent container path. Make sure code compiles again

msevestre:
@rwmcintosh If you plan to work on mobi and need a core update, branch of this branch for now as core is not compatible anymore with develop


src/MoBi.Core/Commands/SetParentPathCommand.cs
```diff
@@ -30,9 +32,46 @@ protected override void ExecuteWith(IMoBiContext context)
...
+         if (spatialStructure == null)
+            return;
+
+         var oldObjectPathString = createContainerPath(_oldParentPath);
+         var newObjectPathString = createContainerPath(_newParentPath);
+
+         bool referencesPath(ObjectPath objectPath) => objectPath != null && objectPath.PathAsString.StartsWith(oldObjectPathString);
+         
+         //we know that the object path exists in this context
+         void updatePaths(ObjectPath objectPath) => objectPath.ReplaceWith(objectPath.PathAsString.Replace(oldObjectPathString, newObjectPathString).ToPathArray());
```
msevestre, line 52: not sure if there is a better way of doing this. 

How to find a path in the hierarchy. ? it starts with the oldPath

How to update the path? I have to replace the old path with the new one. 
The constant PathAsString is probably not the most efficient. But one can go from A|B to A or to A|B|C so you can't really use indexes etc..

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForRootEventGroup.cs
```diff
@@ -44,20 +40,6 @@ public override IMoBiCommand GetAddCommand(EventGroupBuilder builder, EventGroup
          return new AddRootEventGroupBuilderCommand(buildingBlock, builder);
       }
 
-      public void CreateApplicationsEventGroup(EventGroupBuildingBlock eventGroupBuildingBlock)
```
msevestre, line 47: @Yuri05 Hopefully we don;t need this anymore lol

tests/MoBi.Tests/Presentation/EditContainerPresenterSpecs.cs
```diff
@@ -73,17 +74,59 @@ public void should_tell_parameter_presenter_to_select_parameter()
...
+         _parentPath = "NewPath";
+         _muscleInterstitialPath = new FormulaUsablePath("A", "B", "Muscle", "Interstitial");
+         _muscle = new Container {ParentPath = new ObjectPath("A", "B")}.WithName("Muscle");
+         _neighborhoodReferencingContainer = new NeighborhoodBuilder
+         {
+            FirstNeighborPath = _muscleInterstitialPath,
+            SecondNeighborPath = new ObjectPath("Organism", "Liver", "Interstitial"),
+         };
+
+         _anotherTopContainer = new Container().WithName("Another");
```
msevestre, line 99: @Yuri05 I am testing all possible options here.

msevestre (commented):
LEt me know your thoughts?

src/MoBi.Core/Commands/SetParentPathCommand.cs
```diff
@@ -30,9 +32,46 @@ protected override void ExecuteWith(IMoBiContext context)
...
 
+      private void updateObjectPathsInSpatialStructure()
+      {
+         //should probably never be null
+         var spatialStructure = _buildingBlock as SpatialStructure;
+         if (spatialStructure == null)
+            return;
+
+         var oldObjectPathString = createContainerPath(_oldParentPath);
+         var newObjectPathString = createContainerPath(_newParentPath);
```
Yuri05, line 47: would rename the variables to `oldParentContainerPathString `and `newParentContainerPathString `for clarity

Yuri05: check that the old parent path was not empty. If it was - nothing to do?

src/MoBi.Core/Commands/SetParentPathCommand.cs
Yuri05: I would do rather something like this:
```
void updatePaths(ObjectPath objectPath)
{
   var objectPathAsString = objectPath.PathAsString;
   if (!objectPathAsString.StartsWith(oldParentContainerPathString)
      return;
   
   objectPathAsString = newParentContainerPathString + objectPathAsString.Substring(oldParentContainerPathString.Length);
  objectPath.ReplaceWith(objectPathAsString.ToPathArray());
}
```

This checks, that the object paths **starts** with the old parent container path and replaces **only the first occurrence** of the old parent container path with the new one.
So if the old container path was A and the new container path is B:
_**A**|X|A|Y_ will be renamed to _**B**|X|A|Y_
_C|X|A|Y_ will be not modified

msevestre: This referencePath already checks this. We know it starts with the path

msevestre: Replace also only replaces the first occurrence I think but I can add a test to prove this. I don't we need to change the code

Yuri05: > Replace also only replaces the first occurrence I think

it replaces all occurrences. It's very unlikely situation that we would have a situation like this, but you know... "should never happen" 😉 

```
void updatePath(ObjectPath objectPath) => 
 objectPath.ReplaceWith(newContainerPathString + objectPath.PathAsString.Substring(oldContainerPathString.Length)).ToPathArray());
```

PR #1159, opened 2023-10-06
1151 enable editing of neighbors paths

msevestre:


msevestre:
@Yuri05 Much more work than expected.

src/MoBi.Presentation/Presenter/SelectNeighborPathPresenter.cs
```diff
@@ -43,11 +44,13 @@ private void onSelectedContainerPathChanged(IEntity entity, ObjectPath container
          if (container.Mode != ContainerMode.Physical) 
             return;
 
-         _selectedPathDTO.Path = containerObjectPath.PathAsString;
+         var parentPath = container.RootContainer.ParentPath?.Clone<ObjectPath>() ?? new ObjectPath();
```
msevestre, line 47: @Yuri05 I am changing thius so that if we select a container with a root container with a parent path, we add the parent path as well.

msevestre (commented):
@Yuri05 I'd like a review first. I need to add tests

src/MoBi.Presentation/Presenter/SelectNeighborPathPresenter.cs
```diff
@@ -43,11 +44,13 @@ private void onSelectedContainerPathChanged(IEntity entity, ObjectPath container
...
+         containerObjectPath.Each(parentPath.Add);
```
Yuri05, line 48: Questions:
1) don't we have a function for concatenation of 2 paths?
(OK just checked: seems we don't. Created an issue for this: https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2140 )
2) Is actually checking that the selected neighbor is a container and a _physical_ container not a task of the **caller**?

msevestre: >Is actually checking that the selected neighbor is a container and a physical container not a task of the caller?

I don't get this. It's a presenter showing a hierarchical view of the structure. The user can click on a NON physical container (muscle) just to expand a node. I don't see who the caller should be as we are here in the upper most use case

Or do you suggest to allow the user to select Muscle as a neighbour (non physical container), click ok, and then be told it' snot ok?

msevestre: 1-: Yes we need a concat method indeed. I've been meaning to create one for a long time

PR #1164, opened 2023-10-12
Fixes #1132 simulation settings should operate independently from the project

### LINKED ISSUE #1132: Simulation settings should operate independently from the project
Need to be able to 

- [x] add output intervals in a simulation when there are none defined (you can remove all intervals in R, but not in MoBi)
- [x] Do not replace them when configuring a simulation from context menu

rwmcintosh:
image

### DIFF
--- a/src/MoBi.Core/Services/SimulationConfigurationFactory.cs
+++ b/src/MoBi.Core/Services/SimulationConfigurationFactory.cs
@@ -8,9 +8,10 @@ namespace MoBi.Core.Services
    public interface ISimulationConfigurationFactory
    {
       /// <summary>
-      ///    Creates a new SimulationConfiguration with default SimulationSettings and calculation methods from the project
+      ///    Creates a new SimulationConfiguration with calculation methods from the project and
+      ///    either <paramref name="settings" /> if specified, or default simulation settings
       /// </summary>
-      SimulationConfiguration Create();
+      SimulationConfiguration Create(SimulationSettings settings = null);
 
       /// <summary>
       ///    Creates a new SimulationConfiguration from the current project modules and building blocks based on the given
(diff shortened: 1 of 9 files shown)


src/MoBi.Core/Services/SimulationConfigurationFactory.cs
```diff
@@ -37,11 +38,11 @@ public class SimulationConfigurationFactory : ISimulationConfigurationFactory
          _templateResolverTask = templateResolverTask;
       }
 
-      public SimulationConfiguration Create()
+      public SimulationConfiguration Create(SimulationSettings settings = null)
       {
          var simulationConfiguration = new SimulationConfiguration
          {
-            SimulationSettings = _cloneManager.Clone(_projectRetriever.Current.SimulationSettings)
+            SimulationSettings = settings ?? _cloneManager.Clone(_projectRetriever.Current.SimulationSettings)
```
msevestre, line 45: I am not sure this is right here. you are going to use the same references (but it comes from another simulation)
Won't you be corrupting it?

src/MoBi.Presentation/Presenter/CreateSimulationConfigurationPresenter.cs
```diff
@@ -50,17 +49,19 @@ public class CreateSimulationConfigurationPresenter : WizardPresenter<ICreateSim
...
          UpdateControls();
          _view.Display();
          if (_view.Canceled)
          {
             return null;
          }
 
-         var simulationConfiguration = _simulationConfigurationFactory.Create();
+         // *Configuring* a simulation with new modules and building blocks should not replace the simulation settings
+         var simulationConfiguration = _simulationConfigurationFactory.Create(settings: isNew ? null : moBiSimulation.Settings);
```
msevestre, line 63: here. you are passing the settings from another simulation. Feels wrong

src/MoBi.Core/Services/SimulationConfigurationFactory.cs
rwmcintosh: Are there references to be worried about? Or do you mean that if we edit the properties of one simulation, we will change another as well?

src/MoBi.Presentation/Presenter/CreateSimulationConfigurationPresenter.cs
rwmcintosh: Maybe this should be a clone of the settings then. The idea is to create a simulation configuration but use the settings, or a copy of the settings from an existing simulation.

In this specific use-case, they go back to the same simulation because configure creates a new simulationConfiguration with new building blocks, but obviously that's not enforced from within this method.

I think if a clone is created of the settings, then it should work even if they're used in a different simulation.

msevestre: Yes
BEcause otherwise, editing one simulation would change the other. I do not know what's the original simulation is but a clone would make it obvious

PR #1165, opened 2023-10-14
Fixes #1155: Renaming container should change paths of neighborhoods …

### LINKED ISSUE #1155: Renaming containers should change paths of neighborhoods
- The same action should be triggered by changing 'ParentPath'

msevestre:
…as well as name of neighborhood

### DIFF
--- a/src/MoBi.Core/Commands/SetParentPathCommand.cs
+++ b/src/MoBi.Core/Commands/SetParentPathCommand.cs
@@ -1,6 +1,7 @@
 ﻿using System.Linq;
 using MoBi.Assets;
 using MoBi.Core.Domain.Model;
+using MoBi.Core.Events;
 using OSPSuite.Assets;
 using OSPSuite.Core.Commands.Core;
 using OSPSuite.Core.Domain;
@@ -32,15 +33,14 @@ namespace MoBi.Core.Commands
       {
          base.ExecuteWith(context);
          _container.ParentPath = _newParentPath;
-         updateObjectPathsInSpatialStructure();
+         updateObjectPathsInSpatialStructure(context);
          Description = AppConstants.Commands.UpdateParentPath(_container.EntityPath(), (_newParentPath ?? new ObjectPath()).ToPathString());
       }
 
-      private void updateObjectPathsInSpatialStructure()
+      private void updateObjectPathsInSpatialStructure(IMoBiContext context)
       {
          //should probably never be null
-         var spatialStructure = _buildingBlock as SpatialStructure;
-         if (spatialStructure == null)
+         if (!(_buildingBlock is SpatialStructure spatialStructure))
             return;
 
          var oldContainerPathString = createContainerPath(_oldParentPath);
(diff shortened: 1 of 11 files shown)


src/MoBi.Core/Domain/StringChanges.cs
```diff
@@ -7,8 +7,14 @@ namespace MoBi.Core.Domain
 {
    public class StringChanges : IEnumerable<IStringChange>
    {
+      private readonly IBuildingBlock _buildingBlock;
       private readonly IList<IStringChange> _list = new List<IStringChange>();
 
+      public StringChanges(IBuildingBlock buildingBlock)
+      {
+         _buildingBlock = buildingBlock;
```
msevestre, line 15: Refactored this guy so that I could simplify how it was called

tests/MoBi.Tests/Core/CheckNameVisitorSpecs.cs
```diff
@@ -149,23 +150,23 @@ protected override void Context()
...
          _initialCondition2 = new InitialCondition();
-         _path = new ObjectPath(new[] { "A", _changedName, "B" });
+         _path = new ObjectPath(new[] {"A", _changedName, "B"});
          _initialCondition2.Path = _path;
          _initialConditionsBuildingBlock.Add(_initialCondition2);
       }
 
       protected override void Because()
       {
-         _changes = sut.GetPossibleChangesFrom(_changedObject, _newName, _module, _changedName);
```
msevestre, line 168: funny how changing the actual code to be used as in producttion shows wrong usage in the tests. The fact that it worked here is only because of how the tree is being traversed (via visitor)

tests/MoBi.Tests/Core/CheckNameVisitorSpecs.cs
```diff
@@ -505,21 +550,29 @@ public class When_renaming_an_application_builder : concern_for_CheckNameVisitor
...
+         };
          _applicationBuilder.AddTransport(appTransport);
+
+         _buildingBlock.Add(_applicationBuilder);
       }
 
       protected override void Because()
       {
-         _resultChanges = sut.GetPossibleChangesFrom(A.Fake<IObjectBase>(), "NewName", _applicationBuilder, _oldName);
+         _resultChanges = sut.GetPossibleChangesFrom(_applicationBuilder, "NewName", _buildingBlock, _oldName);
```
msevestre, line 575: I have no idea why this worked here. The argument were even reversed!! with the object being renamed not being passed in the right place lol

src/MoBi.Core/Commands/SetParentPathCommand.cs
```diff
@@ -56,18 +56,22 @@ void updatePaths(ObjectPath objectPath)
...
-            .Where(referencesPath);
+            .SelectMany(x => new[] {new {neighborhood = x, neighbor = x.FirstNeighborPath}, new {neighborhood = x, neighbor = x.SecondNeighborPath}})
+            .Where(x => referencesPath(x.neighbor));
+
 
-         allReferencingPathsInNeighborhood.Each(updatePaths);
+         allReferencingPathsInNeighborhood.Each(x =>
+         {
+            updatePaths(x.neighbor);
+            context.PublishEvent(new NeighborhoodChangedEvent(x.neighborhood));
```
msevestre, line 73: I refactored this to ensure that I am publishing the neighborhood changed event. This defacto allows for a refresh of the tree at the node level. Works well :)

src/MoBi.Presentation/Tasks/CheckNameVisitor.cs
```diff
@@ -120,48 +120,81 @@ public void Visit(IFormula formula)
...
-         return replaceIn.Replace(_oldName, _newName);
+         checkObjectBase(objectBase);
       }
 
-      private bool containsOldName(string stringToCheck)
+      public void Visit(NeighborhoodBuilder neighborhoodBuilder)
       {
-         return stringToCheck.Contains(_oldName);
+         checkObjectBase(neighborhoodBuilder);
+         //If not renaming a container or we are renaming a neighborhood builder, we exist
```
rwmcintosh, line 162: ```we exit``` instead of ```we exist```

src/MoBi.Presentation/Tasks/CheckNameVisitor.cs
```diff
@@ -120,48 +120,81 @@ public void Visit(IFormula formula)
...
+
+         //now the name of the container is used in the neighborhood name. Propose update
+         if (containsOldName(neighborhoodBuilder.Name))
+         {
+            var newNeighborhoodName = stringReplace(neighborhoodBuilder.Name);
+            _changes.Add(neighborhoodBuilder, renameCommand(neighborhoodBuilder, newNeighborhoodName), renameDescription(neighborhoodBuilder, neighborhoodBuilder.Name, newNeighborhoodName, _namePropertyName));
+         }
+
+         //check first and second neighbors
+         if (neighborhoodBuilder.FirstNeighborPath?.Contains(_oldName) ?? false)
```
rwmcintosh, line 174: this code (174-178) is nearly identical to 180-184

rwmcintosh (approved):

src/MoBi.Presentation/Tasks/CheckNameVisitor.cs
msevestre: It can but I don't think this makes sense. There are two many differences in what is set or what is checked.

rwmcintosh: makes sense

PR #1166, opened 2023-10-17
Fixes #1143 select individual when exporting container from spatial structure

### LINKED ISSUE #1143: Select individual when exporting container from Spatial Structure
The list of parameters has to be completed by the selected individual, and all parameters have to be exported.
Look through the individual for paths that match the container

Implement as an additional context menu entry and keep the existing 'Save as PKML...' when user wants to just export the container

rwmcintosh:
You open the context menu and select 'Export With Individual'.  Any suggestions for this context menu are appreciated
image

Can't close without specifying both individual and file path. The file path button opens the standard file browser.
image

Ok, saves the container PKML with additional parameters from the individual where they match the container path
image

### DIFF
--- a/src/MoBi.Presentation/Presenter/SelectFolderAndIndividualFromProjectPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/SelectFolderAndIndividualFromProjectPresenter.cs
@@ -13,32 +13,22 @@ namespace MoBi.Presentation.Presenter
    public interface ISelectFolderAndIndividualFromProjectPresenter : IDisposablePresenter
    {
       /// <summary>
-      /// Opens a dialog where the user will select an IndividualBuildingBlock from the project and a file path to export
+      ///    Opens a dialog where the user will select an IndividualBuildingBlock from the project and a file path to export
       /// </summary>
-      void GetPathAndIndividualForExport(IContainer container);
+      (string, IndividualBuildingBlock) GetPathAndIndividualForExport(IContainer container);
 
       /// <summary>
-      /// Returns a list of all IndividualBuildingBlocks in the project
+      ///    Returns a list of all IndividualBuildingBlocks in the project
       /// </summary>
       IReadOnlyList<IndividualBuildingBlock> AllIndividuals { get; }
 
       /// <summary>
-      /// Opens a dialog for the user to select file path
+      ///    Opens a dialog for the user to select file path
       /// </summary>
       /// <returns>The path if dialog is dismissed with ok, empty string if canceled</returns>
       string BrowseFilePath();
-
-      /// <summary>
-      /// After dismissing the dialog, this returns the selected IndividualBuildingBlock or null if the dialog was canceled
-      /// </summary>
-      IndividualBuildingBlock SelectedIndividual { get; }
-      
-      /// <summary>
-      /// After dismissing the dialog, this returns the selected file path for export, or an empty string if the dialog was canceled
-      /// </summary>
-      string SelectedFilePath { get; }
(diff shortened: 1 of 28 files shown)


src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
```diff
@@ -40,45 +63,98 @@ protected override IEnumerable<string> GetUnallowedNames(IContainer container, I
...
+      private void addIndividualParametersToContainer(IndividualBuildingBlock individual, IContainer container, ObjectPath containerPath)
+      {
+         
+         individual.Where(individualParameter => individualParameter.ContainerPath.Equals(containerPath)).Each(x => addIndividualParameterToContainer(x, container));
+      }
 
+      private void addIndividualParameterToContainer(IndividualParameter individualParameter, IContainer container)
+      {
+         var parameter = _individualParameterToParameterMapper.MapFrom(individualParameter);
+         if (container.ContainsName(parameter.Name))
```
rwmcintosh, line 143: I'm not 100% sure on this. The scenario is that the parameter is defined in both individual and container. For now, I'm replacing the parameter from the spatial structure with a new one from the individual.

Mostly, this did not happen, but I ran into a case when exporting the 'Organism' where 'MeanHeight' is declared in both.

rwmcintosh:
@PavelBal @georgeDaskalakis Do you have any suggestions for a concise context menu entry for this feature?

rwmcintosh:
Converting to draft. I'll see if it's possible to handle the renaming during import since https://github.com/Open-Systems-Pharmacology/MoBi/issues/1155 was implemented already

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
Yuri05: Looks good for me so far.

msevestre (changes requested):
Nothing wrong but I think this can be improved.
Resharper should be run however => request change

src/MoBi.UI/Views/SelectFolderAndIndividualFromProjectView.cs
```diff
@@ -0,0 +1,64 @@
...
+      }
+
+      public void AttachPresenter(ISelectFolderAndIndividualFromProjectPresenter presenter)
+      {
+         _presenter = presenter;
+      }
+
+      public void BindTo(IndividualAndFilePathDTO dto)
+      {
+         descriptionLabel.Text = ExportContainerDescription(dto.ContainerPath.PathAsString);
```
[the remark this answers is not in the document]
rwmcintosh: yes, it works.

src/MoBi.Presentation/Presenter/SelectFolderAndIndividualFromProjectPresenter.cs
```diff
@@ -0,0 +1,62 @@
...
+using OSPSuite.Presentation.Presenters;
+
+namespace MoBi.Presentation.Presenter
+{
+   public interface ISelectFolderAndIndividualFromProjectPresenter : IDisposablePresenter
+   {
+      /// <summary>
+      ///    Opens a dialog where the user will select an IndividualBuildingBlock from the project and a file path to export
+      /// </summary>
+      (string, IndividualBuildingBlock) GetPathAndIndividualForExport(IContainer container);
```
msevestre, line 18: I think you can name the return values also but I am not sure (string exportPath, xxx)

PR #1182, opened 2023-11-01
Fixes #1144 Extend Initial Conditions should allow selection of molecules

### LINKED ISSUE #1144: Extend Initial Conditions should allow selection of which molecules to extend
Allow selection of proteins too. Right now, they are not selected.

rwmcintosh:
Fixes #1144

# Description
Allows users to select which molecules they want to use for the initial conditions extend. Users can select from any of the molecules in the project modules. The molecules selected must have unique names

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

Default grouping and you can use the group checkbox to select/deselect all molecules from the group
image

If you select multiple molecules with the same name, both will be invalid in the grid. When you deselect one, the other will become valid
image

# Questions (if appropriate):


src/MoBi.Presentation/DTO/MoleculeSelectionDTO.cs
```diff
@@ -0,0 +1,54 @@
...
+
+      public MoleculeBuilder MoleculeBuilder { get; set; }
+
+      public bool Selected { get; set; }
+
+      public string Icon => MoleculeBuilder.Icon;
+
+      public void AddSelectedMoleculeRetriever(Func<IReadOnlyList<MoleculeSelectionDTO>> getSelectedMolecules)
+      {
+         _getSelectedMolecules = getSelectedMolecules;
```
rwmcintosh, line 32: @msevestre my question is here. I need to invalidate at the selected molecule level so that the grid shows messages on the appropriate rows.

So every MoleculeSelectionDTO gets this retriever that calculates the already selected molecules, which is used to check for other selections with the same name.

I can't pass that as a list, because it will be calculated at time of construction of the object instead of at time of validation.

Do you think there's some other way to invalidate this object if another object is selected that has the same name?

Issue #1206, opened 2023-11-15
Refresh/commit simulation settings from project defaults


rwmcintosh:

rwmcintosh:
The simulation settings are not a real building block now, they are stored as project level defaults that are applied when the simulation is created.

Users can change the simulation settings, but when they do, there is no way to revert to the project defaults as there was when the simulation settings were a building block.

So, that ability should be recovered, to restore or refresh simulation settings from project defaults, but also to commit simulation settings to the project defaults as well. That way, settings from one simulation could easily be used in other simulations by committing to the project default, then refreshing other simulations.

PavelBal:
@rwmcintosh I could not find how this is implemented, re-opening. If its already there, please show me.

PavelBal:
OK found it... but let's discuss.

PR #1242, opened 2023-12-18
Fixes #1163 add output selections to project simulation settings editor

### LINKED ISSUE #1163: Add output selections to project simulation settings editor
ATM, you can't add output selections to the project's default simulation settings. Create a way to add them by path.

Users can manually edit paths, or browse from paths present in the project simulations already.

rwmcintosh:
Fixes #1163 

# Description
Allow users to edit/add/remove selections from the default project settings. In v11 you added/removed/updated these settings by committing from a simulation. We'll add that in #1206. 

This allows users to directly edit the paths in their default simulation settings using a tree browser, or by hand editing the path.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

In the simulation settings / output selections tab this grid view has new buttons added to each row
image

- You can type in the path column to make changes
- If I want to remove, that is straightforward the row disappears.

- If I edit a row, or I use the +button in a specific row, the tree selection appears with all simulations from the project. For paths that can be resolved in any simulation, the focused element will be pre-selected.

image

- Non-quantity selections prevent user from clicking 'OK'
image

- Trying to duplicate a selection, either through adding or editing, a dialog informs the user and prevents the output being added or updated.
image

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1578,6 +1578,8 @@ namespace MoBi.Assets
          public static readonly string SelectSpatialStructureAndMolecules = "Select a spatial structure and molecules";
          public static readonly string SelectSpatialStructure = "Select a spatial structure";
          public static readonly string ExtendDescription = "<b><i>Initial conditions</i> will be created for molecules in all physical containers in the selected <i>spatial structure</i></b>";
+         public static readonly string AddDefaultCurveForNewSimulations = "Add default curve for new simulations";
+         public static readonly string ChangeDefaultCurveForNewSimulations = "Change default curve for new simulations";
          public static string ExportContainerDescription(string exportedContainerPath) => $"Select an individual and file path for container export. Parameters from the individual that match the path {exportedContainerPath} will be addd to the container before exporting.";
          public static readonly string SelectIndividualAndPathForContainerExport = "Select an individual and path for container export";
 
@@ -2029,6 +2031,11 @@ namespace MoBi.Assets
          public static string NameIsAlreadyUsedInThisContainer(string containerPath, string name)
          {
             return $"A value with the name '{name}' already exists for the container {containerPath}";
+         }
+
+         public static string ThePathIsAlreadySelectedAsAnOutput(ObjectPath objectPath)
+         {
+            return $"The path {objectPath} is already selected as an output.";
          }
       }
 
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -2033,10 +2033,7 @@ namespace MoBi.Assets
             return $"A value with the name '{name}' already exists for the container {containerPath}";
          }
 
-         public static string ThePathIsAlreadySelectedAsAnOutput(ObjectPath objectPath)
-         {
-            return $"The path {objectPath} is already selected as an output.";
-         }
+         public static string ThePathIsAlreadySelectedAsAnOutput(ObjectPath objectPath) => $"The path {objectPath} is already selected as an output.";
       }
 
       public static readonly string TimeColumnName = "Simulationtime";
--- a/src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
+++ b/src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
@@ -1,0 +1,150 @@
+﻿using System.Linq;
+using MoBi.Assets;
+using MoBi.Presentation.Presenter;
+using OSPSuite.Core.Commands;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Repositories;
+using OSPSuite.Core.Extensions;
+using OSPSuite.Core.Services;
+
+namespace MoBi.Presentation.Tasks
+{
+   public interface IOutputSelectionsTask
+   {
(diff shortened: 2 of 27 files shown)


src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
```diff
@@ -0,0 +1,150 @@
...
+      public void AddOutputSelection(SimulationSettings simulationSettings, QuantitySelection preSelectedQuantitySelection = null)
+      {
+         var pathFromSimulations = selectPathFromSimulationsFor(AppConstants.Captions.AddDefaultCurveForNewSimulations, preSelectedQuantitySelection);
+         if(pathFromSimulations == null)
+            return;
+
+         if (!validateAndWarnNewPath(pathFromSimulations, simulationSettings))
+            return;
+
+         simulationSettings.OutputSelections.AddOutput(new QuantitySelection(pathFromSimulations, QuantityType.Undefined));
```
rwmcintosh, line 61: ```QuantityType.Undefined```? I don't know about this. The quantity type is not needed for this view, but I think it's still required in the domain object.

When I tested, the simulation view shows the correct QuantityTypes after the simulation is created

src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
```diff
@@ -0,0 +1,150 @@
...
+      {
+         return newPath == null || Equals(selection.Path, newPath.ToPathString());
+      }
+
+      private ObjectPath selectPathFromSimulationsFor(string caption, QuantitySelection preSelectedQuantity = null)
+      {
+         using (var modalPresenter = _applicationController.Start<IModalPresenter>())
+         {
+            modalPresenter.Text = caption;
+            var selectPathPresenter = _applicationController.Start<IHierarchicalQuantitySelectionPresenter>();
```
rwmcintosh, line 101: All this Hierarchical Selection Presenter stuff is MoBi-specific. At first, I tried to move it to Core, but there were a few MoBi-specific things I couldn't move easily.

When I looked at the implementation of other SimulationSettingsItems presenters, they were all implemented in MoBi, so I didn't think it would be the worst idea to move output selections there as well.

I could have only added the interface ```IOutputSelectionsTask``` to Core, and implemented it in MoBi, but I chose to move the presenter to MoBi instead.

src/MoBi.Assets/AppConstants.cs
```diff
@@ -2030,6 +2032,11 @@ public static string NameIsAlreadyUsedInThisContainer(string containerPath, stri
          {
             return $"A value with the name '{name}' already exists for the container {containerPath}";
          }
+
+         public static string ThePathIsAlreadySelectedAsAnOutput(ObjectPath objectPath)
```
msevestre, line 2036: inline?

src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
msevestre: hum this will be used later to filter match entries by path etc when using the chart template
Can't you find it from the selection? (e.g. pathFromSimulations => instead of string, path and quantitytype?)

src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
```diff
@@ -0,0 +1,150 @@
...
+         if (!alreadyContainsPath(selectedPath, simulationSettings)) 
+            return true;
+         
+         showWarningMessage(selectedPath);
+         return false;
+      }
+
+      public void UpdateOutputSelection(SimulationSettings simulationSettings, QuantitySelection selection, string newPath)
+      {
+         if (noChangesRequired(selection, new ObjectPath(newPath.ToPathArray())))
```
msevestre, line 126: maybe a private method to create and objectPath from a path as string (don't we have that already? somewhere)

msevestre (commented):
LKooks good to me.
I am wondering only about the Undefined. IF feelt that since you are selecting an output, you should know what it is

rwmcintosh:
> LKooks good to me. I am wondering only about the Undefined. IF feelt that since you are selecting an output, you should know what it is

Well, **if** it is selected from the tree then that might work. But what about the situation where it's hand added, or hand edited?

src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
rwmcintosh: Yeah, probably, but it could be hand edited after as well. When I use this default output selection to create new simulations, then they look correct in the simulation settings. 

image

src/MoBi.Presentation/Tasks/OutputSelectionsTask.cs
rwmcintosh: We have ObjectPathFactory, but you still have to use ```xxx.ToPathArray```

Should we add that in Core? Something like?

```
// something new
      public FormulaUsablePath CreateFormulaUsablePathFrom(string path)
      {
         return CreateFormulaUsablePathFrom(path.ToPathArray());
      }

// this is already there
      public FormulaUsablePath CreateFormulaUsablePathFrom(params string[] entries)
      {
         return CreateFormulaUsablePathFrom(entries.ToList());
      }
```

Or just an extension method in MoBi?

PR #1273, opened 2024-02-15
Hotfix 11.3 core update

rwmcintosh:
This just updates core version for hotfix 11.3

Unsure about the nightly build trigger since it does not seem to have a way to trigger from a specific branch in Appveyor-nightly.yml

### DIFF
--- a/appveyor-nightly.yml
+++ b/appveyor-nightly.yml
@@ -4,7 +4,7 @@ image: Visual Studio 2019
 version: '{build}'
 
 init:
-- ps: Update-AppveyorBuild -Version "$($env:ospsuite_version).$($env:appveyor_build_version)"
+- ps: Update-AppveyorBuild -Version "11.3.$($env:appveyor_build_version)"
 
 environment:
   MIK_TEX_INSTALL_DIR: 'C:\MikTex\'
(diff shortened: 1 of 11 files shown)


appveyor-nightly.yml
```diff
@@ -1,64 +1,64 @@
...
-
-skip_commits:
-  files:
+configuration: Debug
+image: Visual Studio 2019
+
+version: '{build}'
+
+init:
+- ps: Update-AppveyorBuild -Version "11.3.$($env:appveyor_build_version)"
```
msevestre, line 7: wondering if there is not a better way of doing this
e.g. define a variable in the file that points to this value and then use it. No idea. Anyways good for now

appveyor.yml
```diff
@@ -1,58 +1,58 @@
...
-
-pull_requests: 
-  do_not_increment_build_number: true     
+configuration: Debug
+image: Visual Studio 2019
+
+version: '{build}'
+
+init:
+- ps: Update-AppveyorBuild -Version "11.3.$($env:appveyor_build_version)"
```
msevestre, line 7: not sure if we need to change this one. It's not were the build are created from

msevestre (approved):

PR #1286, opened 2024-03-22
Fixes #1285 Neighborhoods are not exported for compartments

### LINKED ISSUE #1285: Neighborhoods are not imported for compartments
When saving an organ to pkml and loading it in a new module, the neighborhoods within the organ and between the organ and the rest of the body (blood organs) are imported. E.g., exporting "Bone" and importing it in a new module:

image

The behavior is **as expected**.

However, if I export "Bone|Intracelluar" only and load it in a new module, no neighborhoods are created:

image

I expect the neighborhood `Bone_int_Bone_cell` being loaded as well.

rwmcintosh:
Fixes #1285

# Description
We were using the wrong overload to append paths together. The extension method first converts the object path to string, then appends the whole string into one element of the object path. So, ```Bone|Intracellular``` was not two elements, but one. 

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
+++ b/src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
@@ -28,8 +28,7 @@ namespace MoBi.Core.Domain.Model
             if (parentContainer?.ParentPath == null)
                return absolutePath;
 
-            absolutePath.AddAtFront(parentContainer.ParentPath);
-            return absolutePath;
+            return absolutePath.AndAddAtFront(parentContainer.ParentPath);
          }
 
          //Returns all possible physical containers that can be taken into consideration
--- a/src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
+++ b/src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
@@ -28,7 +28,8 @@ namespace MoBi.Core.Domain.Model
             if (parentContainer?.ParentPath == null)
                return absolutePath;
 
-            return absolutePath.AndAddAtFront(parentContainer.ParentPath);
+            absolutePath.AddAtFront(parentContainer.ParentPath);
+            return absolutePath;
          }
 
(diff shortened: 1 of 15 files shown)


src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
```diff
@@ -28,7 +28,8 @@ ObjectPath getContainerPath(IContainer container)
             if (parentContainer?.ParentPath == null)
                return absolutePath;
 
-            return absolutePath.AndAddAtFront(parentContainer.ParentPath);
+            absolutePath.AddAtFront(parentContainer.ParentPath);
```
rwmcintosh, line 31: Adds paths by appending two ```ObjectPath``` together

src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
```diff
@@ -28,7 +28,8 @@ ObjectPath getContainerPath(IContainer container)
             if (parentContainer?.ParentPath == null)
                return absolutePath;
 
-            return absolutePath.AndAddAtFront(parentContainer.ParentPath);
```
rwmcintosh, line 31: Adds paths by appending ```string```.

src/MoBi.Core/Domain/Model/MoBiSpatialStructure.cs
msevestre: this is NOT The fix. The fix is to make the two methods behave the same. One is just syntax sugar over the other

msevestre (changes requested):
Root cause is found but the error will happen again. We nee dto make those two methods the same

rwmcintosh:
I'm going to update core, revert the change in code, but keep the test. I'll use this PR to update core version when it builds

PR #1288, opened 2024-03-25
Fixes #1218 "Refresh from source" for Initial Conditions

### LINKED ISSUE #1218: "Refresh from source" for Initial Conditions
We have removed "Refresh from srouce" for IC and PV. While the function does not make sense for PV (PV BB should only contain values different from source, so "refreshing" equals to deliting these entries), it should be supported by IC.

Example: I want to define start values of a molecule by an equation. If I change the start value in the molecules-BB, there is currently no way to propagate it to the IC-BB.

rwmcintosh:
Fixes #1218

# Description
Bringing back the previous functionality of refreshing initial conditions from a spatial structure and molecules.

Previously, we always refreshed from the template building blocks, but here the user has to choose what molecules and spatial structure to base their refresh from. That follows the pattern of ```Extend```.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
+++ b/src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
@@ -5,6 +5,7 @@ using MoBi.Core.Commands;
 using MoBi.Core.Domain.Builder;
 using MoBi.Core.Domain.Extensions;
 using MoBi.Core.Domain.Services;
+using MoBi.Core.Services;
 using MoBi.Presentation.DTO;
 using MoBi.Presentation.Mappers;
 using MoBi.Presentation.Tasks.Edit;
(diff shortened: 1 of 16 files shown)


src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -132,6 +140,55 @@ public IMoBiCommand UpdateInitialConditionScaleDivisor(TBuildingBlock buildingBl
...
+            return new MoBiEmptyCommand();
+
+         var macroCommand = new MoBiMacroCommand
+         {
+            CommandType = AppConstants.Commands.EditCommand,
+            Description = AppConstants.Commands.RefreshInitialConditionsFromBuildingBlocks,
+            ObjectType = ObjectTypes.InitialCondition
+         };
+
+         initialConditions.Each(initialCondition =>
```
msevestre, line 158: neat

msevestre (approved):
looks good to me at first sight.

PR #1300, opened 2024-03-27
Fixes #1290 When loading an organ, individual parameters are not created

### LINKED ISSUE #1290: When loading an organ, individual parameters are not created.
1. Load a PK-Sim model in MoBi
2. Export an organ as PKML and select an individual to create the parameters from
3. Create a new module and load the organ in this module
4. None of the individual-specific parameters are created:

image

rwmcintosh:
Fixes #1290

# Description
This brings back code that exports ```IndividualParameters``` in the container itself when exporting a container with individual and expression. ```ExpressionParameters``` are still exported as ```ParameterValues```

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
+++ b/src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
@@ -11,6 +11,8 @@ using MoBi.Presentation.Tasks.Interaction;
 using OSPSuite.Core.Commands.Core;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
+using OSPSuite.Core.Domain.Formulas;
+using OSPSuite.Core.Domain.Mappers;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Extensions;
 using OSPSuite.Utility.Extensions;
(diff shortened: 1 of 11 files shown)


src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
```diff
@@ -86,6 +94,66 @@ public override void Save(IContainer container)
          exportContainer(container, fileName);
       }
 
+      private void addIndividualParametersToContainers(IndividualBuildingBlock individual, IContainer container)
```
Yuri05, line 97: would rename to `addIndividualParametersToContainer` (singular)
(and before rename the current `addIndividualParametersToContainer` to `addIndividualParametersToContainerByPath` or something like that)

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
```diff
@@ -86,6 +94,66 @@ public override void Save(IContainer container)
...
+            addIndividualParametersToContainer(individual, subContainer, new ObjectPath(subContainerPath));
+         });
+      }
+
+      private void addIndividualParametersToContainer(IndividualBuildingBlock individual, IContainer container, ObjectPath containerPath)
+      {
+         var individualParametersToExport = individual.Where(individualParameter => individualParameter.ContainerPath.StartsWith(containerPath)).ToList();
+
+         // Ensure that the distributed parameters are added to the container first. After all distributed parameters are added, then the other parameters can be added.
+         individualParametersToExport.Where(x => x.DistributionType != null).Each(x => addIndividualParameterToContainer(x, container, containerPath));
```
Yuri05, line 115: shouldn't we better use `.Where(x => !x.IsDistributed())` instead?

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
```diff
@@ -86,6 +94,66 @@ public override void Save(IContainer container)
...
+      }
+
+      private IParameter createParameter(IndividualParameter individualParameter)
+      {
+         var parameterToAdd = _individualParameterToParameterMapper.MapFrom(individualParameter);
+
+         if (individualParameter.Formula != null)
+            parameterToAdd.Formula = individualParameter.Formula;
+         if (individualParameter.Value.HasValue)
+            parameterToAdd.Formula = _formulaFactory.ConstantFormula(individualParameter.Value.Value, individualParameter.Dimension);
```
Yuri05, line 142: Questions on this:

a) shouldn't it be encapsulated in `IndividualParameterToParameterMapper.MapFrom()` ?
b) don't we need to **clone** the formula in `parameterToAdd.Formula = individualParameter.Formula;` ?
c) the second assignment overrides the first one (if the condition is true), so I would rather do it as
```
if (individualParameter.Value.HasValue)
...
else if (individualParameter.Formula != null)
...
```

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
rwmcintosh: I'm going with ```addIndividualParametersToContainerAndSubContainers``` for this one and ```addIndividualParametersToContainerByPath ``` for the other

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
rwmcintosh: That does not yet exist for ```PathAndValueEntity``` based entities. I'll add the extension in Core, then add a core update here to use it after you approve the core PR

src/MoBi.Presentation/Tasks/Edit/EditTaskForContainer.cs
rwmcintosh: Well, now I'm not so sure this should be encapsulated. We are using the same mapper for model creation, so the cloning works differently. I'll put the else-if here and add a comment why this is not encapsulated.

Yuri05: ok

Yuri05 (approved):

PR #1303, opened 2024-04-05
Fixes #1302 Consolidate StartValues type presenters

### LINKED ISSUE #1302: Consolidate StartValues type presenters
There is a lot of code duplication that probably not necessary now and is making it more challenging to resolve bugs that are common in all of these presenter/view combinations like #1291.

It's probably time for this to get refactored

rwmcintosh:
Fixes #1302

# Description
As I refactored I found it useful to remove more of the old terminology ```StartValue```. We are calling these ```PathAndValueEntity``` now as the base type. Individual, Expression, Initial Conditions, Parameter Values are all derived from this.

The real difference between the first two and the last two is not 'Start Value', but the fact that the individual and expression cannot be extended, where the last two can be extended.

Then I tried to get rid of more naming and comments using the older 'Start Value' terminology, but did not complete. This will be ongoing I'm sure.

At the end of the day, there's much noise here and not much real change

## Type of change

- [x] Other (please describe):

A task that needs to be completed to make it more clear how the presenters and views and tasks are related

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [x] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/MoBi.Presentation/Presenter/PathAndValueBuildingBlockPresenter.cs
```diff
@@ -95,7 +102,7 @@ public virtual void SetUnit(TParameterDTO parameterDTO, Unit unit)
          SetUnit(parameterDTO.PathWithValueObject, unit);
       }
 
-      public void SetParameterValue(TParameterDTO parameterDTO, double? newValue)
```
rwmcintosh, line 98: This had been implemented identically a couple of times with a different name

Yuri05 (approved):

PR #1315, opened 2024-04-17
Fixes #1291 Cannot change value of a distributed parameter

### LINKED ISSUE #1291: Cannot change value of a distributed parameter
1. Exported an organ (Muscle) to pkml with an individual selected
2. Loaded in a new module
3. Parameter "Volume" is stored in the PV BB as distributed parameter
4. Cannot change its value

image

Best solution would be if the parameters from the individual were created in the container upon export/loading (related to https://github.com/Open-Systems-Pharmacology/MoBi/issues/1290).

So after export, PV BB only contains expression values.

rwmcintosh:
Fixes #1291

# Description
Changing a distributed parameter will now update the parameter so that it no longer has a distribution type or any sub-parameters in the building block.

The value of the parameter will come from the `MEAN` Subparameter.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

When you view a distributed Individual Parameter
image

Click the edit button converts the parameter to a constant value parameter with the value set to whatever the `MEAN` was before. It also removes sub-parameters from the building block.

image
This is immediate after pressing edit. You can type a new number.

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForPathAndValueEntity.cs
+++ b/src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForPathAndValueEntity.cs
@@ -3,6 +3,7 @@ using MoBi.Assets;
 using MoBi.Core.Commands;
 using MoBi.Core.Domain.Model;
 using MoBi.Core.Domain.Services;
+using MoBi.Core.Helper;
 using MoBi.Presentation.Tasks.Edit;
 using OSPSuite.Core.Commands.Core;
 using OSPSuite.Core.Domain;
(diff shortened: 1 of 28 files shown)


src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForPathAndValueEntity.cs
```diff
@@ -75,7 +78,32 @@ public ICommand SetValueOrigin(TBuildingBlock buildingBlock, ValueOrigin valueOr
       {
          return new UpdateValueOriginInPathAndValueEntityCommand<TBuilder>(pathAndValueEntity, valueOrigin, buildingBlock).Run(Context);
       }
-      
+
+      public IMoBiCommand ConvertDistributedParameterToConstantParameter(TBuilder distributedParameter, TBuildingBlock buildingBlock, IReadOnlyList<TBuilder> subParameters)
+      {
+         var meanEntity = subParameters.FindByName(Constants.Distribution.MEAN);
```
Yuri05, line 84: Why do we need to get the mean parameter and then create a constant parameter with `meanEntity.Value` instead of with `distributedParameter.Value`?
Especially because
a) not all distributions have a "Mean" sub-parameter (e.g. uniform distribution)
b) the value of the distributed parameter is not necessarily the mean value of the underlying distribution

rwmcintosh: This is not a real 'Parameter' or 'DistributedParameter'. It's just a collection of 'PathAndValueEntity' like 'ParameterValues'. So when converting this, there is no value in the `Value` property, nor a formula in the `Formula` field. Both are set to null. I believe the expectation is that the parameter is created when the model is constructed and the value will be calculated at that time.

This initial view of distributed parameter is arranged by the convention that the paths are coincidental for Mean/Percentile, or whatever subparameters are there.

So, not sure how to proceed with that. I could add other conventions than `Mean`

rwmcintosh: The type is stored there.

Yuri05: Yep, that was exactly what I thought. Create temporary parameter with the proper distribution and percentile and get its value, then save it

Yuri05 (approved):

PR #1342, opened 2024-05-08
Fixes #1269 When importing containers with additional parameter values, allow user to select target for parameter values

### LINKED ISSUE #1269: When importing containers with additional parameter values, allow user to select target for parameter values
You should be able to either select an existing building block, or create a new building block in the module.

Watch out for the container being renamed during import.

Follow up to #1145

rwmcintosh:
Fixes #1269

# Description
When a user imports a spatial structure transfer, it contains parameter values for the containers that were exported. Up to now, a new PV building block was always created for them. Now the user can select an existing one, or create a new one.

If an existing one is selected, then if there are colliding paths, the old parameter value is removed and the new one is added.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -260,6 +260,7 @@ namespace MoBi.Assets
          public static readonly string UpdateProjectDefaultSimulationSettings = "Update project default simulation settings";
          public static readonly string RemoveTrackedQuantityChanges = "Remove tracked quantity changes";
          public static readonly string AddedMultipleBuildingBlocksFromFile = "Added multiple building blocks from file";
+         public static readonly string AddOrReplaceCommand = "Add or replace";
 
          public static string ConvertDistributedPathAndValueEntityToConstantValue(string type, string path) => $"Convert distributed {type} '{path}' to constant value";
 
@@ -804,6 +805,11 @@ namespace MoBi.Assets
          {
             return $"Changed the simulation configuration of simulation '{simulationName}'";
          }
+
+         public static string AddOrReplaceParameterValuesToBuildingBlock(string buildingBlockName)
+         {
+            return $"Add or replace multiple parameter values to '{buildingBlockName}'";
+         }
       }
 
       public static class BarNames
@@ -860,6 +866,7 @@ namespace MoBi.Assets
          public static readonly string ApplyDefaultNaming = "Apply default renaming";
          public static readonly string RemoveSelectedResultsFromSimulations = "Do you really want to remove the selected results from simulations";
          public static readonly string RemoveSelectedResultsFromProject = "Do you really want to remove the selected result(s) from the project";
+         public static Size SelectSingleSize = new Size(475, 160);
 
          public static string RemoveSimulationsFromProject(string projectName)
          {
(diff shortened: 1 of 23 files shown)


src/MoBi.Assets/AppConstants.cs
```diff
@@ -860,6 +860,7 @@ public static class Dialog
          public static readonly string ApplyDefaultNaming = "Apply default renaming";
          public static readonly string RemoveSelectedResultsFromSimulations = "Do you really want to remove the selected results from simulations";
          public static readonly string RemoveSelectedResultsFromProject = "Do you really want to remove the selected result(s) from the project";
+         public static readonly Size SELECT_SINGLE_SIZE = new Size(475, 160);
```
msevestre, line 863: all the sizes are defined in another file no ? UICOnstants or sthg? or am I wrong

src/MoBi.Presentation/Presenter/SelectManyPresenter.cs
```diff
@@ -0,0 +1,31 @@
...
+   {
+      protected SelectManyPresenter(ISelectManyView<T> view, IItemToListItemMapper<T> itemToListItemMapper)
+         : base(view, itemToListItemMapper)
+      {
+
+      }
+
+      protected override IEnumerable<ListItemDTO<T>> MapAllItems()
+      {
+         return base.MapAllItems().OrderBy(x => x.DisplayName);
```
msevestre, line 26: why would the behavior be different than in the base class? Any reason not to move this up?

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForContainerBase.cs
```diff
@@ -121,6 +127,48 @@ private ICommand addParameterValues(ParameterValuesBuildingBlock parameterValues
...
+            return null;
+
+         using (var modal = ApplicationController.Start<IModalPresenter>())
+         {
+            var presenter = ApplicationController.Start<ISelectSinglePresenter<ParameterValuesBuildingBlock>>();
+            presenter.SetDescription(AppConstants.Captions.SelectTheBuildingBlockWhereParameterValuesWillBeAddedOrUpdated);
+            modal.Text = AppConstants.Captions.SelectParameterValuesBuildingBlock;
+            modal.Encapsulate(presenter);
+            
+            presenter.InitializeWith(new List<ParameterValuesBuildingBlock>
```
msevestre, line 158: I think this should be rewritten
 var toto = new List<PVV>(moduleBuildingBlockks){NullPathAndValueEntityBuildingBlocks.NewParameterValues}

Or is this not correct?

src/MoBi.Presentation/Views/IContainerModalView.cs
```diff
@@ -6,5 +7,7 @@ public interface IContainerModalView : IModalView<IModalPresenter>
    {
       void AddSubView(IView subView);
       bool Show();
+      bool Show(Size modalSize);
```
msevestre, line 10: probably size could be optional with a default one? not sure if this is use often

msevestre (changes requested):

src/MoBi.Assets/AppConstants.cs
rwmcintosh: Yeah, but in this case we want to set the size of the modal from the presentation layer. UIConstants is in MoBi.UI. I agree this is a little out of place, but I don't think I can put it in UIConstants.

Maybe a separate static class in this file because it's a bit out of place with these strings.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForContainerBase.cs
rwmcintosh: The goal is to have the `NewParameterValues` at the top of the list. I have this other comment from you about the `OrderBy`, so I'm thinking that the presenter could maybe default `OrderBy`. Then here, where I am adding this special item, I can place it at the top somehow.

src/MoBi.Assets/AppConstants.cs
msevestre: why do we set the size from the presentation layer?

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForContainerBase.cs
msevestre: oh so the order is important? Then forget it/ But I think a comment would help to understand why the syntax is unexpected

src/MoBi.Assets/AppConstants.cs
rwmcintosh: This is a modal that encapsulates another view. So the pattern is to resolve this modal in a presenter, call encapsulate with another presenters view. The modal has no idea what's there.

src/MoBi.Presentation/Presenter/SelectManyPresenter.cs
rwmcintosh: I changed this to allow the caller to override the default sorting from `InitializeWith`. Given no argument, the sort will default to the object name.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForContainerBase.cs
rwmcintosh: Yes, order is important, but I got to thinking that maybe it's not important for the SelectManyPresenter, but only for the particular caller.

The order is only important in this case because of this special building block that should be at the top.

Anyway, it's been replaced with something that allows the caller to modify the order while initializing.

msevestre (approved):

PR #1346, opened 2024-05-14
Fixes #1085 Error when modifying a table formula with x argument

### LINKED ISSUE #1085: Error when modifying a table formula with x argument
When trying to change path to table object or to a-argument of a table formula with x argument, an error is thrown:

image

Application:
MoBi® 12.0.45

Stack trace:
```
at MoBi.Presentation.Presenter.EditTableFormulaWithXArgumentFormulaPresenter.selectFormulaUsablePath(Func`2 predicate, String caption)
   at MoBi.Presentation.Presenter.EditTableFormulaWithXArgumentFormulaPresenter.SetXArgumentFormulaPath()
   at OSPSuite.UI.Controls.BaseUserControl.<>c__DisplayClass23_0.<OnEvent>b__0()
   at OSPSuite.Utility.Extensions.ExceptionExtensions.DoWithinExceptionHandler(Object callerObject, Action actionToExecute)
```

Already in v11.

rwmcintosh:
Fixes #1085

# Description
In this case, we are editing the formula from the formula cache and do not have the context of what is using the formula. There seems to be some provisions for that and I just made some more so the application will not crash.

I'm not sure if there's another way that would be preferred. Essentially, if you edit from the formula cache, you lose the ability to specify relative paths (it seems). It seems to make sense to me.

For now, I would like a quick comment on whether this approach makes sense, then I'll add some tests for it.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditTableFormulaWithXArgumentFormulaPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditTableFormulaWithXArgumentFormulaPresenter.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using MoBi.Assets;
 using MoBi.Core.Domain.Services;
 using MoBi.Presentation.DTO;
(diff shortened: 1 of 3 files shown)


src/MoBi.Presentation/Presenter/EditTableFormulaWithXArgumentFormulaPresenter.cs
```diff
@@ -87,8 +88,8 @@ private FormulaUsablePath selectFormulaUsablePath(Func<IObjectBase, bool> predic
       {
          using (var presenter = _applicationController.Start<ISelectFormulaUsablePathPresenter>())
          {
-            var referencePresenter = _selectReferencePresenterFactory.ReferenceAtParameterFor(UsingObject.ParentContainer);
-            presenter.Init(predicate, UsingObject, new[] {UsingObject.RootContainer}, caption, referencePresenter);
+            var referencePresenter = _selectReferencePresenterFactory.ReferenceAtParameterFor(UsingObject?.ParentContainer);
```
msevestre, line 91: Why is UsingObject null? isn't it the problem?

msevestre (approved):
looks good but I am wondering why we are  guarding ourselves against an object that is null that maybe should not be null in the first place. Should the caller be changed?
Approved otherwise if this makes sense

PR #1348, opened 2024-05-16
1347 UI for default merge behavior for modules

### LINKED ISSUE #1347: UI for default merge behavior for modules

rwmcintosh:
Fixes #1347 

# Description
This is just the UI part, not any implementation of merge behavior

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

image

Context menu with check state
image

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1606,8 +1606,8 @@ namespace MoBi.Assets
          public static readonly string MakeDefault = "Make defaults";
          public static readonly string LoadFromDefaults = "Load from defaults";
          public static readonly string DefaultMergeBehavior = "Default Merge Behavior";
-         public static readonly string ExtendMergeBehaviorDescription = "<b>The module containers will be <i>merged recursively using add and update behavior</i></b>";
-         public static readonly string OverwriteMergeBehaviorDescription = "<b>The module containers will be <i>replaced</i> by path</b>";
+         public static readonly string ExtendMergeBehaviorDescription = "The module containers will be merged recursively using add and update behavior";
+         public static readonly string OverwriteMergeBehaviorDescription = "The module containers will be replaced by path";
 
          public static string SelectEntitiesThatWillBeReplaced(string entityType) => $"Select {entityType.Pluralize()} that will be replaced";
          public static string SelectEntitiesThatWillBeReplacedDescription(string entityType, string buildingBlockName) => $"<b>Selected <i>{entityType.Pluralize()}</i> will replace existing <i>{entityType.Pluralize()}</i> in the building block <i>{buildingBlockName}</i></b>";
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1593,6 +1593,8 @@ namespace MoBi.Assets
          public static readonly string AddExpressionDescription = "<b>The building block will be extended with expression parameter values for selected <i>molecules</i> in the selected <i>organ</i> </b>";
          public static readonly string AddDefaultCurveForNewSimulations = "Add default curve for new simulations";
          public static readonly string ChangeDefaultCurveForNewSimulations = "Change default curve for new simulations";
+         public static readonly string DefaultMergeBehavior = "Default Merge Behavior";
+
          public static string SelectEntitiesThatWillBeReplaced(string entityType) => $"Select {entityType.Pluralize()} that will be replaced";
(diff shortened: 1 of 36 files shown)


src/MoBi.Assets/AppConstants.cs
```diff
@@ -804,6 +804,11 @@ public static string ChangeSimulationConfiguration(string simulationName)
          {
             return $"Changed the simulation configuration of simulation '{simulationName}'";
          }
+
+         public static string SetModuleMergeBehavior(string moduleName, string newMergeBehavior)
+         {
+            return $"Set merge behavior to '{newMergeBehavior}' for module '{moduleName}'";
```
msevestre, line 810: Should we have as text in the UI that says what this means? (When you create the module?)

src/MoBi.Core/Commands/SetDefaultMergeBehaviorCommand.cs
```diff
@@ -0,0 +1,42 @@
...
+
+      protected override void ExecuteWith(IMoBiContext context)
+      {
+         _oldBehavior = _module.DefaultMergeBehavior;
+         _module.DefaultMergeBehavior = _newMergeBehavior;
+      }
+
+      protected override void ClearReferences() => _module = null;
+
+      protected override ICommand<IMoBiContext> GetInverseCommand(IMoBiContext context) => new SetDefaultMergeBehaviorCommand(_module, _oldBehavior);
```
msevestre, line 38: the AsInversteFor is missing here

msevestre: can you check that this is not forgotten in other commands you wrote lately. I may have missed it

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForModule.cs
```diff
@@ -46,6 +48,29 @@ public IContextMenu InitializeWith(ObjectBaseDTO dto, IPresenter presenter)
...
+               .WithCheckedAction(defaultOverwrite => makeModuleOverwrite(module, defaultOverwrite)))
+            .WithItem(CreateMenuCheckButton.WithCaption(MergeBehavior.Extend.ToString())
+               .WithChecked(module.DefaultMergeBehavior == MergeBehavior.Extend)
+               .WithCheckedAction(defaultExtend => makeModuleExtend(module, defaultExtend)));
+      }
+
+      private void makeModuleExtend(Module module, bool defaultExtend)
+      {
+         if (defaultExtend && module.DefaultMergeBehavior != MergeBehavior.Extend)
+            _container.Resolve<MakeExtendModuleUICommand>().For(module).ExecuteWithinExceptionHandler();
```
msevestre, line 65: Why do this look all so weird?

msevestre (changes requested):
AsInverse is missing
I'd love to see an explanation in the UI of what the merge option mean

src/MoBi.Assets/AppConstants.cs
rwmcintosh: ok, but that's the command description. I'll make the combobox where you pick it have an explanation and I'll probably switch the explanation when the combobox changes.

rwmcintosh: What about like this?
image

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForModule.cs
rwmcintosh: It's because the MenuBarCheckItem overrides the default Click handler from MenuBarItem.

Do you think we should just change it in Core? We probably just need to check if there's a command registered and if so, call it. We wouldn't want to call it unless there's a real command though because all the spots where this is used up to now do not use it with a command, so they will have the `NotImplementedCommand`

```
   public class MenuBarCheckButton : MenuBarButton, IMenuBarCheckItem
   {
      private bool _checked;
      public event Action<bool> CheckedChanged = delegate { };

      public bool Checked
      {
         set { _checked = value; CheckedChanged(value); }
         get { return _checked; }
      }

      public override void Click()
      {
         Checked = !Checked;
      }
   }

   public class MenuBarButton : MenuBarItem, IMenuBarButton
   {
      public IUICommand Command { get; set; }

      public MenuBarButton()
      {
         Command = new NotImplementedCommand();
      }

      public virtual void Click()
      {
         Command.ExecuteWithinExceptionHandler();
      }
   }
```

PR #1349, opened 2024-05-16
Fixes #1022

### LINKED ISSUE #1022: Neighbors in the "Neighborhoods"-tree - what information to show?
In the tree view of a spatial structure, neighbors of a neighborhood are listed as sub-entries of a neighborhood:

image

Selecting these entries does not show any information in the frame to the right, and the information of the previously selected entity is shown. Example - I first selected "SmallIntestine" and then the neighbor of the neighborhood. The frame to the right shows Properties of "SmallIntestine".

image

I think this might be confusing for the user. Do we need these sub-entries? I propose to either

1) Remove these entries, or
2) Show the information of the selected container

e.g. for this selected entry
image

Show "Edit view" of `Organism|Brain|Interstitial`. so this entry is simply a shortcut to the container. I think this could be very useful, as the user could quickly see what parameters of neighbors are referenced in the neighborhood's parameters.

rwmcintosh:
Fixes #1022

# Description
When you click on neighbors in the spatial structure view the neighbor node will act as a link to the container that it references. So you can edit the container that the neighbor references when you select the neighbor

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/HierarchicalStructureEditPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/HierarchicalStructureEditPresenter.cs
@@ -34,14 +34,12 @@ namespace MoBi.Presentation.Presenter
             RaiseUserDefinedSelectedEvent();
 
          else
-            raiseEntitySelectedEvent(objectBaseDTO);
+            RaiseEntitySelectedEvent(objectBaseDTO);
       }
 
-      private void raiseEntitySelectedEvent(ObjectBaseDTO objectBaseDTO)
+      protected virtual void RaiseEntitySelectedEvent(ObjectBaseDTO objectBaseDTO)
       {
-         // First and second neighbor node selections should not trigger an EntitySelectedEvent
-         // because they are not a selectable entity
-         if(objectBaseDTO.ObjectBase != null)
+         if(!(objectBaseDTO is NeighborDTO))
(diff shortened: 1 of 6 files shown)


src/MoBi.Presentation/Presenter/HierarchicalStructureEditPresenter.cs
```diff
@@ -34,14 +34,12 @@ public virtual void Select(ObjectBaseDTO objectBaseDTO)
...
+            RaiseEntitySelectedEvent(objectBaseDTO);
       }
 
-      private void raiseEntitySelectedEvent(ObjectBaseDTO objectBaseDTO)
+      protected virtual void RaiseEntitySelectedEvent(ObjectBaseDTO objectBaseDTO)
       {
-         // First and second neighbor node selections should not trigger an EntitySelectedEvent
-         // because they are not a selectable entity
-         if(objectBaseDTO.ObjectBase != null)
+         if(!(objectBaseDTO is NeighborDTO))
```
msevestre, line 42: This is funky. IHere you know that the DERIVED Class is going to overwrite for NEighbor. .. this is why you are bypassing
It's a bit of leak but why not. However what about a comment explaining why this is excluded like this?

msevestre (approved):

PR #1357, opened 2024-06-06
Fixes #159: The contents of the cell are now shown on the first row in the tooltip. 

### LINKED ISSUE #159: Favorites list: Tooltip should show the content of the cell under mouse cursor
In the favorites list of a simulation, only the name and description of the parameter under cursor is shown as tooltip. It would be of much better help if the content of the respective cell, e.g., the "Container" cell, would be shown.
tooltip

peterdew:
Fixes #159 

# Description

Added cellvalue of cell under mousecursor to the top of the tooltip.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image
image

# Questions (if appropriate):


rwmcintosh (approved):

rwmcintosh (commented):
@peterdew what do you think about adding a line to the second section of the tooltip?

So in your screenshot it does not just say `Concentration`, but instead

**Name**
Concentration

Similar to the other sections

rwmcintosh (changes requested):
We decided to add this section header as either

**Name**
Concentration

or

**Parameter Name**
Concentration

peterdew:
I felt it would be even nicer to keep the description where available and make a 3-line parameter section. 

**Parameter**
 Name
Description
--------------
**Other Section**
text

rwmcintosh (approved):

src/MoBi.UI/Services/ToolTipCreator.cs
```diff
@@ -13,7 +14,7 @@ namespace MoBi.UI.Services
    public interface IToolTipCreator : OSPSuite.UI.Services.IToolTipCreator
    {
       SuperToolTip ToolTipFor(NotificationMessage notificationMessage);
-      SuperToolTip ToolTipFor(IMoBiParameterDTO parameterDTO);
+      SuperToolTip ToolTipFor(IMoBiParameterDTO parameterDTO, string cellValue = null);
```
msevestre, line 17: this `cellValue` parameter name is very confusing to me. What is a cell in the context of a parameter? I think this should be renamed and ideally, an example added if the context is hard to explaijn

src/MoBi.UI/Services/ToolTipCreator.cs
```diff
@@ -65,8 +69,31 @@ private static SuperToolTip addFormulaToolTip(FormulaBuilderDTO formula, SuperTo
...
       {
-         var toolTip = CreateToolTip(calculationMethod.Description, calculationMethod.CalculationMethodName);
+         return CreateToolTip(calculationMethod.Description, calculationMethod.CalculationMethodName);
+      }
+
+      private SuperToolTip createCellValueToolTip(string cellValue)
+      {
+         if (cellValue == string.Empty)
+            return CreateToolTip();
+
```
msevestre, line 79: what's with the ...?

msevestre (commented):
image

I Don't understand the ... at the beginning? What does this mean

peterdew:
> image
> 
> I Don't understand the ... at the beginning? What does this mean

It gives an indication that the text in the grid cell is repeated. That text ends with ... when it is truncated due to lack of space. 
@PavelBal What do you think?

src/MoBi.UI/Services/ToolTipCreator.cs
peterdew: It's the implementation of what is discussed in another comment.

src/MoBi.UI/Services/ToolTipCreator.cs
peterdew: The tooltip is at a gridcell, and this cell has a value / text. The underlaying object is a parameter, but it has a lot of properties. You could argue that it is thus the property value that matters, but only the grid cell displays the value including the correct formatting. Hence, I chose the name cellValue

rwmcintosh:
What about like this
image

image

rwmcintosh:
I made the change to put the hovered column caption in the tooltip and also refactored and used that same tooltip creation technique in another grid parameter view.

PR #1360, opened 2024-06-10
Fixes #472-Copy-a-path-to-clipboard-from-any-entity-in-a-simulation

### LINKED ISSUE #472: Feature: copy a path to clipboard from any entity in a simulation
As proposed by @msevestre in https://github.com/Open-Systems-Pharmacology/Forum/issues/486 it would be a good feature to have a way to copy a path to clipboard from any entity in a simulation. (for example for a parameter, a container etc..)

peterdew:
Fixes #472 

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

image
(from clipboardmanager: )
image

# Questions (if appropriate):


src/MoBi.Presentation/Presenter/HierarchicalSimulationPresenter.cs
```diff
@@ -1,40 +1,45 @@
...
 using OSPSuite.Presentation.Core;
 using OSPSuite.Presentation.Presenters;
 using OSPSuite.Presentation.Presenters.ContextMenus;
+using OSPSuite.Utility.Events;
+using OSPSuite.Utility.Extensions;
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Linq;
+using System.Windows.Forms;
```
rwmcintosh, line 20: Can you check that these are all necessary? I'm especially worried about dependency on System.Windows.Forms, but sometimes they get added during refactoring and then become stale. Check all the files you changed please.

We normally don't have a dependency on System.Windows.Forms in the Presentation layer.

tests/MoBi.Tests/Presentation/HierarchicalSimulationPresenterSpecs.cs
```diff
@@ -37,8 +39,9 @@ protected override void Context()
          _simulationSettingsMapper = A.Fake<ISimulationSettingsToObjectBaseDTOMapper>();
          _dtoMapper = A.Fake<IObjectBaseToObjectBaseDTOMapper>();
          _neighborhoodDTOMapper = A.Fake<INeighborhoodToNeighborDTOMapper>();
+         _entityPathResolver = A.Fake<IEntityPathResolver>();
 
-         _favorites = new ObjectBaseDTO
+            _favorites = new ObjectBaseDTO
```
rwmcintosh, line 44: spacing is off for this line.

rwmcintosh (changes requested):
There's just a question about the 'using' directives that are in some files and also, a spot where spacing is off.

If you think the single line methods do not have a very long line, then consider using expression body.

Nice feature!

src/MoBi.Presentation/Presenter/HierarchicalSimulationPresenter.cs
peterdew: It's because of the Clipboard static class I use now in the CopyCurrentPathToClipBoard-method.

peterdew: I also added OSPSuite.Core.Domain.Services for the EntityPathResolver. No other usings added.

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForSimulationEntities.cs
```diff
@@ -0,0 +1,30 @@
...
+
+      public ContextMenuForSimulationEntities(ObjectBaseDTO dto, IMoBiContext context, IHierarchicalSimulationPresenter presenter)
+      {
+         _presenter = presenter;
+         _entity = context.Get<IEntity>(dto.Id);
+      }
+
+      public override IEnumerable<IMenuBarItem> AllMenuItems()
+      {
+         yield return CreateMenuButton.WithCaption("Copy path")
```
rwmcintosh, line 25: Can you move "Copy path" into `MoBi.Assets.Captions`

src/MoBi.Presentation/Presenter/HierarchicalSimulationPresenter.cs
rwmcintosh: We should move the ClipBoard code into the UI layer as discussed so that the dependency on Forms can be moved.

peterdew: Moved the clipboard usage to the view and thus removed the dependency on System.Windows.Forms from the Presentation layer.

rwmcintosh (approved):

Issue #1308, opened 2024-04-16
"Refresh from source" for IC BB does not always refresh the formula cache view


PavelBal:
[Two-Compartment.zip](https://github.com/Open-Systems-Pharmacology/MoBi/files/14998333/Two-Compartment.zip)

In the attached project:

- Module "TowCompartmenalModel" has "Molecule B" which start amount was previously defined as "Container B|Volume". The IC BB of this module was extended and the formula has the path to the volume "Container B|Volume":

![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/c6ffa4a9-ecc8-4efd-955b-033fe84a457b)

- The path of the alias has been changed to "Organism|Container B|Volume":

![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/db3ac0d9-7ce3-48a4-b250-89536724d13d)

- Open IC BB in the module "TowCompartmenalModel"
- Select entries for "Molecule B" and click "Refresh selected from source"
- In the dialog, select the molecules from this module:

![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/c51d65c9-d4bb-4c53-af5d-e7e476b5d1fd)

- There are TWO formulas with the same name now:

![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/80aa906b-6453-44c4-a957-249f475abc7e)

Also this did not work first, it did not change the value at all, and I am not sure what exactly I did to finally manage to update. But the fact that there are two formulas with the same name is already wrong.

PavelBal:
This obviously does not work properly... Let's try this out live tomorrow.

Notes for myself:

- Change start value of Molecule A in the module "CompA" to 15
- IC of Module A: select "Container A|Molecule A" and select "Refresh selected from source"
- Select this:

![image](https://github.com/Open-Systems-Pharmacology/MoBi/assets/25903040/90286550-2fae-4e88-8ca2-3713d1af3ec2)

Do we need to select the spatial structure at all? I don't think so, the start values are coming from the Molecules BB only (constant value or a formula)

- Click refresh -> nothing changes

rwmcintosh:
We haven't been able to reproduce the 'refresh not working' part. Other issues mentioned have been fixed in related issues.

PR #1372, opened 2024-06-13
#1308 refresh from source

benjaperez1983:
Fixes #

# Description

Added Event to add formula on refresh and tests for each scenario

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
+++ b/src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
@@ -1,14 +1,17 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Remoting.Contexts;
 using MoBi.Assets;
 using MoBi.Core.Commands;
 using MoBi.Core.Domain.Builder;
 using MoBi.Core.Domain.Extensions;
 using MoBi.Core.Domain.Services;
+using MoBi.Core.Events;
 using MoBi.Core.Services;
 using MoBi.Presentation.DTO;
 using MoBi.Presentation.Mappers;
 using MoBi.Presentation.Tasks.Edit;
+using NPOI.POIFS.Properties;
 using OSPSuite.Assets;
 using OSPSuite.Core.Commands.Core;
 using OSPSuite.Core.Domain;
@@ -201,7 +204,11 @@ namespace MoBi.Presentation.Tasks.Interaction
          // The clone manager can return an existing formula from the formula cache if it is present and equivalent to the formula being cloned
          // This is important for renaming step. We can't just rename this formula if one already exists in the cache with the same name because
          // it might actually be the same one.
+         var formulaCount = buildingBlock.FormulaCache.Count;
          var clonedFormula = _cloneManagerForBuildingBlock.Clone(moleculeBuilder.DefaultStartFormula, buildingBlock.FormulaCache);
+
+         if (formulaCount != buildingBlock.FormulaCache.Count)
+            Context.PublishEvent(new AddedEvent<IFormula>(clonedFormula, buildingBlock));
 
          // Update this formula in the initial condition
          macroCommand.Add(ChangeValueFormulaCommand(buildingBlock, initialCondition,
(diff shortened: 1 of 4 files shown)


src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -1,14 +1,17 @@
 ﻿using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.Remoting.Contexts;
```
rwmcintosh, line 3: Check all the usings. You can use ReSharper to automatically clean up unnecessary ones. I'm sure this one and NPOI are not used, but they sometimes get added automatically by VS or ReSharper.

tests/MoBi.Tests/Presentation/Tasks/InitialConditionsTaskSpecs.cs
```diff
@@ -667,26 +668,38 @@ public void should_set_the_value_to_null()
...
+   }
+
+   public class When_updating_a_molecule_start_value_from_original_building_block_and_all_formulas_will_be_replaced : concern_for_initialConditions_With_formulas_for_Renaming
+   {
+      protected override void Context()
+      {
+         base.Context();
+         A.CallTo(() => _cloneManagerForBuildingBlock.Clone(_builder.DefaultStartFormula, _initialConditionsBuildingBlock.FormulaCache)).Invokes(x => _initialConditionsBuildingBlock.FormulaCache.Add(_refreshedFormula)).Returns(_refreshedFormula);
+         A.CallTo(() => _context.Context.PublishEvent(A<AddedEvent<IFormula>>._))
+            .Invokes(x => _addEvent = x.GetArgument<AddedEvent<IFormula>>(0));
```
rwmcintosh, line 702: Could you move `.Invokes` to the second line to be consistent with the second `A.CallTo` because I think they both benefit from the additional line due to length.

tests/MoBi.Tests/Presentation/Tasks/InitialConditionsTaskSpecs.cs
```diff
@@ -700,13 +713,50 @@ public void the_formula_that_was_refreshed_is_in_the_cache()
...
+
+
       [Observation]
       public void the_formula_that_was_replaced_is_removed_from_cache()
       {
          _initialConditionsBuildingBlock.FormulaCache.Contains(_replacedFormula).ShouldBeFalse();
       }
    }
 
+   public class When_updating_a_molecule_start_value_from_original_building_block_and_all_formulas_will_be_replaced_and_AddEvent_Only_executed_once : concern_for_initialConditions_With_formulas_for_Renaming
```
rwmcintosh, line 730: This name probably should indicate that `added_event_should_not_be_called` instead.

rwmcintosh (approved):
Just some names to reconsider and formatting for me.

rwmcintosh:
Fill in the # in the description

image

rwmcintosh (approved):

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
```diff
@@ -201,8 +202,12 @@ private void updateFormulaUsages(TBuildingBlock buildingBlock, IReadOnlyList<Ini
          // The clone manager can return an existing formula from the formula cache if it is present and equivalent to the formula being cloned
          // This is important for renaming step. We can't just rename this formula if one already exists in the cache with the same name because
          // it might actually be the same one.
+         var formulaCount = buildingBlock.FormulaCache.Count;
          var clonedFormula = _cloneManagerForBuildingBlock.Clone(moleculeBuilder.DefaultStartFormula, buildingBlock.FormulaCache);
 
+         if (formulaCount != buildingBlock.FormulaCache.Count)
```
msevestre, line 208: This required a comment for me explaining WHY the count can be different. It looks like it can never be the case but in fact it might be because of the side effect of the cloning. I don't like comment but this is where it is required

msevestre (changes requested):

msevestre:
also run resharper in the test file. It looks like it might be out of plcae

src/MoBi.Presentation/Tasks/Interaction/InitialConditionsTask.cs
benjaperez1983: This condition is to ensure we are not executing this AddEvent twice for the same formula
         Since _cloneManagerForBuildingBlock.Clone(moleculeBuilder.DefaultStartFormula, buildingBlock.FormulaCache)
         has the Cache as input parameter, and it is adding the formula if it is not already there
         We need to check if the count has changed to know if the formula was added
         If the count has changed, then we need to publish the event.
         
         Adding all this comment to the code anyways

PR #1390, opened 2024-06-28
1362 show module level in the possible referenced objects list

### LINKED ISSUE #1362: Show Module level in the "Possible referenced objects" list
A project with three modules:

- `TwoContainerModel`
  - Spatial structure `Organism` with two containers `Container A` and `Container B`
- `Second module`
  - Spatial structure `Organism` with a container `tmp`
- `Third module`
  - Spatial structure `Organism2` with a container `test`

When defining an equation, the "Possible referenced objects" list only groups by the names of spatial structure, but not by the modules. This results in containers from the two first modules being shown under one node:

image

This might be confusing. I suggest to add an additional grouping layer by the modules.

benjaperez1983:
Fixes #1362

# Description

Show Module level in the "Possible referenced objects" list 

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):


tests/MoBi.Tests/Presentation/SelectReferencePresenterSpecs.cs
```diff
@@ -229,4 +232,88 @@ public void should_return_an_empty_list()
          _result.ShouldBeEmpty();
       }
    }
+
+   public abstract class concern_for_SelectReferencePresenter_with_node_text : ContextSpecification<ISelectReferencePresenter>
```
msevestre, line 236: typically we derive from concern_for_SelectReferencePresenter
Why aren't you doing this here? you would inherit most of the fake already?

benjaperez1983: in fact, I started doing that, but, I need some things which are different. 
As this actual implementations instead of the fakes:
   _objectBaseDTOMapper = new ObjectBaseToObjectBaseDTOMapper();
   _referenceMapper = new ObjectBaseDTOToReferenceNodeMapper(_objectBaseDTOMapper);
   
   otherwise, the Id`s of the objects are fake and I get this random string on them.
   
   If I alter this for the previous tests, they fail, since they are relying on this fake implementation as mocks.
   
   So I figured out this should be a different class. 
   
   let me know your thoughts about it. Maybe I could take this to a different file?

msevestre:
you can just override the context and update those with the instance you want 
(even add a comment and say why you need to update). you just need to mark them as protected in the main class (which they should be anyways)

src/MoBi.Presentation/Presenter/SelectReferencePresenterBase.cs
```diff
@@ -241,7 +241,14 @@ private void selectInView(IEntity entityToSelect)
       protected void AddSpatialStructures()
       {
          var spatialStructures = _buildingBlockRepository.SpatialStructureCollection;
-         _view.AddNodes(spatialStructures.MapAllUsing(_referenceMapper));
+         var nodes = spatialStructures.Select(x =>
+         {
+            var mapped = _referenceMapper.MapFrom(x);
+            mapped.Text = x.DisplayName;
+            return mapped;
+         });
```
rwmcintosh, line 249: I didn't realize as we discussed this together.

There's an extension method called `WithText` and you can chain it. That would make this LINQ into

`var nodes = spatialStructures.Select(x => _referenceMapper.MapFrom(x).WithText(x.DisplayName));`

Which could also be inserted directly into `AddNodes` instead of declaring a variable (if you think it's still readable).

rwmcintosh (changes requested):

rwmcintosh (approved):

Yuri05 (approved):

PR #1394, opened 2024-07-09
#1278 Prevent parent containter from showing on selection

### LINKED ISSUE #1278: Selection of parent container path
When selecting the parent container of a top-level container of a module:

image

1. When showing container from the *same* module (here, the "Two-container module"), only show containers that are on parallel branches to the currently selected container. In this case, "Container A" (and all of his descendants, if present) should **not** be shown. Only show "Container B" (and its descendants, if present).
2. ~~[When selecting any container, *prepend* its "parent container path". E.g., "Container A" from "Second module" has the "Organism" container set as parent container. When the user selects "Container A" as the parent container for "Container A" from the "Two-container module" (as currently shown in the screenshot), the path that is created should be "Organism|Container A" (now, its only "Container A").]~~(#1330)
3. ~~[What happens when the user changes the reference container for any container, that is a parent for another? E.g. "Container B" from "Two-container module" is set as a parent to "Container A" in "Second module". The user changes the parent container of "Container A".
  a. **No** automatic changes in parent container paths, no verifications. Easy to implement, but in this case we have to add a message to the user somewhere.
  b. All other paths are adjusted, circular references are checked, sanity checks... Probably too complex to cover all possible use cases. Not V1.]~~(#1393)

### LINKED ISSUE #1330: Selection of parent container path - append parent path of the selected container
Original issue: https://github.com/Open-Systems-Pharmacology/MoBi/issues/1278

When selecting any container, prepend its "parent container path". E.g., "Container A" from "Second module" has the "Organism" container set as parent container. When the user selects "Container A" as the parent container for "Container A" from the "Two-container module" (as currently shown in the screenshot), the path that is created should be "Organism|Container A" (now, its only "Container A").

is actually done when creating neighborhoods:

image

benjaperez1983:
Fixes #1278
Fixes #1330

# Description

When selecting the parent container of a top-level container of a module,  only show containers that are on parallel branches to the currently selected container.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/SelectContainerPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/SelectContainerPresenter.cs
@@ -1,4 +1,6 @@
-﻿using MoBi.Assets;
+﻿using System.Linq;
+using FluentNHibernate.Utils;
+using MoBi.Assets;
 using MoBi.Core.Domain.Repository;
 using MoBi.Presentation.DTO;
 using MoBi.Presentation.Mappers;
@@ -6,8 +8,6 @@ using MoBi.Presentation.Tasks.Interaction;
 using MoBi.Presentation.Views;
 using OSPSuite.Core.Domain;
 using OSPSuite.Presentation.Presenters;
-using System.Linq;
-using FluentNHibernate.Utils;
 
 namespace MoBi.Presentation.Presenter
 {
@@ -22,7 +22,7 @@ namespace MoBi.Presentation.Presenter
       private readonly IModuleToModuleAndSpatialStructureDTOMapper _moduleToModuleAndSpatialStructureDTOMapper;
       private readonly IBuildingBlockRepository _buildingBlockRepository;
       private readonly IInteractionTasksForTopContainer _tasksForTopContainer;
-      
+
       public SelectContainerPresenter(
          ISelectObjectPathView view,
          ISelectContainerInTreePresenter selectContainerInTreePresenter,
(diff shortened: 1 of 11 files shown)


src/MoBi.Presentation/Presenter/SelectContainerPresenter.cs
```diff
@@ -32,18 +36,28 @@ public class SelectContainerPresenter : AbstractDisposablePresenter<ISelectObjec
...
+         var parentObjectPath = _tasksForTopContainer.BuildObjectPath(containerDTO.ObjectBase as IContainer);
+         var list = _buildingBlockRepository.SpatialStructureCollection.Select(x => _moduleToModuleAndSpatialStructureDTOMapper.MapFrom(x.Module))
+            .Select(item =>
+            {
+               item.SpatialStructure.TopContainers = item.SpatialStructure.TopContainers
+                  .Where(container => !parentObjectPath.Equals(_tasksForTopContainer.BuildObjectPath(container.ObjectBase as IContainer)))
+                  .ToList();
+               return item;
+            })
+            .ToList();
```
rwmcintosh, line 60: Can you build this list of DTOs first, then use the `Each` extension method to iterate through and remove the containers with matching paths. I think that will make it easier to read. I would also introduce a private method to do the work on each one. Something like

```
private void removeTopContainersWithPath(ModuleAndSpatialStructureDTO item, ObjectPath objectPathToRemove)
```

src/MoBi.Presentation/Presenter/SelectContainerPresenter.cs
```diff
@@ -32,19 +37,26 @@ public class SelectContainerPresenter : AbstractDisposablePresenter<ISelectObjec
...
+         init(excludedObjectPath);
          _view.Display();
          return _view.Canceled ? null : _selectContainerInTreePresenter.SelectedEntityPath;
       }
 
-      private void init()
+      private void init(ObjectPath excludedObjectPath)
       {
          var list = _buildingBlockRepository.SpatialStructureCollection.Select(x => _moduleToModuleAndSpatialStructureDTOMapper.MapFrom(x.Module)).ToList();
+         list.Each(item => removeTopContainersWithPath(item, excludedObjectPath));
```
rwmcintosh, line 53: and rename this `item` to `x`

rwmcintosh (approved):
Just a couple naming items to take care of.

Could you also go back and run ReSharper cleanup in the files to fix up the spacing between methods? Seems like they are not quite right, at least in the test files, but just run that cleanup tool everywhere.

Once you have that, add Michael.

Yuri05 (approved):

PR #1425, opened 2024-07-19
1327 multiselect context menu for modules

### LINKED ISSUE #1327: Multiselect context menu for modules
Select multiple modules -> right click. I think the following commands should be supported:

- Delete
- Save as pkml (?)

benjaperez1983:
Fixes #1327

# Description

This is for the deletion of multiple selected modules with the context menu.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -278,6 +280,22 @@ public void RemoveResultsFromSimulations(IReadOnlyList<DataRepository> resultsTo
...
+         {
+            Description = AppConstants.Commands.RemoveMultipleModules,
+            ObjectType = AppConstants.MoBiObjectTypes.Module,
+            CommandType = AppConstants.Commands.DeleteCommand
+         };
+
+         modulesToRemove.Each(x => macroCommand.Add(new RemoveModuleCommand(x)));
+         _context.AddToHistory(macroCommand.Run(_context));
+      }
+
```
rwmcintosh, line 298: This is in the wrong task. There should be a task just like this for modules.

src/MoBi.Presentation/UICommand/RemoveMultipleModulesUICommand.cs
```diff
@@ -0,0 +1,23 @@
...
+using OSPSuite.Core.Domain.Data;
+using OSPSuite.Presentation.UICommands;
+
+namespace MoBi.Presentation.UICommand
+{
+   public class RemoveMultipleModulesUICommand : ObjectUICommand<IReadOnlyList<Module>>
+   {
+      private readonly IObservedDataTask _observedDataTask;
+
+      public RemoveMultipleModulesUICommand(IObservedDataTask observedDataTask)
```
rwmcintosh, line 13: Here you inject the correct task. Probably `IInteractionTasksForModule`

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -278,6 +280,22 @@ public void RemoveResultsFromSimulations(IReadOnlyList<DataRepository> resultsTo
...
+            return;
+
+         var macroCommand = new MoBiMacroCommand
+         {
+            Description = AppConstants.Commands.RemoveMultipleModules,
+            ObjectType = AppConstants.MoBiObjectTypes.Module,
+            CommandType = AppConstants.Commands.DeleteCommand
+         };
+
+         modulesToRemove.Each(x => macroCommand.Add(new RemoveModuleCommand(x)));
```
rwmcintosh, line 295: When you move this code to `IInteractionTasksForModule`, you can use `GetRemoveCommand` to create the commands.

rwmcintosh (changes requested):

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
benjaperez1983: hmm this is strange.
As now, I`m creating the command like this:

But I can see the "GetRemoveCommand" has 3 parameters:
GetRemoveCommand(Module objectToRemove, MoBiProject parent, IBuildingBlock buildingBlock)
2 of them which I do not have at this stage.
Checking the implementetion, I can see it is ignoring the 2 extra parameters:
public override IMoBiCommand GetRemoveCommand(Module objectToRemove, MoBiProject parent, IBuildingBlock buildingBlock) => new RemoveModuleCommand(objectToRemove);

Would it be ok to  use this setting null to the 2nd and 3rd parameter?

Let me know!

rwmcintosh: Yes, you can use `null` in these cases. This is written this way in the inherited class `InteractionTasksForChildren<TParent, TChild, TEditTask>`. 

Let's change these two to optional parameters in that base class. There are other cases where they are not required also. And we won't need to keep using `null`

rwmcintosh: > Let's change these two to optional parameters in that base class. There are other cases where they are not required also. And we won't need to keep using `null`

Actually, let's not do that right now - there's so much additional changes required that it will just make too much noise in the PR.

We can do that separately in another PR.

rwmcintosh (changes requested):

src/MoBi.Presentation/Tasks/Interaction/InteractionTask.cs
```diff
@@ -124,5 +127,17 @@ private void updateFormulaCacheOfClone(IBuildingBlock originBuildingBlock, IBuil
       public bool CorrectName<T>(T objectBase, IEnumerable<string> forbiddenNames) where T : IObjectBase => _nameCorrector.CorrectName(forbiddenNames, objectBase);
 
       public string PromptForNewName<T>(T objectBase, IEnumerable<string> forbiddenNames) where T : IObjectBase => _nameCorrector.PromptForCorrectName(forbiddenNames.ToList(), objectBase);
+
+      public void SaveMultiple<T>(IReadOnlyList<T> entitiesToSerialize)
```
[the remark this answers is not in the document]
benjaperez1983: Yes we could, but every time we consume this we will have to change the caller, creating a list and adding whatever object into it.
This seems to be easy since this "Save" function now has only 2 references, but this is being called from here:
EditTaskFor.cs:
image

Which as you can see on the image, it has 7 references.
So shall I go for it and change all the 7 callers as well? 
I`d rather not since it will create a lot of changes outside this scope. But let me know, happy to do it if you think it should be the way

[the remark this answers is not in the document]
rwmcintosh: Yeah, you could just override in InteractionTask, then you only have to create the enumerable in EditTaskFor, so don't change the signature in EditTaskFor, but do change in InteractionTask.

rwmcintosh (approved):

Yuri05 (approved):

PR #1428, opened 2024-07-22
Fixes #1396 improve table formula UX and add support for distributed table formula

### LINKED ISSUE #1395: Table formula: load values from excel
In MoBi:

- Add parameter -> Type "Table"
It is currently only possible to add values one by one:

image

- Add  an option to load table from excel
- reuse PK-Sim view:

image

### LINKED ISSUE #1396: Create new view and presenter for DistributedTableFormula
It should be similar to TableFormula view, with an additional column and the user should be able to select the distribution type

https://github.com/Open-Systems-Pharmacology/MoBi/issues/1040#issuecomment-1746645846

rwmcintosh:
Fixes #1396
Fixes #1395

# Description
Just enabling a TableFormulaPresenter to show a DistributedTableFormula

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


rwmcintosh:
If you check the issue that was created, the request is for the ability to edit the distribution type and the values from the distribution meta data.

I chatted with @msevestre about that and it seems that being able to edit this information will have no effect on the formula because the points are not going to be recalculated.

@PavelBal can you say whether you need this meta data for viewing only? It's a few extra columns in the table if so. All read-only.

PavelBal:
Yes, read only should suffice for now.

rwmcintosh:
> Yes, read only should suffice for now.

This PR as-is, gives you the same as for TableFormula without extra columns. If we need the additional data, then it will all be per-point.

- DistributionType,
- Mean,
- Deviation

Do you want all 3?

PavelBal:
For now, showing the same view as TableFormula should suffice, as the distribution information is not used in MOBi. Create a separate issue (for vNext) to extend this view to also show mean, deviation, and distribution type.

src/MoBi.Presentation/Tasks/TableFormulaTask.cs
```diff
@@ -64,5 +89,50 @@ public IMoBiCommand SetRestartSolver(TableFormula formula, ValuePoint valuePoint
...
+
+         return formula;
+      }
+
+      private TableFormula newTableFormula()
+      {
+         var tableFormula = _context.Create<TableFormula>();
+         tableFormula.XDimension = _context.DimensionFactory.Dimension(TIME);
+         tableFormula.XName = TIME;
+         tableFormula.Dimension = _context.DimensionFactory.Dimension(AppConstants.DimensionNames.FRACTION);
```
Yuri05, line 133: Why do we assume here that the X dimension is always Time and Y dimension is always fraction?
Isn't it better to set the X dimension to `baseGrid.Dimension` and Y dimension to `valueColumn.Dimension`?

And don't we need to set the formula name somewhere?

rwmcintosh: > Why do we assume here that the X dimension is always Time and Y dimension is always fraction? Isn't it better to set the X dimension to `baseGrid.Dimension` and Y dimension to `valueColumn.Dimension`?
> 
> And don't we need to set the formula name somewhere?

This tableformula is actually going to be thrown away. The interface for this is borrowed from PK-Sim where the formula in the parameter is swapped. In MoBi, we aren't going to change the formula,  we are only going to clear the points from the existing formula and add these new ones.

So, maybe we could change this in Core so the task only needs to return a data repository, then in PK-Sim the TableFormula is created, where in MoBi, I can just use the points from the DataRepository.

src/MoBi.Presentation/Tasks/TableFormulaTask.cs
```diff
@@ -64,5 +89,50 @@ public IMoBiCommand SetRestartSolver(TableFormula formula, ValuePoint valuePoint
...
+            dataImporterSettings,
+            _dialogCreator.AskForFileToOpen(Captions.Importer.OpenFile, Captions.Importer.ImportFileFilter, DirectoryKey.OBSERVED_DATA)
+         ).DataRepositories.FirstOrDefault();
+         return importedFormula == null ? null : formulaFrom(importedFormula);
+      }
+
+      private TableFormula formulaFrom(DataRepository dataRepository)
+      {
+         var baseGrid = dataRepository.BaseGrid;
+         var valueColumn = dataRepository.AllButBaseGrid().Single();
```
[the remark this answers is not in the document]
rwmcintosh: I will check on this to make sure.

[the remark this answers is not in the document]
rwmcintosh: This is now per construction

src/MoBi.Presentation/Tasks/TableFormulaTask.cs
rwmcintosh: Refactored so that we don't needlessly create the TableFormula. Instead, we use the Imported DataRepository

Yuri05 (approved):

PR #1446, opened 2024-07-31
#1367 better UI for add parameter value

### LINKED ISSUE #1367: Parameter Values BB -> Add parameter value: better UI
If the user wants to overwrite a default parameter value in the PV BB, he has to manually add this parameter as long it is not a protein expression or a local molecule parameter. The currently implemented workflow is not intuitive:

image

After clicking on "New Parameter Value," an empty row with the columns "Parameter Name" and "Parameter Value" is created. The user would have to manually add columns "Path Element n" to be able to enter the path.

A possible solution would be a dialog where the user can either select a parameter that is present somewhere (like in all the places where the user selects a parameter), or enters the whole path which is then separated by `|` to create an entry in the PV BB.

benjaperez1983:
Fixes #1367 

# Description

Change UI when creating new parameter value.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image
image
image

# Questions (if appropriate):


src/MoBi.Presentation/Presenter/SelectReferencePresenterBase.cs
```diff
@@ -246,7 +246,7 @@ protected void AddSpatialStructures()
          _view.AddNodes(nodes);
       }
 
-      private bool shouldCreateAbsolutePaths => _view.ObjectPathType.Equals(ObjectPathType.Absolute);
+      protected bool ShouldCreateAbsolutePaths => _view.ObjectPathType.Equals(ObjectPathType.Absolute);
```
msevestre, line 249: I feel this property should be singular `ShouldCreateAbsolutePath`

msevestre:
@benjaperez1983 Can you add screenshots of new UI so that I can see what has changed without having to download the code. Hard to tell otherwise

benjaperez1983:
> @benjaperez1983 Can you add screenshots of new UI so that I can see what has changed without having to download the code. Hard to tell otherwise

@msevestre Added, sorry!

src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
```diff
@@ -0,0 +1,60 @@
...
+         var selection = getSelected<IParameter>();
+
+         return ShouldCreateAbsolutePaths ? _objectPathFactory.CreateAbsoluteObjectPath(selection) : _objectPathFactory.CreateRelativeObjectPath(_refObject, selection);
+      }
+
+      protected override T getSelected<T>()
+      {
+         var dto = _view.SelectedDTO;
+         return dto == null ? null : _context.Get<T>(dto.ObjectBase.Id);
+      }
```
rwmcintosh, line 58: I see... This is about what we were discussing.

The base method does exactly this, except uses `dto.Id` instead of `dto.ObjectBase.Id`

I think we can change the behavior in the base class and remove this override. For some reason, there are some DTO's who do not have the identical Id to the object base, but, no matter what, we probably want to use the `ObjectBase.Id` when we are searching the context.

Let's test it tomorrow.

src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
```diff
@@ -0,0 +1,60 @@
...
+      public override bool CanClose => isSelectionParameterType();
+
+      private bool isSelectionParameterType() => getSelected<IParameter>() != null;
+
+      public override ObjectPath GetSelection()
+      {
+         var selection = getSelected<IParameter>();
+
+         return ShouldCreateAbsolutePaths ? _objectPathFactory.CreateAbsoluteObjectPath(selection) : _objectPathFactory.CreateRelativeObjectPath(_refObject, selection);
+      }
```
rwmcintosh, line 52: Maybe we can remove this override too. The `GetSelection` should only be called if the user was able to dismiss the dialog, which should only happen if an `IParameter` was selected.

src/MoBi.Presentation/Presenter/SelectReferencePresenterBase.cs
rwmcintosh, line 249: We can probably revert this back to private.

src/MoBi.UI/Views/ParameterValuesView.cs
```diff
@@ -77,6 +85,15 @@ private void setParameterUnit(ParameterValueDTO parameterValue, Unit unit)
          });
       }
 
+      private void onGridViewMouseDown(MouseEventArgs e)
+      {
+         if (e.Button != MouseButtons.Right) return;
+         var location = new Point(e.X, e.Y + 50);
```
rwmcintosh, line 91: We have to find the source of this `+ 50`. I did some digging around and came up with this little experiment. There are a configurable number of boxes that can be viewed here

image

In the ParameterValuesPresenter I arranged for them all to be shown by commenting out
```
         view.HideIsPresentView();
         view.HideRefreshView();
         view.HideNegativeValuesAllowedView();
```
I thought that the context menu would be shown in the right spot, I clicked in the very top/left corner of the grid and I saw that the problem was actually twice as bad now (top image none of the controls hidden, bottom with all but one hidden)!

image

So, at least we know what is causing it. Can you do some more digging tomorrow to see what this could be? This will probably mean digging through some DevExpress documentation.

We can also discuss tomorrow.

rwmcintosh (changes requested):

src/MoBi.Presentation/ModalPresenter.cs
```diff
@@ -13,13 +13,19 @@ public interface IModalPresenter : IDisposablePresenter
...
    internal class ModalPresenter : AbstractDisposablePresenter<IContainerModalView, IModalPresenter>, IModalPresenter
    {
       private readonly IEventPublisher _eventPublisher;
       public IPresenter SubPresenter { get; protected set; }
 
+      public bool InitialOkEnabled
+      {
+         set => _view.OkEnabled = value;
+      }
+
```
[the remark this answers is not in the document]
benjaperez1983: yes but will this apply for every case? meaning what if we consume this modal form from some other place which does not need this behavior?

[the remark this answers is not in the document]
benjaperez1983: Forget it, one one of the further comments you already suggested to revert the encpasulate and init, and that solves it.
Removed! Great catch

src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
benjaperez1983: Great, I`ve solved the coment on the code, but will leave itas unsolved since I need to know how to properly test it other than for this particular case.

src/MoBi.UI/Views/ParameterValuesView.cs
benjaperez1983: I think I know why this is happening.
There is a control of type: 
private DevExpress.XtraLayout.EmptySpaceItem emptySpaceItem;
which has a height of 50px:
this.emptySpaceItem.Size = new System.Drawing.Size(849, 50);

if I set this control to not visible with the method:
   public void HideSubPresenterGrouping()
   {
      emptySpaceItem.Visibility = LayoutVisibility.Never;
      layoutGroupPanel.GroupBordersVisible = false;
   }
then the menu is showing fine. Do you think this should be the solution?

src/MoBi.Presentation/Presenter/ParameterValuesPresenter.cs
```diff
@@ -80,5 +94,30 @@ public void UpdateDimension(ParameterValueDTO valueObject, IDimension newDimensi
...
+
+      public void AddNewEmptyPathAndValueEntity(ObjectPath entityPath)
+      {
+         var newParameterValue = _emptyStartValueCreator.CreateEmptyStartValue(_interactionTasksForExtendablePathAndValueEntity.GetDefaultDimension());
+         newParameterValue.Path = entityPath;
+         var newRecord = _valueMapper.MapFrom(newParameterValue, _buildingBlock);
+
+         _startValueDTOs.Insert(0, newRecord);
+         BindToView();
+      }
```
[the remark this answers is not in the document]
benjaperez1983: Almost done; which method do you mean by "from the method that's there where the path is not set." exactly?

src/MoBi.UI/Views/ParameterValuesView.cs
rwmcintosh: Does that hide the `Delete Values` controls?

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForContainer.cs
```diff
@@ -108,8 +112,11 @@ protected override void AddSaveItems(IContainer container)
          _allMenuItems.Add(CreateMenuButton.WithCaption(AppConstants.MenuNames.SaveAsPKML.WithEllipsis())
             .WithCommandFor<SaveWithIndividualAndExpressionUICommand, IContainer>(container, _container)
             .WithIcon(ApplicationIcons.PKMLSave));
-      }
 
+         _allMenuItems.Add(CreateMenuButton.WithCaption(AppConstants.Captions.CopyPath)
+            .WithActionCommand(() => Clipboard.SetText(_entityPathResolver.FullPathFor(container as IEntity)))
```
rwmcintosh, line 117: I don't think you need `as IEntity` in here. Can you check it?

src/MoBi.Presentation/Presenter/ExtendablePathAndValueBuildingBlockPresenter.cs
```diff
@@ -33,9 +33,9 @@ public abstract class ExtendablePathAndValueBuildingBlockPresenter<TView, TPrese
    {
       protected readonly IPathAndValueEntityToPathAndValueEntityDTOMapper<TPathAndValueEntity, TStartValueDTO> _valueMapper;
 
-      private readonly IInteractionTasksForExtendablePathAndValueEntity<TBuildingBlock, TPathAndValueEntity> _interactionTasksForExtendablePathAndValueEntity;
+      protected readonly IInteractionTasksForExtendablePathAndValueEntity<TBuildingBlock, TPathAndValueEntity> _interactionTasksForExtendablePathAndValueEntity;
       protected BindingList<TStartValueDTO> _startValueDTOs;
-      private readonly IEmptyStartValueCreator<TPathAndValueEntity> _emptyStartValueCreator;
+      protected readonly IEmptyStartValueCreator<TPathAndValueEntity> _emptyStartValueCreator;
```
rwmcintosh, line 38: Can these two go back to private since you pulled up the code for `AddNewPathAndValueEntity`?

src/MoBi.Presentation/Presenter/ExtendablePathAndValueBuildingBlockPresenter.cs
```diff
@@ -274,5 +265,22 @@ public void Handle(BulkUpdateStartedEvent eventToHandle)
          _focusedStartValue = PathAndValueEntityFrom(_view.FocusedStartValue);
          _handleChangedEvents = false;
       }
+
+      public void AddNewPathAndValueEntity(ObjectPath entityPath) => addNewPathAndValue(entityPath);
+
+      public void AddNewEmptyPathAndValueEntity() => addNewPathAndValue(null);
+
+      private void addNewPathAndValue(ObjectPath entityPath)
```
rwmcintosh, line 273: If you make `entityPath` optional then you can call this on line 271 without the null. Just a bit tidier.

src/MoBi.Presentation/Presenter/ParameterValuesPresenter.cs
```diff
@@ -80,5 +92,22 @@ public void UpdateDimension(ParameterValueDTO valueObject, IDimension newDimensi
...
          AddCommand(macroCommand);
       }
+
+      public void ShowContextMenu(IViewItem objectRequestingPopup, Point popupLocation) =>
+         _viewItemContextMenuFactory.CreateFor(objectRequestingPopup, this).Show(_view, popupLocation);
+
+      public void AddNewParameterValue()
+      {
+         _modalPresenter.Text = AppConstants.Captions.SelectLocalReferencePoint;
+         //The next 2 lines should not be changed, as they are used to encapsulate the presenter
```
rwmcintosh, line 102: can you just state that the 'order of the next two lines' should not be changed instead of the 'lines should not be changed'

src/MoBi.Presentation/Presenter/ParameterValuesPresenter.cs
```diff
@@ -80,5 +92,22 @@ public void UpdateDimension(ParameterValueDTO valueObject, IDimension newDimensi
 
          AddCommand(macroCommand);
       }
+
+      public void ShowContextMenu(IViewItem objectRequestingPopup, Point popupLocation) =>
+         _viewItemContextMenuFactory.CreateFor(objectRequestingPopup, this).Show(_view, popupLocation);
+
+      public void AddNewParameterValue()
+      {
+         _modalPresenter.Text = AppConstants.Captions.SelectLocalReferencePoint;
```
rwmcintosh, line 101: I think this caption can change too. We are selecting a parameter, so it can probably be something simple like that.

src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
```diff
@@ -0,0 +1,47 @@
...
+
+      public override void SelectionChanged(ITreeNode treeNode)
+      {
+         OnStatusChanged(null, null);
+      }
+
+      public override bool CanClose => isSelectionParameterType();
+
+      private bool isSelectionParameterType() => getSelected<IParameter>() != null;
```
rwmcintosh, line 45: beauty.

src/MoBi.Presentation/Presenter/SelectReferencePresenterBase.cs
```diff
@@ -167,28 +168,27 @@ private ReferenceDTO timeReference()
...
+         return shouldCreateAbsolutePath ? _objectPathFactory.CreateAbsoluteObjectPath(selection) : _objectPathFactory.CreateRelativeObjectPath(_refObject, selection);
       }
 
-      private T getSelected<T>() where T : class, IObjectBase
+      protected T getSelected<T>() where T : class, IObjectBase
       {
          var dto = _view.SelectedDTO;
+
          return dto == null ? null : _context.Get<T>(dto.Id);
       }
```
rwmcintosh, line 187: `GetSelection` can be non-virtual I think since we changed the DTO behavior. If `getSelected<T>` needs to stay protected then the name has to be capitalized.

src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
```diff
@@ -0,0 +1,47 @@
...
+      protected override void AddSpecificInitialObjects()
+      {
+         AddSpatialStructures();
+
+         _view.ChangeLocalisationAllowed = false;
+      }
+
+      public override void SelectionChanged(ITreeNode treeNode)
+      {
+         OnStatusChanged(null, null);
```
rwmcintosh, line 40: should you call base.SelectionChanged too?

rwmcintosh:
Needs testing added

rwmcintosh (approved):

Yuri05 (approved):

PR #1452, opened 2024-08-01
#1437  fix possible null reference

### LINKED ISSUE #1437: Exception when creating an new IC BB
[TwoContainer_V2.zip](https://github.com/user-attachments/files/16377144/TwoContainer_V2.zip)

In the attached project, open the module "Pysical containers - Molecule A", right click on "Initial Conditions" and select "Create..."

image

Exception:

Application:
MoBi® 12.0.291

Stack trace:
```
at MoBi.Presentation.Presenter.AddContentToModulePresenter`2.AddMoleculesSelectionChanged(Boolean moleculesSelected)
   at MoBi.UI.Views.AddContentToModuleView`2.moleculesChanged()
   at OSPSuite.UI.Views.BaseView.<>c__DisplayClass25_0.<OnEvent>b__0()
   at OSPSuite.Utility.Extensions.ExceptionExtensions.DoWithinExceptionHandler(Object callerObject, Action actionToExecute)
```

benjaperez1983:
Fixes #1437 

# Description

Fix a null referenced object when creating a new Initial Condition

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/AddContentToModulePresenter.cs
+++ b/src/MoBi.Presentation/Presenter/AddContentToModulePresenter.cs
@@ -23,7 +23,7 @@ namespace MoBi.Presentation.Presenter
 
       public void AddMoleculesSelectionChanged(bool moleculesSelected)
       {
-         if (!moleculesSelected || Module.InitialConditionsCollection.Any())
+         if (!moleculesSelected || (Module?.InitialConditionsCollection != null && Module.InitialConditionsCollection.Any()))
             return;
 
          ContentDTO.WithInitialConditions = true;
--- a/src/MoBi.Presentation/Presenter/AddContentToModulePresenter.cs
+++ b/src/MoBi.Presentation/Presenter/AddContentToModulePresenter.cs
@@ -23,7 +23,7 @@ namespace MoBi.Presentation.Presenter
 
       public void AddMoleculesSelectionChanged(bool moleculesSelected)
       {
-         if (!moleculesSelected || (Module?.InitialConditionsCollection != null && Module.InitialConditionsCollection.Any()))
+         if (!moleculesSelected || Module.InitialConditionsCollection.Any())
             return;
 
          ContentDTO.WithInitialConditions = true;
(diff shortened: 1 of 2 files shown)


src/MoBi.Presentation/Presenter/AddContentToModulePresenter.cs
```diff
@@ -23,7 +23,7 @@ protected AddContentToModulePresenter(TView view) : base(view)
 
       public void AddMoleculesSelectionChanged(bool moleculesSelected)
       {
-         if (!moleculesSelected || Module.InitialConditionsCollection.Any())
+         if (!moleculesSelected || (Module?.InitialConditionsCollection != null && Module.InitialConditionsCollection.Any()))
```
rwmcintosh, line 26: I think there is an earlier issue than this. By the time this get's called, the presenter should already be initialized and that initialization should always include the module to which we are adding initial conditions.

Can you check what happens when you left click, then select "Add Building Block" the pick "Initial Conditions" I think the Module will be non-null.

rwmcintosh (changes requested):

src/MoBi.Presentation/Presenter/AddContentToModulePresenter.cs
benjaperez1983: Yeah, I`ve made some analysis before submitting.
As you mention, the only place where we initialize this variable is on the "AddBuildingBlocksToModule" method:
image

Other than that, the Module property is null.
What I think (might be mistaken of course), is that this module variable is being set only on this public method, instead of the private method (with the same name) for a reason, but checking references, I can see that this property is only being used for this case mentioned on the ticket, so it might be safe to move it into the private method.

Let me know your thoughts on this.

rwmcintosh (approved):

Yuri05 (approved):

PR #1482, opened 2024-08-13
1443 create proccess rate property should not change a pk sim module to an extension module

### LINKED ISSUE #1443: "Create proccess rate" property should not change a PK-Sim module to an extension module
E.g. in a passive transport "Create process rate".

benjaperez1983:
Fixes #1443 

# Description

When changing Create process rate value, should not convert to extension module, but mark the simulation as not in sync.

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/MoBi.Core/Services/SimulationEventsOnlyBuildingBlockVersionUpdater.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+   public interface ISimulationEventsOnlyBuildingBlockVersionUpdater
+   {
+      void UpdateBuildingBlockVersion(IBuildingBlock buildingBlock, uint newVersion);
+      void UpdateBuildingBlockVersion(IBuildingBlock buildingBlock, bool shouldIncrementVersion);
+   }
+
+   public class SimulationEventsOnlyBuildingBlockVersionUpdater : BuildingBlockVersionUpdater, ISimulationEventsOnlyBuildingBlockVersionUpdater
+   {
+      public SimulationEventsOnlyBuildingBlockVersionUpdater(IMoBiProjectRetriever projectRetriever, IEventPublisher eventPublisher, IDialogCreator dialogCreator) : base(projectRetriever, eventPublisher, dialogCreator)
```
msevestre, line 15: I don't understand the name of this class. 
SimulationEvents?

rwmcintosh (changes requested):
We need to change the approach as discussed

src/MoBi.Core/Commands/BuildingBlockIrreversibleChangeCommandBase.cs
rwmcintosh, line 1: Let's remove this class. No one is implementing it

src/MoBi.Core/Commands/SetCreateProcessRateParameterCommand.cs
```diff
@@ -27,6 +27,7 @@ public SetCreateProcessRateParameterCommand(bool createProcessRate, IProcessBuil
 
       protected override void ExecuteWith(IMoBiContext context)
       {
+         ShouldConvertPKSimModule = false;
```
rwmcintosh, line 30: I would just do this in the constructor as part of the command initialization.

src/MoBi.Core/Services/BuildingBlockVersionUpdater.cs
```diff
@@ -29,15 +29,19 @@ public BuildingBlockVersionUpdater(IMoBiProjectRetriever projectRetriever, IEven
          _dialogCreator = dialogCreator;
       }
 
-      public void UpdateBuildingBlockVersion(IBuildingBlock buildingBlock, uint newVersion)
+      public virtual void UpdateBuildingBlockVersion(IBuildingBlock buildingBlock, uint newVersion, bool shouldConvertPKSimModule)
```
rwmcintosh, line 32: Does not need to be virtual with this approach

src/MoBi.Core/Services/BuildingBlockVersionUpdater.cs
```diff
@@ -29,15 +29,19 @@ public BuildingBlockVersionUpdater(IMoBiProjectRetriever projectRetriever, IEven
...
       {
          if (buildingBlock == null)
             return;
 
-         if (shouldShowModuleConversionWarning(buildingBlock))
+         if (shouldConvertPKSimModule && shouldShowModuleConversionWarning(buildingBlock))
```
rwmcintosh, line 37: I would move this bool check into the `shouldShow` method. Then I would take line 43 and move it inside this condition. The problem with line 42/43 is that not all building blocks belong to a Module. This method `shouldShow` also checks for that.

Basically, the conditions for showing the warning are the same for the conditions to convert the module, so let's only check it once.

src/MoBi.Core/Services/BuildingBlockVersionUpdater.cs
```diff
@@ -76,10 +80,10 @@ public void UpdateBuildingBlockVersion(IBuildingBlock buildingBlock, bool should
          else
             version--;
 
-         UpdateBuildingBlockVersion(buildingBlock, version);
+         UpdateBuildingBlockVersion(buildingBlock, version, shouldConvertPKSimModule);
       }
 
-      private void publishSimulationStatusChangedEvents(IBuildingBlock changedBuildingBlock)
+      protected void publishSimulationStatusChangedEvents(IBuildingBlock changedBuildingBlock)
```
rwmcintosh, line 86: that can go back to private

tests/MoBi.Tests/Core/Commands/SetCreateProcessRateParameterCommandSpecs.cs
```diff
@@ -41,5 +46,11 @@ public void should_set_the_plot_process_rate_parameter_to_false_if_the_create_pr
       {
          _reaactionBuilder.ProcessRateParameterPersistable.ShouldBeFalse();
       }
+
+      [Observation]
+      public void the_module_is_not_pk_sim_module()
```
rwmcintosh, line 51: This should be called the_module_remains_a_pk_sim_module

rwmcintosh (changes requested):

src/MoBi.Core/Services/BuildingBlockVersionUpdater.cs
```diff
@@ -51,9 +48,15 @@ private void showModuleConversionWarning(IBuildingBlock buildingBlock)
          _dialogCreator.MessageBoxInfo(AppConstants.Captions.TheModuleWillBeConvertedFromPKSimToExtensionModule(buildingBlock.Module.Name));
       }
 
-      private bool shouldShowModuleConversionWarning(IBuildingBlock buildingBlock)
+      private bool shouldShowModuleConversionWarning(IBuildingBlock buildingBlock, bool shouldConvertPKSimModule)
       {
-         return buildingBlock.Module != null && buildingBlock.Module.IsPKSimModule;
+         if (buildingBlock.Module?.IsPKSimModule == true && shouldConvertPKSimModule == true)
+         {
+            buildingBlock.Module.IsPKSimModule = false;
```
rwmcintosh, line 55: I would not do this quite this way, the semantics of the name `shouldShowModuleConversionWarning` does not convey that anything will be changed as a result of calling the method. Instead, I would move this up one level in the call stack so you'd have

```
         if (shouldShowModuleConversionWarning(buildingBlock, shouldConvertPKSimModule))
         {
            showModuleConversionWarning(buildingBlock);
            buildingBlock.Module.IsPKSimModule = false;
         }
```

It probably makes sense then to rename the method too

```

I think that would make it easier to read.
shouldShowModuleConversionWarning -> shouldConvertModule
```

Then it reads

```
         if (shouldConvertModule(buildingBlock, shouldConvertPKSimModule))
         {
            showModuleConversionWarning(buildingBlock);
            buildingBlock.Module.IsPKSimModule = false;
         }
```

src/MoBi.Core/MoBi.Core.csproj
```diff
@@ -33,7 +33,7 @@
   <ItemGroup>
     <PackageReference Include="FluentNHibernate" Version="2.1.2" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
-    <PackageReference Include="OSPSuite.Utility" Version="4.1.0.5" />
+    <PackageReference Include="OSPSuite.Utility" Version="4.1.0.6" />
```
rwmcintosh, line 36: Can you roll this back

src/MoBi.Core/Services/BuildingBlockVersionUpdater.cs
```diff
@@ -47,12 +51,10 @@ private void showModuleConversionWarning(IBuildingBlock buildingBlock)
          _dialogCreator.MessageBoxInfo(AppConstants.Captions.TheModuleWillBeConvertedFromPKSimToExtensionModule(buildingBlock.Module.Name));
       }
 
-      private bool shouldShowModuleConversionWarning(IBuildingBlock buildingBlock)
-      {
-         return buildingBlock.Module != null && buildingBlock.Module.IsPKSimModule;
-      }
+      private bool shouldShowModuleConversionWarning(IBuildingBlock buildingBlock) =>
```
rwmcintosh, line 54: can you rename this method to `shouldConvertModule`

src/MoBi.Presentation/MoBi.Presentation.csproj
```diff
@@ -27,7 +27,7 @@
   <ItemGroup>
     <PackageReference Include="Northwoods.GoWin" Version="5.2.0" />
     <PackageReference Include="OSPSuite.TeXReporting" Version="3.0.0.5" />
-    <PackageReference Include="OSPSuite.Utility" Version="4.1.0.5" />
+    <PackageReference Include="OSPSuite.Utility" Version="4.1.0.6" />
```
rwmcintosh, line 30: roll back all these Utility ok?

rwmcintosh (approved):

src/MoBi.Core/Commands/BuildingBlockChangeCommandBase.cs
```diff
@@ -23,7 +25,7 @@ protected override void ExecuteWith(IMoBiContext context)
       {
          if (_buildingBlock == null) return;
          var buildingBlockVersionUpdater = context.Resolve<IBuildingBlockVersionUpdater>();
-         buildingBlockVersionUpdater.UpdateBuildingBlockVersion(_buildingBlock, ShouldIncrementVersion);
+         buildingBlockVersionUpdater.UpdateBuildingBlockVersion(_buildingBlock, ShouldIncrementVersion, ConversionOption);
```
Yuri05, line 28: I think the behaviour ist still not fully correct.
E.g. if I have a command with `ConversionOption=SetAsExtensionModule` and apply it to a module which is already an ExtensionModule - nothing will change, so the inverse command should become `ConversionOption=None` - but it will become `ConversionOption=SetAsPKSimModule`

`buildingBlockVersionUpdater.UpdateBuildingBlockVersion` should return if the module state was changed or not.
And if not: current command should become `ConversionOption=None`

src/MoBi.Core/Commands/BuildingBlockChangeBaseCommandExtensions.cs
```diff
@@ -9,7 +10,24 @@ public static BuildingBlockChangeCommandBase<T> AsInverseFor<T>(this BuildingBlo
       {
          CommandExtensions.AsInverseFor(inverseCommand, originalCommand);
          inverseCommand.ShouldIncrementVersion = !originalCommand.ShouldIncrementVersion;
+         inverseCommand.ConversionOption = getReverseConversionOption(originalCommand.ConversionOption, originalCommand.HasChangedModuleType);
          return inverseCommand;
       }
+
+      private static PKSimModuleConversion getReverseConversionOption(PKSimModuleConversion conversionOption, bool hastChangedModuleType)
```
Yuri05, line 17: typo: _has**t**ChangedModuleType_

src/MoBi.Core/Commands/BuildingBlockChangeCommandBase.cs
```diff
@@ -21,9 +24,11 @@ protected BuildingBlockChangeCommandBase(T buildingBlock)
 
       protected override void ExecuteWith(IMoBiContext context)
       {
+         var originalPkSimModuleState = _buildingBlock.Module?.IsPKSimModule ?? false;
```
Yuri05, line 27: `buildingBlock.Module?.IsPKSimModule ?? false` is used 2 times here and in `shouldConvertToExtensionModule()` function below.
Would encapsulate it in a property of `IBuildingBlock`

rwmcintosh: If it goes an extension of Module then we can probably hide all the null operators in the extension method.

Yuri05: or like this, yes

benjaperez1983: If we do it as an extension of the MODULE instead of BUILDINGBLOCK, it maybe confusing. 
Since the Module has the property IsPkSimModule, therefore, adding a method which would return the same value (similar name and information) may be confusing, would you agree? 
I added it at a building blokc level, but of course, feel free to point out whatever you think about it.

Yuri05 (approved):

src/MoBi.Core/Commands/BuildingBlockChangeCommandBase.cs
Yuri05: Agree.

src/MoBi.Core/Domain/Extensions/BuildingBlockExtensions.cs
```diff
@@ -71,6 +71,9 @@ private static bool modulesShareName(Module module1, Module module2)
       {
          return module1 != null && module2.IsNamed(module1.Name);
       }
+
+      public static bool IsPkSimModule(this IBuildingBlock buildingBlock) =>
+         buildingBlock.Module?.IsPKSimModule ?? false;
```
Yuri05, line 76: somehow line 76 crashes in several tests. probably should check if `buildingBlock `is not null as well

Yuri05 (approved):

PR #1536, opened 2024-09-06
Fixes #1517 It should be possible to create different formula types in the PV BB (and IC BB?)

### LINKED ISSUE #1517: It should be possible to create different formula types in the PV BB (and IC BB?)
Currently, when creating a new formula in the PV BB:

image

Only "explicit formulas" can be created. Creation of other types of formulas should be supported:

- Table
- Table with offset
- Table with X-Argument
- Sum formula

rwmcintosh:
Fixes #1517

# Description
Refactor the `EditFormulaPresenter` to accommodate a slightly different workflow in `PathAndValueBuildingBlock`. 

Previously the popup only allowed you to create/edit an explicit formula, now we are using the usual formula editor from formulas in containers. 

Mostly the changes have to do with the fact that this is in a modal dialog. The expectation is that if the dialog is canceled, the commands are not executed. In this case, rather than refactor all formula presenters to not execute commands, I have implemented the commands to be rolled back when changing formula types.

If you change the formula type a number of times, not all formulas are going to be added to the formula cache, only the formula being edited when user presses OK.

Also, if cancel is pressed, all commands are rolled back.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):


src/MoBi.Presentation/Presenter/EditFormulaInPathAndValuesPresenter.cs
rwmcintosh, line 1: This is the basis of the real changes here. We need a slightly different workflow when editing formula in PathAndValues building blocks.

src/MoBi.UI/Views/CreateFormulaView.cs
```diff
@@ -0,0 +1,34 @@
...
+         layoutItemFormulaSelect.Visibility = LayoutVisibility.Never;
+         layoutItemAddFormula.Visibility = LayoutVisibility.Never;
+         layoutItemCloneFormula.Visibility = LayoutVisibility.Never;
+         cbFormulaName.Properties.Buttons.Clear();
+         tbFormulaName.Enabled = false;
+      }
+
+      protected override void InitializeNameBinding() =>
+         _screenBinder.Bind(x => x.Name)
+            .To(tbFormulaName);
```
rwmcintosh, line 32: Instead of binding to the combo box like in the container style editor, we just bind to a text box, which is disabled.

src/MoBi.Presentation/Presenter/EditFormulaInPathAndValuesPresenter.cs
```diff
@@ -0,0 +1,70 @@
...
+         SelectFormula(newFormula);
+         UpdateFormula();
+      }
+
+      private void rollBackChanges()
+      {
+         if (_localMacroCommand.IsEmpty) 
+            return;
+
+         _interactionTaskContext.CancelCommand(_localMacroCommand);
```
msevestre, line 58: I Wonder how well this works...feels dangerous to me lol. Can't we just swap out the formula?

src/MoBi.Presentation/Presenter/EditTableFormulaWithOffSetFormulaPresenter.cs
```diff
@@ -87,15 +89,18 @@ private bool isValidTableObject(IObjectBase objectBase)
...
+            var referencePresenter = _selectReferencePresenterFactory.ReferenceAtParameterFor(UsingObject?.ParentContainer);
             referencePresenter.DisableTimeSelection();
-            presenter.Init(predicate, UsingObject, new[] {UsingObject.RootContainer}, caption, referencePresenter);
+            presenter.Init(predicate, UsingObject, contextList(UsingObject), caption, referencePresenter);
             return presenter.GetSelection();
          }
       }
+
+      private IReadOnlyList<IObjectBase> contextList(IUsingFormula usingObject) =>
+         usingObject?.RootContainer != null ? new List<IObjectBase> { usingObject.RootContainer } : Enumerable.Empty<IObjectBase>().ToList();
```
msevestre, line 104: there is an Array.Empty or something lile this.

src/MoBi.Presentation/Presenter/EditTableFormulaWithXArgumentFormulaPresenter.cs
```diff
@@ -91,9 +91,12 @@ private FormulaUsablePath selectFormulaUsablePath(Func<IObjectBase, bool> predic
...
             var referencePresenter = _selectReferencePresenterFactory.ReferenceAtParameterFor(UsingObject?.ParentContainer);
             referencePresenter.DisableTimeSelection();
-            presenter.Init(predicate, UsingObject, UsingObject == null ? Enumerable.Empty<IObjectBase>().ToList() : new List<IObjectBase> {UsingObject.RootContainer}, caption, referencePresenter);
+            presenter.Init(predicate, UsingObject, contextList(UsingObject), caption, referencePresenter);
             return presenter.GetSelection();
          }
       }
+
+      private IReadOnlyList<IObjectBase> contextList(IUsingFormula usingObject) =>
+         usingObject?.RootContainer != null ? new List<IObjectBase> { usingObject.RootContainer } : Enumerable.Empty<IObjectBase>().ToList();
```
msevestre, line 100: same

msevestre (commented):
Looks good to me
the macro command undo while editing surprises me. Why do we need this?

src/MoBi.Presentation/Presenter/EditFormulaInPathAndValuesPresenter.cs
rwmcintosh: Let me see how this can be minimally refactored for it. The base presenter uses the formulaOwner and buildingBlock for some stuff. Maybe we can get around it easily

rwmcintosh: All the underlying specific formula type editors perform their actions using commands, implying that the intention is to operate directly on the formula while it's in-place.

Rather than refactor all that, I have done this by cloning the building block and operating on the cloned building block formula cache and the cloned formula owner.

Then those commands are just thrown away and we use an update formula command and an add to cache command to make the change.

msevestre: yeah good idea. 
I just did not expect the undo to work as smooth as we expected it...

msevestre (approved):

src/MoBi.Presentation/Presenter/EditFormulaInPathAndValuesPresenter.cs
rwmcintosh: It did work, but I was uneasy about it. Since you also had reservations, I revised it.

PR #1546, opened 2024-09-18
#1527 add warning for parameters value dimension

### LINKED ISSUE #1527: PV BB: New parameter value - Notification for dimension check
Now, when adding new entries through the "New Parameter Value" button, all entries have the dimension "Dimensionless". Instead, they should have the dimension of the parameter selected from the tree view.

benjaperez1983:
Fixes #1527

# Description

 add warning for parameters value dimension

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1183,7 +1183,6 @@ namespace MoBi.Assets
             return string.Format("Cannot find a parameter for path {1} in simulation {0} during import. Only updates are allowed", name, path);
          }
 
-
          public static readonly string CheckDimensionOfAddedParameters = "Check the \"Dimension\" property of the added parameters";
          public static readonly string ThisItNotATemplateBuildingBlock = "This is not the template building block!";
       }
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1183,7 +1183,7 @@ namespace MoBi.Assets
             return string.Format("Cannot find a parameter for path {1} in simulation {0} during import. Only updates are allowed", name, path);
          }
 
-         public static readonly string CheckDimensionOfAddedParameters = "Check the \"Dimension\" property of the added parameters";
+         public static readonly string CheckDimensionOfAddedParameters = "Check the <b><i>Dimension</i></b> property of the added parameters";
          public static readonly string ThisItNotATemplateBuildingBlock = "This is not the template building block!";
       }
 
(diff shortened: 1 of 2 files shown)


src/MoBi.Presentation/Presenter/ParameterValuesPresenter.cs
```diff
@@ -119,6 +119,8 @@ private IMoBiCommand addParameterValuesForObjectPaths(IReadOnlyList<ObjectPath>
          if (allSkipped.Any())
             _dialogCreator.MessageBoxInfo(AppConstants.Captions.BuildingBlockAlreadyContains(allSkipped.Select(x => x.PathAsString).ToList()));
 
+         _dialogCreator.MessageBoxInfo(AppConstants.Warnings.CheckDimensionOfAddedParameters);
```
rwmcintosh, line 122: Could it happen that there are no parameter values added? If so, I'd skip the dialog.

Try to select nothing in the selection dialog, or select one, add it, then add the same one again. It will be skipped.

If nothing gets added to the BB let's skip the notifications

src/MoBi.Assets/AppConstants.cs
```diff
@@ -1183,6 +1183,7 @@ public static string CannotAddNewParameterFromImportToSimulation(string name, st
             return string.Format("Cannot find a parameter for path {1} in simulation {0} during import. Only updates are allowed", name, path);
          }
 
+         public static readonly string CheckDimensionOfAddedParameters = "Check the \"Dimension\" property of the added parameters";
```
rwmcintosh, line 1186: Check if this can be italics instead of quotation marks.

src/MoBi.Presentation/Presenter/ParameterValuesPresenter.cs
```diff
@@ -119,7 +119,8 @@ private IMoBiCommand addParameterValuesForObjectPaths(IReadOnlyList<ObjectPath>
          if (allSkipped.Any())
             _dialogCreator.MessageBoxInfo(AppConstants.Captions.BuildingBlockAlreadyContains(allSkipped.Select(x => x.PathAsString).ToList()));
 
-         _dialogCreator.MessageBoxInfo(AppConstants.Warnings.CheckDimensionOfAddedParameters);
+         if(allSkipped.Any() || objectPathsToAdd.Any())
```
rwmcintosh, line 122: This will mean that if we skip all paths the prompt will still appear, but I think it should be prompt only if any object paths are added.

If all paths are skipped, then no prompt.

So you only need `if(objectPathsToAdd.Any())`

rwmcintosh (changes requested):

src/MoBi.Presentation/Presenter/ParameterValuesPresenter.cs
rwmcintosh: oh, could you also remove the bold, I think italic only is the normal convention

Yuri05 (approved):

PR #1573, opened 2024-10-07
#1554 rename module

### LINKED ISSUE #1554: Rename module when renaming imported simulation
When importing a simulation from pkml, whose name already exists in the project: MoBi asks to rename the simulation and suggests `<old simulation name> N` as the new default name:

grafik

However, if I enter another name for the simulation: the module of the imported simulation still has the default name 
`<old simulation name> N`:

grafik grafik

Would be nice to rename the module parallel to the simulation

benjaperez1983:
Fixes #1554 

# Description

 When importing a simulation from pkml, whose name already exists in the project: MoBi asks to rename the simulation and suggests <old simulation name> N as the new default name:
However, if I enter another name for the simulation: the module of the imported simulation still has the default name

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Core/Services/SimulationLoader.cs
+++ b/src/MoBi.Core/Services/SimulationLoader.cs
@@ -56,12 +56,13 @@ namespace MoBi.Core.Services
          var moBiSimulation = simulation;
 
          var project = _context.CurrentProject;
-
-         moBiSimulation.ResultsDataRepository = simulation.ResultsDataRepository;
+         var originalSimulationName = moBiSimulation.Name;
          if (!_nameCorrector.CorrectName(project.Simulations, moBiSimulation))
             return;
 
-         renameCollidingModules(simulation.Modules, moBiSimulation.Name, project.Modules);
+         if (originalSimulationName != moBiSimulation.Name) //has been renamed
+            renameColidingModules(simulation.Modules, moBiSimulation.Name, project.Modules, originalSimulationName);
+
          renameCollidingEntities(simulation.Configuration.ExpressionProfiles, project.ExpressionProfileCollection);
 
          if (simulation.Configuration.Individual != null)
@@ -72,24 +73,32 @@ namespace MoBi.Core.Services
 
          addSimulationConfigurationToProject(moBiSimulation, loadCommand);
 
+         moBiSimulation.ResultsDataRepository = simulation.ResultsDataRepository;
+
          moBiSimulation.HasChanged = true;
          loadCommand.AddCommand(new AddSimulationCommand(moBiSimulation));
       }
 
-      private void renameCollidingModules(IEnumerable<Module> modulesToRename, string simulationName, IReadOnlyList<IWithName> existingEntities)
+      private void renameColidingModules(IReadOnlyList<Module> modulesToRename, string simulationName, IReadOnlyList<Module> existingModules, string originalSimulationName)
       {
-         var takenNames = existingEntities.AllNames();
+         //these names are unique so better creating a HashSet.
+         var takenNames = existingModules.AllNames().ToHashSet();
 
          modulesToRename
-            .Where(x => takenNames.Contains(x.Name))
             .Select((x, index) => new { Module = x, Index = index })
-            .Each(item => item.Module.Name = getModuleNameByIndex(simulationName,item.Index) );
+            .Each(item => renameModulesAfterSimulation(item.Module, simulationName, originalSimulationName));
 
-         //If there are still colliding names based on the simulation name, we need to correct them
+         // Correct any remaining name conflicts
          modulesToRename.Where(x => takenNames.Contains(x.Name)).Each(x => _nameCorrector.AutoCorrectName(takenNames, x));
       }
 
-      private string getModuleNameByIndex(string name, int index)=>
+      private void renameModulesAfterSimulation(Module module, string simulationName, string originalSimulationName)
+      {
+         if (module.Name.StartsWith(originalSimulationName))
+            module.Name.Replace(originalSimulationName, simulationName);
+      }
+
+      private string getModuleNameByIndex(string name, int index) =>
          $"{name}{(index == 0 ? string.Empty : " " + index)}";
 
       private void renameCollidingEntities(IEnumerable<IObjectBase> entitiesToRename, IReadOnlyList<IWithName> existingEntities)
--- a/src/MoBi.Core/Services/SimulationLoader.cs
+++ b/src/MoBi.Core/Services/SimulationLoader.cs
@@ -56,13 +56,6 @@ namespace MoBi.Core.Services
          var moBiSimulation = simulation;
 
          var project = _context.CurrentProject;
-         var originalSimulationName = moBiSimulation.Name;
-         if (!_nameCorrector.CorrectName(project.Simulations, moBiSimulation))
-            return;
-
-         if (originalSimulationName != moBiSimulation.Name) //has been renamed
-            renameColidingModules(simulation.Modules, moBiSimulation.Name, project.Modules, originalSimulationName);
-
          renameCollidingEntities(simulation.Configuration.ExpressionProfiles, project.ExpressionProfileCollection);
 
          if (simulation.Configuration.Individual != null)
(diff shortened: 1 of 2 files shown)


src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -68,14 +72,26 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
+            .Where(x => takenNames.Contains(x.Name))
+            .Select((x, index) => new { Module = x, Index = index })
+            .Each(item => item.Module.Name = getModuleNameByIndex(simulationName,item.Index) );
+
+         //If there are still colliding names based on the simulation name, we need to correct them
+         modulesToRename.Where(x => takenNames.Contains(x.Name)).Each(x => _nameCorrector.AutoCorrectName(takenNames, x));
+      }
+
+      private string getModuleNameByIndex(string name, int index)=>
+         $"{name}{(index == 0 ? string.Empty : " " + index)}";
```
rwmcintosh, line 93: You can probably use the `nameCorrector.AutoCorrectName` which does the index generation for you. I think you could get rid of the `Select` above and just do an `Each` and it should be simple enough that you don't need this method then.

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -68,14 +72,26 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
       }
 
+      private void renameCollidingModules(IEnumerable<Module> modulesToRename, string simulationName, IReadOnlyList<IWithName> existingEntities)
+      {
+         var takenNames = existingEntities.AllNames();
+
+         modulesToRename
+            .Where(x => takenNames.Contains(x.Name))
+            .Select((x, index) => new { Module = x, Index = index })
+            .Each(item => item.Module.Name = getModuleNameByIndex(simulationName,item.Index) );
```
rwmcintosh, line 86: here - try to use _nameCorrector.AutoCorrect method

rwmcintosh (changes requested):

src/MoBi.Core/Services/SimulationLoader.cs
rwmcintosh: ah, no, that won't work with Autocorrect because it just enumerates the current name, you actually want to rename and index in one step

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -68,14 +72,26 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
 
-         moBiSimulation.ResultsDataRepository = simulation.ResultsDataRepository;
-         if (!_nameCorrector.CorrectName(project.Simulations, moBiSimulation))
-            return;
-
          moBiSimulation.HasChanged = true;
          loadCommand.AddCommand(new AddSimulationCommand(moBiSimulation));
       }
 
+      private void renameCollidingModules(IEnumerable<Module> modulesToRename, string simulationName, IReadOnlyList<IWithName> existingEntities)
```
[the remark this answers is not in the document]
rwmcintosh: Make the first argument an `IReadOnlyList` instead of `IEnumerable`

src/MoBi.Core/Services/SimulationLoader.cs
rwmcintosh, line 86: What will happen if a simulation is loaded and there is a colliding simulation name, but no colliding module names? I think the loaded modules will not be renamed right?

In this step here, you want to check that any modules which were named after the old simulation name are renamed to the new simulation name even if there isn't already an existing module in the project with that name already. I would do that with a string replace because the modules being loaded might already be indexed.

Then you are correctly indexing if a collision is introduced on the renamed modules (below)

rwmcintosh (changes requested):

tests/MoBi.Tests/Core/SimulationLoaderSpecs.cs
```diff
@@ -14,6 +14,8 @@
 using MoBi.Helpers;
 using OSPSuite.Core.Commands.Core;
 using OSPSuite.Utility.Extensions;
+using DevExpress.DataProcessing.InMemoryDataProcessor;
+using OSPSuite.SimModel;
```
rwmcintosh, line 18: These are not needed and should be removed

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -56,8 +56,13 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
          var moBiSimulation = simulation;
 
          var project = _context.CurrentProject;
+         var originalSimulationName = moBiSimulation.Name;
+         if (!_nameCorrector.CorrectName(project.Simulations, moBiSimulation))
+            return;
+
+         if (originalSimulationName != moBiSimulation.Name) //has been renamed
+            renameColidingModules(simulation.Modules, moBiSimulation.Name, project.Modules, originalSimulationName);
```
rwmcintosh, line 64: I would move this all down below line 76 as it was before because this changes the behavior slightly now. Instead of renaming only a cloned simulation this will rename the original too. I'm not sure we should change that in this case.

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -69,13 +74,33 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
 
          moBiSimulation.ResultsDataRepository = simulation.ResultsDataRepository;
-         if (!_nameCorrector.CorrectName(project.Simulations, moBiSimulation))
-            return;
 
          moBiSimulation.HasChanged = true;
          loadCommand.AddCommand(new AddSimulationCommand(moBiSimulation));
       }
 
+      private void renameColidingModules(IReadOnlyList<Module> modulesToRename, string simulationName, IReadOnlyList<Module> existingModules, string originalSimulationName)
```
rwmcintosh, line 82: In this case we aren't only renaming colliding modules so let's just take that out of the method name. Maybe this method should be called `correctModuleNames` because there is two types of correction happening

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -69,13 +74,33 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
+      {
+         //these names are unique so better creating a HashSet.
+         var takenNames = existingModules.AllNames().ToHashSet();
+
+         modulesToRename
+            .Select((x, index) => new { Module = x, Index = index })
```
rwmcintosh, line 88: You don't need the select anymore since you aren't using the index

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -69,13 +74,33 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
+         modulesToRename.Where(x => takenNames.Contains(x.Name)).Each(x => _nameCorrector.AutoCorrectName(takenNames, x));
+      }
+
+      private void renameModulesAfterSimulation(Module module, string simulationName, string originalSimulationName)
+      {
+         if (module.Name.StartsWith(originalSimulationName))
+            module.Name.Replace(originalSimulationName, simulationName);
+      }
+
+      private string getModuleNameByIndex(string name, int index) =>
```
rwmcintosh, line 101: this method is unused

rwmcintosh (changes requested):

rwmcintosh (approved):

src/MoBi.Core/Services/SimulationLoader.cs
```diff
@@ -69,13 +67,36 @@ private void addSimulationToProject(IMoBiSimulation simulation, MoBiMacroCommand
...
+         modulesToRename
+            .Each(module=> renameModulesAfterSimulation(module, simulationName, originalSimulationName));
+
+         // Correct any remaining name conflicts
+         modulesToRename.Where(x => takenNames.Contains(x.Name)).Each(x => _nameCorrector.AutoCorrectName(takenNames, x));
+      }
+
+      private void renameModulesAfterSimulation(Module module, string simulationName, string originalSimulationName)
+      {
+         if (module.Name.StartsWith(originalSimulationName))
```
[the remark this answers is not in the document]
Yuri05: ok, thx

Yuri05 (approved):

PR #1580, opened 2024-10-16
1557 building blocks multiselect delete

### LINKED ISSUE #1557: Building Blocks - Multiselect - Delete
Add the "Delete" action on multi-selected Individuals.

benjaperez1983:
Fixes #1557

# Description

Allow deletion of multiple Building Blocks at the same time with multiselect + right click + Delete context menu.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

image

# Questions (if appropriate):


src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
```diff
@@ -0,0 +1,134 @@
...
+
+         public InteractionTasksForMultipleBuildingBlocks(
+            IInteractionTasksForChildren<Module, EventGroupBuildingBlock> interactionTasksForApplicationMoleculeBuilder,
+            IInteractionTasksForChildren<Module, InitialConditionsBuildingBlock> interactionTasksForInitialCondition,
+            IInteractionTasksForChildren<Module, MoleculeBuildingBlock> interactionTasksForMolecule,
+            IInteractionTasksForChildren<Module, PassiveTransportBuildingBlock> interactionTasksForPassiveTransport,
+            IInteractionTasksForChildren<Module, ParameterValuesBuildingBlock> interactionTasksForParameterValues,
+            IInteractionTasksForChildren<Module, MoBiReactionBuildingBlock> interactionTasksForMobiReaction,
+            IInteractionTasksForChildren<Module, MoBiSpatialStructure> interactionTasksForMobiSpatialStructure,
+            IInteractionTasksForChildren<Module, ObserverBuildingBlock> interactionTasksForObserver,
```
rwmcintosh, line 40: These should be injected with the concrete class names. eg. `IInteractionTasksForMoleculeBuildingBlock`. I'm guessing you are getting the write classes, but the naming is off.

image


There are also some missing. For example `IndividualBuildingBlock` is not a child of `Module` so you don't have the interaction task for that. Same for `ExpressionBuildingBlock`

There should be a concrete type for all these

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
```diff
@@ -0,0 +1,134 @@
...
+         {
+            _interactionTasksForApplicationMoleculeBuilder = interactionTasksForApplicationMoleculeBuilder;
+            _interactionTasksForInitialCondition = interactionTasksForInitialCondition;
+            _interactionTasksForMolecule = interactionTasksForMolecule;
+            _interactionTasksForPassiveTransport = interactionTasksForPassiveTransport;
+            _interactionTasksForParameterValues = interactionTasksForParameterValues;
+            _interactionTasksForMobiReaction = interactionTasksForMobiReaction;
+            _interactionTasksForMobiSpatialStructure = interactionTasksForMobiSpatialStructure;
+            _interactionTasksForObserver = interactionTasksForObserver;
+            _interactionTaskContext = interactionTaskContext;
```
rwmcintosh, line 51: I would expect all these names to end with `...BuildingBlock`. There are interaction tasks for builders, but in this case we aren't going to use them.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
```diff
@@ -0,0 +1,134 @@
...
+               messageBuilder.AppendLine(AppConstants.Dialog.BuildingBlocksUsedInSimulation);
+               messageBuilder.AppendLine();
+
+               foreach (var item in referringSimulationsAndBuildingBlocks)
+               {
+                  var buildingBlock = item.Item2;
+                  var simulations = item.Item1;
+
+                  messageBuilder.AppendLine($"\t- {AppConstants.CannotRemoveBuildingBlockFromProject(buildingBlock.Name, simulations.AllNames())}");
+               }
```
rwmcintosh, line 86: Can you move all this string manipulation into `AppConstants`. There are a few examples. That removes the complication of building the messages away from this class. As much as possible, pass strings to the AppConstants methods you create rather than the domain objects.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
```diff
@@ -0,0 +1,134 @@
...
+                  break;
+
+               case ObserverBuildingBlock observerBuildingBlock:
+                  _interactionTasksForObserver.Remove(observerBuildingBlock, observerBuildingBlock.Module, observerBuildingBlock, true);
+                  break;
+
+               default:
+                  throw new InvalidOperationException($"No interaction task found for building block type: {buildingBlockToRemove.GetType()}");
+            }
+         }
```
rwmcintosh, line 131: You'll need to capture the returned commands from each of these and make sure they are added to the history.

Also, where you have this constant parameter `true`, use the parameter name so we know what it does `silent: true`

rwmcintosh (changes requested):

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
rwmcintosh: You don't need to use the generic registration. You should just use the convention registration which means
IInteractionTasksForEventBuildingBlock => InteractionTasksForEventBuildingBlock. That's the easiest way to get the tasks you want.

That way you don't have to use the `Module` in the generic, and you can easily get the tasks for individual and expression because they are registered by convention.

Doesn't the multiselect context menu work with Individual BuildingBlocks and Expression Building Blocks?

rwmcintosh: I made a mistake sayngi "These should be injected with the concrete class names. " This is not by class name, but by interface name. I just mean you should use the interface without the generics.

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
```diff
@@ -0,0 +1,135 @@
...
+               _interactionTaskContext.DialogCreator.MessageBoxInfo(messageBuilder.ToString());
+            }
+         }
+
+         private void removeBuildingBlock(IBuildingBlock buildingBlockToRemove)
+         {
+            switch (buildingBlockToRemove)
+            {
+               case EventGroupBuildingBlock eventGroupBuildingBlock:
+                  _interactionTaskContext.Context.AddToHistory(_interactionTasksForEventGroupBuildingBlock.Remove(eventGroupBuildingBlock, eventGroupBuildingBlock.Module, eventGroupBuildingBlock, silent: true));
```
rwmcintosh, line 90: This probably adds all the remove commands as separate entries in the history. Instead of that, can you add them to a macro command and then add the macro command in the history? You'd probably capture all the commands in the foreach above.

rwmcintosh (changes requested):
See comments

src/MoBi.Presentation/Tasks/Interaction/InteractionTasksForMultipleBuildingBlocks.cs
rwmcintosh: The CI build has failed

rwmcintosh (approved):

src/MoBi.Assets/AppConstants.cs
```diff
@@ -888,6 +889,8 @@ public static class Dialog
          public static readonly Size SELECT_SINGLE_SIZE = new Size(475, 160);
          public static readonly Size PARAMETER_SELECTION_SIZE = new Size(1200, 800);
          public static readonly string RemoveMultipleModules = "Do you really want to remove the selected Modules?";
+         public static readonly string RemoveMultipleBuildingBlocks = "Do you really want to remove the selected Building Blocks?";
+         public static readonly string BuildingBlocksUsedInSimulation = "The following building blocks could not be removed from the project";
```
Yuri05, line 893: should we consistenly use low case? 
Because the 1st message is "...Building Blocks..." and the 2nd is is "...building blocks..."

src/MoBi.Presentation/MenusAndBars/ContextMenus/MultipleBuildingBlockNodeContextMenuFactory.cs
```diff
@@ -35,10 +40,15 @@ public MultipleBuildingBlockNodeContextMenu(IReadOnlyList<IBuildingBlock> buildi
...
-         if (buildingBlocks.Count == 2)
+         if (buildingBlocks.Count == 2 && (buildingBlocks[0].GetType() == buildingBlocks[1].GetType()))
             yield return ComparisonCommonContextMenuItems.CompareObjectsMenu(buildingBlocks, context, _container);
 
          yield return ObjectBaseCommonContextMenuItems.AddToJournal(buildingBlocks, _container);
+
+         if (buildingBlocks.Count > 1)
+            yield return CreateMenuButton.WithCaption(AppConstants.MenuNames.Delete)
+               .WithCommandFor<RemoveMultipleBuildingBlocksUICommand, IReadOnlyList<IBuildingBlock>>(buildingBlocks, _container)
+               .WithIcon(ApplicationIcons.Delete);
```
Yuri05, line 51: If BBS count is = 1: I don't get the delete menu? Or is it added elsewhere?

rwmcintosh: Yeah it's added elsewhere. This CM is only when you have multiple selected. For that reason, it probably does not need to be checked.

Yuri05: ok

Yuri05 (approved):

PR #1635, opened 2024-11-07
Fixes #1618 It should not be possible to create multiple neighborhoods with same neighbors

### LINKED ISSUE #1618: It should not be possible to create multiple neighborhoods with same neighbors
[neighborhoods-diff-name-same-neighbors.zip](https://github.com/user-attachments/files/17630328/neighborhoods-diff-name-same-neighbors.zip)

Currently, it is possible to create multiple neighborhoods between the same two containers but with different names. This can result in transport processes being created multiple times and unexpected model behavior (spent almost 5 hours debugging a project).

This should not be allowed; only one neighborhood between the same two containers should be allowed. Therefore, at the simulation creation step (when combining the modules), for each NEW neighborhood (i.e., the neighborhood with a new name), it should be checked whether a neighborhood between the specified containers already exists (disregarding what the first or the second neighbor is). 

To speed up the process, maintain a map of neighborhood paths that are already connected?

If a new neighborhood between two already connected containers is being added, show an error "Neighborhood \<NeighborhoodName> cannot be created because a neighborhood between the containers \<Container1> and \<Container2> already exists. Every pair of containers can be connected by one neighborhood only".

Attached is a project for testing. It contains three modules and three combinations of the modules (simulations). "Module 1" has three containers - A, B, and C, and the neighborhoods between containers A and B and between B and C, and the simulation "Valid_Simulation" can be created from this one module:

image

Module 2 defines a neighborhood between containers B and A, and it is combined with the first module to the simulation "Invalid_1", which has two neighborhoods between A and B. This should NOT be allowed:

image

Module 3 defines a neighborhood between A and B (the same as Module 1, but with a different name). It is combined with the first module to the simulation "Invalid_2", which again has two neighborhoods between A and B. This should NOT be allowed:

image

rwmcintosh:
Fixes #1618

# Description
Making sure that when a user creates a new neighborhood within a spatial structure, that one does not already exist for those two neighbors.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

Creating new neighborhoods:
image

image

Editing neighbors in an existing neighborhood
image

image

When you hand edit the paths
image

# Questions (if appropriate):


src/MoBi.Presentation/DTO/NeighborhoodBuilderDTO.cs
```diff
@@ -1,15 +1,56 @@
...
+      {
+         var uniqueNeighbors = existingNeighborhoods.SelectMany(x => new[] { x.FirstNeighborPath, x.SecondNeighborPath }).Distinct().Where(x => x != null);
+         _connections = new Cache<string, List<string>>();
+
+         uniqueNeighbors.Each(x =>
+         {
+            existingNeighborhoods.Where(y => Equals(y.FirstNeighborPath, x)).Each(y => addConnection(_connections, x, y.SecondNeighborPath));
+            existingNeighborhoods.Where(y => Equals(y.SecondNeighborPath, x)).Each(y => addConnection(_connections, x, y.FirstNeighborPath));
+         });
+      }
```
Yuri05, line 42: don't understand the workflow. Why not just 
`existingNeighborhoods.Each(nbh => addConnection(_connections, nbh.FirstNeighborPath, nbh.SecondNeighborPath)`

And `addConnection()` should check that both paths are not null before adding the connection

Yuri05 (approved):

PR #1658, opened 2024-11-14
1578 commit to bb fixes

### LINKED ISSUE #1578: Commit to BB
When commiting changes from simulation, show a dialog:

- "Are you sure"`?
- State the name of the PV BB and the module where the values will be written into.

benjaperez1983:
Fixes #1578

# Description

Changes on message when commiting to Building Blocks

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -283,17 +283,17 @@ namespace MoBi.Assets
 
          public static string CommitingChangesToModulesMessage(ModuleConfiguration moduleConfiguration, bool hasMoleculeChanges, bool hasParameterChanges)
          {
-            string message = $"Changed parameter Values will be committed to the PV of the module <i>{moduleConfiguration.Module.Name}</i>:{Environment.NewLine}";
+            string message = $"Changed parameter values will be committed to the parameter values of the module <i>{moduleConfiguration.Module.Name}</i>:{Environment.NewLine}";
 
             if (hasMoleculeChanges)
             {
-               var icName = moduleConfiguration.SelectedInitialConditions == null ? "A new Initial Conditions Building Block will be created" : $"Initial Conditions will be written to the IC building block <i>{moduleConfiguration.SelectedInitialConditions.DisplayName}</i>";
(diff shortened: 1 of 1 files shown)


src/MoBi.Assets/AppConstants.cs
```diff
@@ -283,17 +283,17 @@ public static string DisplayValue(double displayValue, string displayUnit)
 
          public static string CommitingChangesToModulesMessage(ModuleConfiguration moduleConfiguration, bool hasMoleculeChanges, bool hasParameterChanges)
          {
-            string message = $"Changes will be applied in the module <i>{moduleConfiguration.Module.Name}</i>:{Environment.NewLine}";
+            string message = $"Changed parameter Values will be committed to the PV of the module <i>{moduleConfiguration.Module.Name}</i>:{Environment.NewLine}";
```
rwmcintosh, line 286: Pavel wrote the comments with contractions and with mixed case but this should be fixed in the caption

Don't use PV or IC contractions and usually when we mean building block we have to spell it out as "Parameter Values Building Block".

this also says "parameter Values" where Values has the wrong case.

Make sure this is consistent for all three of the lines changed here.

rwmcintosh (changes requested):

src/MoBi.Assets/AppConstants.cs
```diff
@@ -283,17 +283,17 @@ public static string DisplayValue(double displayValue, string displayUnit)
 
          public static string CommitingChangesToModulesMessage(ModuleConfiguration moduleConfiguration, bool hasMoleculeChanges, bool hasParameterChanges)
          {
-            string message = $"Changes will be applied in the module <i>{moduleConfiguration.Module.Name}</i>:{Environment.NewLine}";
+            string message = $"Changed parameter values will be committed to the parameter values of the module <i>{moduleConfiguration.Module.Name}</i>:{Environment.NewLine}";
```
Yuri05, line 286: would extend the message with _...will be committed to the parameter values **building block** of the module..._

PR #1706, opened 2024-12-16
1110 simulation sync status after project conversion

### LINKED ISSUE #1110: Simulation sync status after project conversion
When I open a project from the test data in MoBi new modules are created and building blocks are grouped.

In the simple case where you have a single simulation, and just enough building blocks for one module, only one module is created and it's linked to the only simulation. 

The sync status (red/green) cannot be reliably established though. MoBi can tell the version of the building blocks, and therefore MoBi checks if those match. However, MoBi keeps track of simulation changes using local change tracking objects, and during a project conversion, there is never any present.

The sync status therefore, is unreliable after a v11 project is updated.

rwmcintosh:
Fixes #1110 

# Description
Changes in V11 simulations cannot be traced back to the building blocks when converted to V12. This PR shows a notification at the time of project conversion and another notification if the simulation changes are committed to building blocks.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

image

# Questions (if appropriate):


src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -49,6 +49,7 @@ public interface IMoBiSimulation : IWithDiagramFor<IMoBiSimulation>, ISimulation
       IReadOnlyList<IBuildingBlock> BuildingBlocks();
 
       IReadOnlyCollection<OriginalQuantityValue> OriginalQuantityValues { get; }
+      bool HasUntraceableChanges { get; set; }
```
msevestre, line 52: I think this guy deserves some comments.
Also it sucks to have a property here that is just required for a one time project conversion...

src/MoBi.Core/Commands/UpdateSimulationCommand.cs
```diff
@@ -22,28 +22,32 @@ public class UpdateSimulationCommand : MoBiReversibleCommand
...
          _changedBuildingBlockType = new ObjectTypeResolver().TypeFor(templateBuildingBlock);
       }
 
-      private UpdateSimulationCommand(IMoBiSimulation simulationToUpdate, IModel newModel, SimulationConfiguration updatedSimulationConfiguration, bool wasChanged,
+      private UpdateSimulationCommand(IMoBiSimulation simulationToUpdate, IModel newModel, SimulationConfiguration updatedSimulationConfiguration, bool wasChanged, bool hasUntraceableChanges,
          string buildingBlockName, string changedBuildingBlockType)
       {
          _hasChanged = wasChanged;
+         _hasUntraceableChanges = hasUntraceableChanges;
+         _hadUntraceableChanges = simulationToUpdate.HasUntraceableChanges;
```
msevestre, line 50: it took me  a minute to see the `has` vs `had`

src/MoBi.Core/Serialization/Converter/v12/Converter11To12.cs
```diff
@@ -28,10 +30,34 @@ public Converter11To12(Converter110To120 coreConverter)
 
          element.DescendantsAndSelfNamed("MoBiSimulation").Each(simulationNode =>
          {
+            var simulationChanges = simulationNode.DescendantsAndSelfNamed("BuildConfiguration").Any(hasSimulationChanges);
+            if (simulationChanges)
+               simulationNode.AddAttribute("hasUntraceableChanges", true.ToString());
```
msevestre, line 35: true to string? lol. what about "true"?

msevestre: yeah... I don't have a better idea but this feels kind of wrong (The property not the boolean)

src/MoBi.Presentation/Presenter/Main/NotificationPresenter.cs
```diff
@@ -250,11 +249,25 @@ public void ShowContextMenu(IViewItem notificationItem, Point popupLocation)
          contextMenu.Show(_view, popupLocation);
       }
 
+      public void Handle(ShowProjectConversionNotificationsEvent eventToHandle)
+      {
+         Handle(eventToHandle as ShowNotificationsEvent);
+         notifyForUntraceableChanges(eventToHandle.NotificationMessages.Select(x => x.Object as MoBiSimulation).Where(x => x != null && x.HasUntraceableChanges).Distinct().ToList());
```
msevestre, line 255: what about we create a local variables here?

msevestre: cannot we have an objec that has the simulation and this flag just for the conversion? instead of havign it into the simulation

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -94,9 +95,18 @@ public IMoBiCommand CommitSimulationChanges(IMoBiSimulation simulationWithChange
 
          macroCommand.RunCommand(_context);
          _context.PublishEvent(new SimulationStatusChangedEvent(simulationWithChanges));
+
+         if (simulationWithChanges.HasUntraceableChanges)
+            showWarningDialogForUntraceableChanges(simulationWithChanges);
```
msevestre, line 100: why are we showing this again here? Isn't the conversion warning enough?

src/MoBi.Presentation/Presenter/Main/NotificationPresenter.cs
rwmcintosh: The simulation is keeping this flag because the sync status remains red in the explorer until it is reconfigured. That's also why you get the reminder each time you commit and not only once during conversion. I'm not sure if that's necessary but it seems right. The simulation is not in sync and it will never be without reconfigure.

rwmcintosh:
Updated warning message during conversion
image

Updated commit message as an error and modified text. Also, no commits to BB's are done when this simulation has this status.
image

PR #1796, opened 2025-02-04
Fixes #1045 Silent crash when creating a circular reference

### LINKED ISSUE #509: Crash when creating a circular reference
1) Open MoBi
2) Create new project
3) In the SpatialStructures-BB (or any other - doesn't matter) create a parameter with the name "Parameter" (name doesn't matter).
4) Change it's value to an explicit formula
5) Create a new reference with the path "..|Parameter" and alias name "Parameter".
6) Type in "Parameter" into the formula field - silent crash.

Sometimes it crashes directly after entering "..|Parameter" into the path field.

### LINKED ISSUE #1045: Silent crash when creating a circular reference
- Create an empty MoBi project
- In the spatial structures BB, create a parameter in the "Individuals"-BB. Name: "test", formula type - "explicit formula"
- Create a formula with one variable "test", path = `..|test`
image

Click OK -> MoBi disappears.

rwmcintosh:
Fixes #1045
Fixes #509

# Description
When adding a new alias in an explicit formula it was possible to crash MoBi on a stack overflow exception.

This is just an explicit check to make sure that the new path does not create a circular reference. The exception throw is the same as the strategy when creating a new parameter.

This fixes both issues because we are catching the circular reference at the stage where the path is entered, which is common to creating or editing a parameter formula. You will not be able to enter the value like this as described in the issue.

image

The exception is thrown for a different parameter name, but the dialog is shown and then the value is cleared

image

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditFormulaPathListPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditFormulaPathListPresenter.cs
@@ -124,6 +124,9 @@ namespace MoBi.Presentation.Presenter
          var path = new ObjectPath(newPath.ToPathArray());
          var formulaUsablePath = _formula.FormulaUsablePathBy(dto.Alias);
 
+         if (_circularReferenceChecker.HasCircularReference(path, _formulaOwner))
+            throw new OSPSuiteException(AppConstants.Exceptions.CircularReferenceException(path, _formula));
+
          AddCommand(_moBiFormulaTask.ChangePathInFormula(_formula, path, formulaUsablePath, BuildingBlock));
       }
 
--- a/src/MoBi.BatchTool/MoBi.BatchTool.csproj
+++ b/src/MoBi.BatchTool/MoBi.BatchTool.csproj
@@ -60,8 +60,8 @@
     <PackageReference Include="OSPSuite.Core" Version="12.0.364" />
     <PackageReference Include="OSPSuite.DevExpress" Version="21.2.3" />
     <PackageReference Include="OSPSuite.FuncParser" Version="4.0.0.68" GeneratePathProperty="true" />
-    <PackageReference Include="OSPSuite.SimModel" Version="4.0.0.68" GeneratePathProperty="true" />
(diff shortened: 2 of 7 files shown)


src/MoBi.Presentation/Presenter/EditFormulaPathListPresenter.cs
```diff
@@ -124,6 +124,9 @@ public void SetFormulaUsablePath(string newPath, FormulaUsablePathDTO dto)
          var path = new ObjectPath(newPath.ToPathArray());
          var formulaUsablePath = _formula.FormulaUsablePathBy(dto.Alias);
 
+         if (_circularReferenceChecker.HasCircularReference(path, _formulaOwner))
```
msevestre, line 127: I think we removed it because it takes way too much time on big projects
That's why it also became an option in the user settings.. Adding it back will create other issus but might be wrong

rwmcintosh: Is that an option only when building a simulation or also when entering each alias?

rwmcintosh: So, should I surround this with an option check or what do you think?

msevestre: I really don't know. I remember a time where EDITING of a fomula took forever in a project (from @PavelBal ) in fact
So we removed this But maybe it was due to the formula ?

PavelBal:
Well I don't know what this change does. I can't remember any more what the exact performance issue was, but it was a pain.

I imagine now it fails once it tries to evaluate the formula? Can't you just catch it and WHEN THE EXCEPTION IS THROWN, perform the circular check and inform the user about a potential problem?

rwmcintosh:
No, a stack overflow exception is terminal. You can only catch it when running it in a debugger. AFAIK.

Yuri05 (approved):

PR #1830, opened 2025-03-27
1718 do not lock UI when running a simulation

### LINKED ISSUE #1718: Do not lock UI when running a simulation

benjaperez1983:
Fixes #1718 

# Description

Allow Simulations to run in parallel, not locking the UI and making them async.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

image
image

# Questions (if appropriate):


rwmcintosh (commented):
I think you're only using the simulation name as part of the status message. Would a rename during the simulation execution throw an exception do you think? Or handled graceful?

Someone could also do something silly like delete a running simulation. Could that be prevented? I think it's a context menu only operation.

src/MoBi.Core/Events/Events.cs
```diff
@@ -20,8 +20,21 @@ public SimulationRunFinishedEvent(IMoBiSimulation simulation)
       }
    }
 
+   public class SimulationsRunCanceledEvent
+   {
+   }
+
    public class SimulationRunStartedEvent
    {
+      public SimulationRunStartedEvent()
```
msevestre, line 29: why the empty constuctor?

src/MoBi.Core/Events/Events.cs
```diff
@@ -20,8 +20,21 @@ public SimulationRunFinishedEvent(IMoBiSimulation simulation)
       }
    }
 
+   public class SimulationsRunCanceledEvent
```
msevestre, line 23: Should we rename this to be clear that it's plural. I missed it the first two times I read?
AllSimulationRunsCanceledEvent?

src/MoBi.Presentation/MenusAndBars/ContextMenus/RemoveEventBuilderFromEventGroupBuilderUICommand.cs
```diff
@@ -4,6 +4,7 @@
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Services;
 using OSPSuite.Presentation.UICommands;
+using System.Threading.Tasks;
```
msevestre, line 7: remove

src/MoBi.Presentation/Presenter/Main/StatusBarPresenter.cs
```diff
@@ -33,7 +40,10 @@ public class StatusBarPresenter : IStatusBarPresenter
       private readonly IMoBiConfiguration _moBiConfiguration;
       public event EventHandler StatusChanged = delegate { };
       private int _numberOfReportsBeingCreated;
+      private readonly ConcurrentDictionary<string, bool> _simulations = new ConcurrentDictionary<string, bool>(); 
 
+      private int activeSimulations => _simulations.Count(x => x.Value == true);
```
msevestre, line 45: no need for ==true. It's already a bool

src/MoBi.Presentation/Presenter/Main/StatusBarPresenter.cs
```diff
@@ -118,14 +128,36 @@ public void Handle(ProgressingEvent eventToHandle)
             .WithValue(eventToHandle.ProgressPercent);
 
          update(StatusBarElements.ProgressStatus)
-            .WithCaption(eventToHandle.Message);
+            .WithCaption($"{activeSimulations} {eventToHandle.Message}");
       }
 
       public void Handle(ProgressDoneEvent eventToHandle)
       {
+         var message = eventToHandle is ProgressDoneWithMessageEvent ?
```
msevestre, line 136: why not listen to two messages? Handle specifically does it for you so you can be specific. There is not enough common code here and if there is, it can be in a function IMO

src/MoBi.Presentation/Presenter/Main/StatusBarPresenter.cs
```diff
@@ -172,5 +204,19 @@ public void Handle(ReportCreationFinishedEvent eventToHandle)
...
+         _simulations[eventToHandle.Simulation.Id] = false;
+         if (activeSimulations == 0)
+         {
+            resetCountersAndHideBar();
+         }
+      }
+
+      public void Handle(SimulationRunStartedEvent eventToHandle)
+      {
+         _simulations[eventToHandle.Simulation.Id] = true;
```
msevestre, line 219: are these the rigth methods to use with a concurrent dictionary? I though we were supposed to use AddOrUpdate etc.. Otherwise you may be using it differently. This needs to be checked

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -52,9 +56,20 @@ public SimulationRunner(IMoBiContext context,
          _simModelManagerFactory = simModelManagerFactory;
          _keyPathMapper = keyPathMapper;
          _entityValidationTask = entityValidationTask;
+
       }
 
-      public void RunSimulation(IMoBiSimulation simulation, bool defineSettings = false)
+      public async void RunSimulation(IMoBiSimulation simulation, bool defineSettings = false)
+      {
+         _ = runSimulationAsync(simulation, defineSettings);
```
msevestre, line 64: what's the _? I dont get it

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -52,9 +56,20 @@ public SimulationRunner(IMoBiContext context,
...
+      }
+
+      public async Task RunSimulationAsync(IMoBiSimulation simulation, bool defineSettings = false)
+      {
+         await runSimulationAsync(simulation, defineSettings);
```
msevestre, line 69: return the task. No need to await

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -97,44 +113,95 @@ private bool settingsRequired(IMoBiSimulation simulation, bool defineSettings)
          return !simulation.OutputSelections.HasSelection;
       }
 
-      public void StopSimulation()
+      public void StopSimulation(IMoBiSimulation simulation = null)
```
msevestre, line 116: yeah clearly two methods since they don't even share one line of code

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -97,44 +113,95 @@ private bool settingsRequired(IMoBiSimulation simulation, bool defineSettings)
...
       }
 
+      private void setSimulationRunning(IMoBiSimulation simulation, bool isRunning)
+      {
+         if (isRunning)
+         {
+            if (!_originalSimulationNames.ContainsKey(simulation))
+               _originalSimulationNames[simulation] = simulation.Name;
+
+            if (!simulation.Name.EndsWith(" (Running)"))
```
msevestre, line 193: NO! Please le'ts npot rename the simulation to run them. Why

src/MoBi.Presentation/UICommand/RunActiveSimulationCommand.cs
```diff
@@ -3,6 +3,7 @@
 using MoBi.Core.Services;
 using OSPSuite.Core.Services;
 using OSPSuite.Presentation.UICommands;
+using System.Threading.Tasks;
```
msevestre, line 6: remove

msevestre (changes requested):
Some code to split IMO
Absolutely against renaming a model object for the sake of UI updates.

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -52,9 +56,20 @@ public SimulationRunner(IMoBiContext context,
          _simModelManagerFactory = simModelManagerFactory;
          _keyPathMapper = keyPathMapper;
          _entityValidationTask = entityValidationTask;
+
       }
 
-      public void RunSimulation(IMoBiSimulation simulation, bool defineSettings = false)
+      public async void RunSimulation(IMoBiSimulation simulation, bool defineSettings = false)
+      {
+         _ = runSimulationInternalAsync(simulation, defineSettings);
```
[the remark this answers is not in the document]
msevestre: why don't we await the code? since the method is async? Am i missing sthg obvious?

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -97,44 +113,95 @@ private bool settingsRequired(IMoBiSimulation simulation, bool defineSettings)
...
+            _context.PublishEvent(new SimulationsRunCanceledEvent());
+         }
       }
 
-      private void startSimulationRun(IMoBiSimulation simulation)
+      private async Task startSimulationRunAsync(IMoBiSimulation simulation)
       {
-         _simulation = simulation;
-         _context.PublishEvent(new SimulationRunStartedEvent());
+         if (simulation.IsRunning)
```
msevestre, line 143: I wonder how we can land this in the first place.

src/MoBi.Presentation/Tasks/SimulationRunner.cs
msevestre: This is presentation. So if you want to show in the tree, it could be done by handling an event (simulariion, running or finished etc.). If something goes wrong for some reason, this could lead to serious inconsistency

tests/MoBi.Tests/Presentation/Tasks/SimulationRunnerSpecs.cs
```diff
@@ -315,4 +320,165 @@ public void should_update_the_molecule_weight_of_all_concentration_columns_relat
...
+                  throw;
+               }
+
+               return new SimulationRunResults(warnings: Enumerable.Empty<SolverWarning>(), results: new DataRepository("R2"));
+            });
+
+         A.CallTo(() => _eventValidationTask.Validate(A<IMoBiSimulation>._)).Returns(true);
+      }
+
+      protected override void Because()
```
msevestre, line 456: Shouldn't this whole test be using the Async base instead of running the test with Wait all the time? 
It's just a quesiton because it's very hard to follow

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -277,5 +277,7 @@ public DataRepository ResultsDataRepository
       }
 
       public bool ComesFromPKSim => Creation.Origin == Origins.PKSim;
+
+      public bool IsRunning { get; set; }
```
msevestre, line 281: let's remove

src/MoBi.Presentation/Tasks/SimulationRunner.cs
```diff
@@ -97,41 +116,87 @@ private bool settingsRequired(IMoBiSimulation simulation, bool defineSettings)
...
+            }
             removeEvents();
-            _context.PublishEvent(new SimulationRunFinishedEvent(_simulation));
-            _simulation = null;
+            _context.PublishEvent(new SimulationRunFinishedEvent(simulation));
+            setSimulationRunning(simulation, false);
+         }
+      }
+
+      private void setSimulationRunning(IMoBiSimulation simulation, bool isRunning)
```
msevestre, line 189: let's remove this too

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
msevestre: REally let's remove it :)

msevestre (approved):
ok looks very good to me @benjaperez1983 
There is still one entry that needs to be removed (isRunning in the simuation)

PR #1849, opened 2025-04-11
1843 show individual in spatial structure

### LINKED ISSUE #1843: Show individual in spatial structure
- [x] #1848

Image

- [x] Change the label to "Select Individual:"
- [x] Tooltip: "Select an individual building block to display parameters that are defined in the individual and not in the spatial structure directly. These parameters will be added during simulation creation."
- [x] Let's disable editing. There are multiple problems with it, as I see.
  - When showing individual parameters in a spatial structure from a PK-Sim module, trying to edit a parameter that is part of the spatial structure converts the module to an extension module, while editing individual parameters is possible without such a message. Yes, technically it makes sense, but the behavior is inconsistent and thus confusing
  - Edited value is not properly updated in the parameters list (but propagated to the individual BB, so its "just" a refresh problem):

Image
  - Most important - "Undo" in History does not work for this change!
  - "Convert to formula" is somewhat confusing as this action does not work like it in other views.
  - AND it does not work ;) If I cancel converting to formula, the value is changed to `NaN` with no way to revert it

So too many things for editing are not working properly yet. Instead of trying to solve them, I suggest to remove this feature. The "Find parameter" feature is good enough, and the user can quickly jump to the parameter in the Individual and change it there  (and be more aware of what they are doing).

In this view, let's have the "Warning" with the "Find parameter" only. The text could be:

"Please edit the parameter in the respective Individual Building Block.

rwmcintosh:
Fixes #1843

# Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

selection with tooltip
image

This is the view. You can only see the name and the instruction with the navigate button
image

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditParametersInContainerPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditParametersInContainerPresenter.cs
@@ -182,6 +182,7 @@ namespace MoBi.Presentation.Presenter
 
          initializeIndividualParameterCache(individualParameters, selectedIndividual);
          _allParametersDTO.AddRange(_individualParameterCache.Keys);
+         _allParametersDTO.Sort((a,b) => string.Compare(a.Name, b.Name, StringComparison.InvariantCultureIgnoreCase));
       }
 
       private void initializeIndividualParameterCache(IReadOnlyList<IndividualParameter> individualParameters, IndividualBuildingBlock selectedIndividual)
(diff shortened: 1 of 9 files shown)


src/MoBi.Presentation/Presenter/EditParametersInContainerPresenter.cs
```diff
@@ -182,6 +182,7 @@ private void createParameterCache(IReadOnlyList<IParameter> parametersToEdit, IR
 
          initializeIndividualParameterCache(individualParameters, selectedIndividual);
          _allParametersDTO.AddRange(_individualParameterCache.Keys);
+         _allParametersDTO.Sort((a,b) => string.Compare(a.Name, b.Name, StringComparison.InvariantCultureIgnoreCase));
```
rwmcintosh, line 185: I added this to sort the list after it's created from two different sources. Previously the individual stuff was always after the container stuff.

PR #1856, opened 2025-04-15
Fixes #1842 Process rate parameter cannot be resolved when creating a simulation

### LINKED ISSUE #1842: Process rate parameter cannot be resolved when creating a simulation
It looks like the paths within the ProcessRate parameter are not properly resolved

> I have a simulation for which I want to plot a process rate. I opened the "Passive transport" BB and ticked the boxes "create process rate parameter" and "plot process rate parameter". The simulation used to run, but once these boxes ticked, I get the following error:

rwmcintosh:
Fixes #1842

# Description
Creating a reference in a transport builder should be more like creating a reference in a reaction builder.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/MoBi.Presentation/Presenter/ObjectPathCreatorAtProcessBuilder.cs
```diff
@@ -0,0 +1,33 @@
...
+      {
+         var parentProcess = GetProcessBuilderFor(parameter);
+
+         // Always Absolute paths
+         if (parameter.BuildMode != ParameterBuildMode.Local)
+            return CreateFormulaUsablePathFrom(new[] { parentProcess.Name, parameter.Name }, parameter);
+
+         // Parameter is child of process
+         if (parentProcess.Equals(ProcessBuilder))
+            return CreateFormulaUsablePathFrom(new[] { parameter.Name }, parameter);
```
Yuri05, line 25: Is this correct? Shouldn't the path be `..|parameter.Name` and not just `parameter.Name`?

rwmcintosh: For ProcessRateParameter the ..| will be added when the parameter is created.

This just makes the alias path a relative path instead of absolute (from the builder root).

If the path is not relative, then when ProcessRate is created the ..| will not be added.

In core the service is called `ProcessRateParameterCreator`

Yuri05: and the function is called only for `ProcessRate `parameters? Or also for another transport/reaction parameters?

rwmcintosh: `ProcessRateParameterCreator` is used for `ProcessRate` parameters. Is that what you mean?

Yuri05: I meant, is the function here `protected FormulaUsablePath CreatePathToProcessProperty(IParameter parameter)` used only for `ProcessRate` parameters?

src/MoBi.Presentation/Presenter/ObjectPathCreatorAtTransport.cs
```diff
@@ -123,13 +128,21 @@ private string getPathKeywordForContainer(IContainer container)
...
 
          return string.Empty;
       }
+
+      protected override TransportBuilder GetProcessBuilderFor(IParameter parameter)
+      {
+         if (parameter.IsAtTransport())
+            return parameter.RootContainer as TransportBuilder;
+
+         throw new MoBiException($"cant find transport for parameter {parameter.Name}");
```
msevestre, line 145: @rwmcintosh WOndering why we start to put error message directly into the code like this. I have seen in multiple times. Is this our new way of doing exception instead of collating the text in one central place?

msevestre (commented):
Looks good to me. Just a phylosophical question really as to how we deal with error constant (I don't mean the debug stuff that can stay as is because it will never be shown to the user)

PR #1910, opened 2025-05-23
1904 fix issues based on pr review 1881

### LINKED ISSUE #1904: Fix issues based on PR review #1881
@rwmcintosh @benjaperez1983 

Some issues open for discussions, some changes required
Please see my comments here

Some tasks just so that we do not lose but please check it out in details 

- [x] Renaming of public properties to avoid the "Get" overused. It makes sense sometimes but not always IMO
- [x] Reuse of existing Factory classes from MoBi.project. In OSPSuite.Core, we created an OSPSuite.HElperForTests project so that we only have ONE central point for all these stuff used by all projects. Maybe we do the same here? Takes 10 min and boom

benjaperez1983:
Fixes  #1904

# Description

Mobi.R changes introduced for both 1904

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/tests/MoBi.HelpersForTests/MoBi.HelpersForTests.csproj
+++ b/tests/MoBi.HelpersForTests/MoBi.HelpersForTests.csproj
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Core" Version="13.0.13" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.14" />
   </ItemGroup>
 
   <ItemGroup>


rwmcintosh (changes requested):
I think you have to merge V13 into this branch. There's too many files listed. For example none of these files are new in this commit

image

It's probably because you branched this before your first PR was merged.

tests/MoBi.HelpersForTests/DomainHelperForSpecs.cs
```diff
@@ -10,27 +10,80 @@
 using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.Utility.Events;
 
-namespace MoBi.Helpers
+namespace MoBi.HelpersForTests
```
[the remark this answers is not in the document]
rwmcintosh: Let's keep it. Sometimes I create separate merges for these kinds of changes when they make this much noise.

src/MoBi.R/Services/SimulationCreateTask.cs
```diff
@@ -0,0 +1,23 @@
...
+using OSPSuite.R.Domain;
+
+namespace MoBi.R.Services
+{
+   public interface ISimulationCreateTask
+   {
+      Simulation CreateSimulationFrom(SimulationConfiguration simulationConfiguration);
+   }
+
+   public class SimulationCreateTask : ISimulationCreateTask
```
rwmcintosh, line 11: I would just call it `SimulationTask`. I bet there will be more operations that will be done with simulations.

tests/MoBi.R.Tests/Services/SimulationCreateTaskSpecs.cs
```diff
@@ -0,0 +1,116 @@
...
+            Module = moduleForSimulation,
+            SelectedInitialCondition = initialConditions.FirstOrDefault(),
+            SelectedParameterValue = parameterValues.FirstOrDefault()
+         });
+
+         simulationConfig.ExpressionProfiles.AddRange(
+            expressionProfileNames
+               .Select(x => _projectTask.ExpressionProfileByName(_project, x))
+               .Where(p => p != null)
+         );
```
rwmcintosh, line 68: Let's consider how this is going to be used from R. Maybe you can find some analogous code in OSPSuite.R/OSPSuite-R and determine if it would be more consistent to create a task method for some of this.

As it is, the R coder will have to call into C# separately to 
1. create the new simulation configuration,
2. create the new module configuration
3. call setters for module, initial conditions, and parameter values
4. retrieve the list of module configurations from the simulation configuration
5. call the add method on the C# object `List`
6. Similar for each `ExpressionProfile`

We have seen instances where C#-ism and domain objects are directly used by R code, but I feel like requiring the R coder to know that they are accessing a C# `List` and that it has an `Add` method.

At the other end of the ease-of-use scale for R coder could be a task/method that creates a `ModuleConfiguration` and it takes as arguments a `Module` and optional Selected...

Then another task that takes an enumeration of `ModuleConfiguration` and optional `Individual` to give a `SimulationConfiguration` more of a factory approach.

rwmcintosh (changes requested):
Before reviewing the rest, I think it's a good idea to consider what the API should look like for R coders. There's a question of how much C#/.NET and our own domain logic we expect the R coder to understand.

msevestre:
>There's a question of how much C#/.NET and our own domain logic we expect the R coder to understand.

We need to abstract as much as possible in C# IMO and try to limite the R call to simple push pull stuff (e.g. call a method, get results), call another one maybe with the results. 
Some logic can be encapsulated in R code as well if it makes sens.e But the end user shoul dnot have to know anything about .NEt

src/MoBi.R/Services/SimulationFactory.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace MoBi.R.Services
+{
+   public interface ISimulationFactory
+   {
+      Simulation CreateSimulation(SimulationConfiguration configuration);
+   }
+
+   public class SimulationFactory : ISimulationFactory
+   {
+      private readonly IMoBiContext _context;
```
rwmcintosh, line 18: that's never used

tests/MoBi.R.Tests/Services/SimulationCreateTaskSpecs.cs
```diff
@@ -0,0 +1,106 @@
...
+         var moduleConfiguration = sut.CreateModuleConfiguration(moduleForSimulation, parameterValuesForModule, initialConditionForModule);
+
+         var moduleConfigurations = new List<ModuleConfiguration> { moduleConfiguration };
+
+         var simulationConfig = sut.CreateConfiguration(_simulationName, moduleConfigurations, expressionProfilesForSimulation, individualForSimulation);
+
+         _simulation = sut.CreateSimulationFrom(simulationConfig);
+      }
+
+      [Test]
```
rwmcintosh, line 60: should be `[Observation]`

src/MoBi.R/RRegister.cs
```diff
@@ -22,6 +28,7 @@ public override void RegisterInContainer(IContainer container)
          {
             scan.AssemblyContainingType<RRegister>();
             scan.IncludeNamespaceContainingType<IProjectTask>();
+            scan.IncludeNamespaceContainingType<ISimulationCreateTask>();
```
rwmcintosh, line 31: This is the same namespace as line 30 so you don't need this.

src/MoBi.R/RRegister.cs
```diff
@@ -30,6 +37,7 @@ public override void RegisterInContainer(IContainer container)
 
       private static void registerMinimalTypes(IContainer container)
       {
+         container.Register<Core.Domain.Services.ISimulationFactory, Core.Domain.Services.SimulationFactory>();
```
rwmcintosh, line 40: That should not be needed. It should be registered by `CoreRegister` try to remove and see if everything still works

src/MoBi.R/RRegister.cs
```diff
@@ -39,6 +47,11 @@ private static void registerMinimalTypes(IContainer container)
          container.Register<IMoBiConfiguration, IApplicationConfiguration, MoBiConfiguration>(LifeStyle.Singleton);
          container.Register<IMoBiXmlSerializerRepository, MoBiXmlSerializerRepository>(LifeStyle.Singleton);
          container.Register<IHistoryManagerFactory, HistoryManagerFactory>(LifeStyle.Singleton);
+         container.Register<IDiagramManagerFactory, DiagramManagerFactory>(LifeStyle.Singleton);
+         container.Register<IDimensionValidator, DimensionValidator>(LifeStyle.Singleton);
+         container.Register<IPresentationUserSettings, IUserSettings, ICoreUserSettings, BatchUserSettings>(LifeStyle.Transient);
+         container.Register<Services.ISimulationFactory, Services.SimulationFactory>(LifeStyle.Singleton);
```
rwmcintosh, line 53: That one should be included by registration convention and **should** not need to be explicitly registered. Try to remove.

src/MoBi.R/Services/ProjectTask.cs
```diff
@@ -3,17 +3,30 @@
...
+      IReadOnlyList<Simulation> AllSimulations(MoBiProject moBiProject);
+      IReadOnlyList<IndividualBuildingBlock> AllIndividuals(MoBiProject moBiProject);
+      IReadOnlyList<InitialConditionsBuildingBlock> AllInitialConditions(MoBiProject moBiProject);
+      IReadOnlyList<ParameterValuesBuildingBlock> AllParameterValues(MoBiProject moBiProject);
+      IReadOnlyList<InitialConditionsBuildingBlock> AllInitialConditionsFromModule(Module module);
+      IReadOnlyList<ParameterValuesBuildingBlock> AllParameterValuesFromModule(Module module);
+      Module ModuleByName(MoBiProject moBiProject, string name);
+      IndividualBuildingBlock IndividualBuildingBlockByName(MoBiProject moBiProject, string name);
+      InitialConditionsBuildingBlock InitialConditionBuildingBlockByName(Module module, string name);
+      ParameterValuesBuildingBlock ParameterValueBuildingBlockByName(Module module, string name);
```
rwmcintosh, line 28: I bet we are going to need a `ModuleTask` soon because there are 4 methods here that don't use project at all. Next when you start loading modules from the file system, you will want to get initial conditions from the module without needing the project task.

src/MoBi.R/Services/ProjectTask.cs
```diff
@@ -27,30 +40,65 @@ public ProjectTask(IMoBiContext moBiContext, IContextPersistor contextPersistor)
...
 
-      public IReadOnlyList<string> GetBuildingBlocksNamesFromModuleName(MoBiProject moBiProject, string moduleName)
-      {
-         var module = moBiProject.ModuleByName(moduleName);
-         if (module != null)
-            return module.BuildingBlocks.AllNames();
+      public List<ExpressionProfileBuildingBlock> ExpressionProfileBuildingBlocksByName(MoBiProject moBiProject, params string[] names) =>
+         moBiProject.ExpressionProfileCollection
+            .Where(p => names.Contains(p.Name))
+            .ToList();
```
rwmcintosh, line 49: Spacing is off.

rwmcintosh:
I'd check those last couple items.

I'm not sure that the intent of @msevestre comment about diagram managers was addressed. I think he wants to know if we can get away without registering these. That will require some refactoring in the presentation project probably.

I split that issue out because I knew that might take more refactoring and we can do it later. So, if you and Michael decide it's solved, then great, otherwise just leave it out from the PR and we can tackle it later.

rwmcintosh:
> Indeed, in fact I started looking at that, and turns out it requires quite some refactoring.
For example: the IMobiSimulation which is used everywhere.

Ok, remove that issue from the description so this PR is not linked to the issue.

src/MoBi.R/Services/SimulationFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+      public SimulationFactory(ISimulationConfigurationFactory configurationFactory, Core.Domain.Services.ISimulationFactory simulationFactory)
+      {
+         _configurationFactory = configurationFactory;
+         _simulationFactory = simulationFactory;
+      }
+
+      public Simulation CreateSimulation(SimulationConfiguration configuration)
+      {
+         if (string.IsNullOrWhiteSpace(configuration.SimulationName))
+            throw new InvalidOperationException("Simulation name is required");
```
Yuri05, line 30: Are the checks that the simulation name does not contain illegal characters and is not one of the forbidden names done elsewhere (e.g. when setting `configuration.SimulationName`) or should they be performed here?

compare 
https://github.com/Open-Systems-Pharmacology/MoBi/blob/472737a20c7514e99fd44cb3b6957c1de59e9844/src/MoBi.Presentation/DTO/ObjectBaseDTO.cs#L89
and
https://github.com/Open-Systems-Pharmacology/MoBi/blob/472737a20c7514e99fd44cb3b6957c1de59e9844/src/MoBi.Core/Services/ForbiddenNamesRetriever.cs#L35

src/MoBi.R/MinimalImplementations/UserSettings.cs
```diff
@@ -0,0 +1,80 @@
...
+      public ObjectPathType ObjectPathType { get; set; }
+      public string ParameterDefaultDimension { get; set; }
+      public void SaveLayout()
+      {
+         throw new NotImplementedException();
+      }
+
+      public void RestoreLayout()
+      {
+         throw new NotImplementedException();
```
Yuri05, line 60: same question

Yuri05 (approved):

PR #1922, opened 2025-06-03
1912 missing functionalities in c service layer needed to support r vignette use cases

### LINKED ISSUE #1912: Missing functionalities in C# service layer needed to support R vignette use cases
The current implementation of ProjectTask and SimulationCreateTask covers many of the core operations, but the following R-side functionalities are not yet supported in C#:

Get the list of parameter identifications

Get observed data as DataSet objects

Create an IndividualBuildingBlock from R

Load a module from a .pkml file

benjaperez1983:
Fixes #1912 #2019 #2020
# Description

Get the list of parameter identifications
Create simulations from PKML or project modules
R layer: load PKML modules + simple model-edit helpers

Get observed data as DataSet objects

Create an IndividualBuildingBlock from R

Load a module from a .pkml file

## Type of change

# How Has This Been Tested?

- [x] Integration tests
- [ ] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/tests/MoBi.R.Tests/Services/SimulationTaskSpecs.cs
+++ b/tests/MoBi.R.Tests/Services/SimulationTaskSpecs.cs
@@ -59,7 +59,7 @@ namespace MoBi.R.Tests.Services
          _moduleTask = Api.GetModuleTask();
          var projectFile = DataTestFileFullPath("SampleProject.mbp3");
          _project = _projectTask.LoadProject(projectFile);
-         LoadSampleProject();
+         LoadSampleProject(projectFile);
       }
 
       protected override void Because()


src/MoBi.R/Services/MoBiRIntegrationService.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace MoBi.R.Services
+{
+   public interface IMoBiRIntegrationService
+   {
+      IndividualBuildingBlock CreateIndividual(string name);
+      void SetParameterValue(ParameterValuesBuildingBlock pvBB, string parameterName, double newValue);
+   }
+
+   public class MoBiRIntegrationService : IMoBiRIntegrationService
```
rwmcintosh, line 18: We usually use Services as the namespace, but the objects and interfaces end with `Task`.

What's the plan for this one, because it looks like it's just going to be a set of unrelated one-off methods, but I'd still favor separate tasks that implement few public methods instead of one that does a lot of unrelated stuff.

src/MoBi.R/Services/MoBiRIntegrationService.cs
```diff
@@ -0,0 +1,32 @@
...
+   {
+      public IndividualBuildingBlock CreateIndividual(string name)
+      {
+         return new IndividualBuildingBlock { Name = name };
+      }
```
rwmcintosh, line 23: Most of the time when you're doing something like this for `IndividualBuildingBlock` there should be something like it for `ExpressionProfileBuildingBlock` as well.

Also, can you verify what's being asked for in this use case? This is an empty building block, and in MoBi an `IndividualBuildingBlock` can't be modified (except the values of individual parameters). So is the user looking for us to create a new PKSim `IndividualBuildingBlock`?

src/MoBi.R/Services/MoBiRIntegrationService.cs
```diff
@@ -0,0 +1,32 @@
...
+
+      public void SetParameterValue(ParameterValuesBuildingBlock pvBB, string parameterName, double newValue)
+      {
+         var parameter = pvBB.FirstOrDefault(p => p.Name == parameterName);
+         if (parameter != null)
+            parameter.Value = newValue;
+      }
```
rwmcintosh, line 30: I bet there is a requirement for the same operation with `InitialConditions`. Can you verify?

src/MoBi.R/Services/ProjectTask.cs
```diff
@@ -23,6 +25,8 @@ public interface IProjectTask
       Module ModuleByName(MoBiProject moBiProject, string name);
       IndividualBuildingBlock IndividualBuildingBlockByName(MoBiProject moBiProject, string name);
       List<ExpressionProfileBuildingBlock> ExpressionProfileBuildingBlocksByName(MoBiProject moBiProject, params string[] names);
+      IReadOnlyList<DataRepository> AllObservedDataSets(MoBiProject project);
+      IReadOnlyList<ParameterIdentification> AllParameterIdentifications(MoBiProject project);
```
rwmcintosh, line 29: Probably should also be created for `SensitivityAnalyses`

tests/MoBi.R.Tests/Services/SimulationTaskSpecs.cs
```diff
@@ -1,106 +1,100 @@
...
+
       public override void GlobalContext()
       {
          base.GlobalContext();
          sut = Api.GetSimulationTask();
+         _projectTask = Api.GetProjectTask();
+         _moduleTask = Api.GetModuleTask();
+      }
+
+      protected void LoadSampleProject()
```
rwmcintosh, line 29: I'd give the file name as an argument here so that this can be used again in the future with another file.

rwmcintosh (changes requested):

src/MoBi.R/Services/ParameterValuesTask.cs
```diff
@@ -0,0 +1,20 @@
...
+
+   public class ParameterValuesTask : IParameterValuesTask
+   {
+      public void SetParameterValue(ParameterValuesBuildingBlock pvBB, string parameterName, double newValue)
+      {
+         var parameter = pvBB.FirstOrDefault(p => p.Name == parameterName);
+         if (parameter != null)
+            parameter.Value = newValue;
+      }
+   }
```
Yuri05, line 19: What is the use case for this function? Usually a parameter should be set by the **full path** and not just by name

PR #1947, opened 2025-07-09
#1929 typing container disabled when selecting parent path

### LINKED ISSUE #1929: Typing container name not possible after choosing parent path
In the Create Container dialogue: If you first choose the Parent Container instead of typing the name, it's not possible to enter a name for the container anymore. Since no name has been entered the `...` button doesn't show either.

Image

benjaperez1983:
Fixes #1929 

# Description

 In the Create Container dialogue: If you first choose the Parent Container instead of typing the name, it's not possible to enter a name for the container anymore. Since no name has been entered the ... button doesn't show either.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.UI/Views/EditContainerView.cs
+++ b/src/MoBi.UI/Views/EditContainerView.cs
@@ -15,7 +15,6 @@ using OSPSuite.Presentation.Extensions;
 using OSPSuite.Presentation.Views;
 using OSPSuite.UI.Controls;
 using OSPSuite.UI.Extensions;
-using OSPSuite.Utility.Extensions;
 using ToolTips = MoBi.Assets.ToolTips;
 
 namespace MoBi.UI.Views
@@ -174,7 +173,6 @@ namespace MoBi.UI.Views
 
       public bool ContainerPropertiesEditable
       {
-         get => cbContainerType.Enabled;
          set
          {
             if (_readOnly && value)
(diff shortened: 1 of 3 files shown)


src/MoBi.UI/Views/EditContainerView.cs
```diff
@@ -144,8 +144,8 @@ private void initNameControl(ContainerDTO dto)
          var isInit = dto.Name.IsNullOrEmpty();
```
rwmcintosh, line 144: I don't think we need this `isInit` flag. Looking at the before code, it looks wrong. We mark the edit as readonly = false, then enabled = false. 

It seems to be working because there's another method that enables the control `ContainerPropertiesEditable` after.

rwmcintosh: In fact, let's remove this method and just use `ContainerPropertiesEditable`

Then in the presenter we can use the `_isNewEntity` flag and always explicitly enable or disable the naming and other controls

src/MoBi.UI/Views/EditContainerView.cs
```diff
@@ -197,6 +188,17 @@ public bool ContainerPropertiesEditable
...
       }
 
+      public bool IsNameEditable
+      {
+         get => _isNameEditable;
+         set
+         {
+            _isNameEditable = value;
+            btName.Enabled = value;
+            btName.ReadOnly = !value;
```
rwmcintosh, line 198: we still have to manage the enable and visible for `editNameButton`

And we don't keep a backing property in the case of `ContainerPropertiesEditable`. It looks like we just use the Enabled property of one of the controls.

PR #2046, opened 2025-08-21
#2045 disable items on simulation run

### LINKED ISSUE #2045: 12.1 Confusing LockUI/ProgressBar behaviour when running simulations #3163
Related to PK Sim -> https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3163

benjaperez1983:
Fixes #2045

# Description

Some items from the UI Should be disabled when a simulation is running. 

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):
@PavelBal  as you can see on the screenshot, the Edit button is not disabled (as you originally asked for).
The reason for this, is that if we do that, the user is not able to open the simulation while running (since the menu is disabled, and double clicking on it just clicks on the menu) 

Let me know if you agree with this please.
Thanks!

### DIFF
--- a/src/MoBi.Core/Commands/SynchronizeParameterValueCommand.cs
+++ b/src/MoBi.Core/Commands/SynchronizeParameterValueCommand.cs
@@ -1,10 +1,10 @@
-﻿using OSPSuite.Core.Commands.Core;
+﻿using MoBi.Assets;
 using MoBi.Core.Domain.Model;
-using OSPSuite.Core.Domain;
-using OSPSuite.Core.Domain.Builder;
-using MoBi.Assets;
 using MoBi.Core.Domain.Services;
 using OSPSuite.Assets;
+using OSPSuite.Core.Commands.Core;
+using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
 
 namespace MoBi.Core.Commands
 {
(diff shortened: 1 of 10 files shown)


rwmcintosh (changes requested):

src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
```diff
@@ -216,6 +223,8 @@ public void Handle(SimulationRunFinishedEvent eventToHandle)
          if (!_simulation.Equals(eventToHandle.Simulation))
             return;
 
+         _view.SetParametersTabEnabled(true);
+
```
Yuri05, line 227: why? if we disable parameters tab during a simulation run and then add obsdata while sim is till running, it will be enabled again(?)

src/MoBi.Presentation/MenusAndBars/ContextMenus/ContextMenuForSimulation.cs
```diff
@@ -204,6 +205,7 @@ private IMenuBarItem createDeleteItem(IMoBiSimulation simulation)
       private IMenuBarItem createDeleteAllResultsItem(IMoBiSimulation simulation)
       {
          return CreateMenuButton.WithCaption(AppConstants.MenuNames.DeleteAllResults)
+            .WithEnabled(!_simulationRunner.IsSimulationRunning(simulation))
```
msevestre, line 208: What about adding an  `IsSimulationIdle` or sthg like this becuase all the time we are doing !xxx.IsSimulationRunning`? Would it maake sense?

src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
msevestre: yes Idon't get this either

msevestre (changes requested):

PR #2072, opened 2025-09-02
Fixes #2068

### LINKED ISSUE #2068: Pk-Sim module -> Try to add reaction parameter -> Cancel -> Try again - error
1. Open a PK-Sim module (with reactions)
2. Open any reaction
3. Try to add new parameter -> "This will convert..."

Image

4. Click **NO**
5. Try the same:

Image

rwmcintosh:
Fixes #2068 add parameter to reaction error

# Description
We should not be adding the parameter to the parent container except in the following command. Here the new parameter is added to the parent before the command is executed.

There was never a problem before because the actions were atomic, but now the user can change their mind before the command is run, but after the task is run.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/MoBi.Presentation/Tasks/Interaction/InteractionTaskForParameter.cs
```diff
@@ -85,7 +85,6 @@ public InteractionTasksForParameter(
       public override IParameter CreateNewEntity(IContainer parent)
       {
          var parameter = base.CreateNewEntity(parent)
-            .WithParentContainer(parent)
```
Yuri05, line 88: and when is the parameter now added to the parent container?

rwmcintosh:
> why do we remoev the stuff for distributed parametr?

It's not used anywhere.

PR #2118, opened 2025-09-24
2099 parameter values bb new parameter value   relative paht does not make sense

### LINKED ISSUE #2099: Parameter Values BB: New Parameter Value -> "Relative paht" does not make sense
When selecting "New Parameter Value" for a Parameter Values BB, the parameters tree on the left is used to select parameters from the different structures. For the PV BB, relative paths do not make any sense. Only absolute paths are allowed, and the selection between relative and absolute pahts should be removed, along with the "Local reference point".

Image

benjaperez1983:
Fixes #2099

# Description

When selecting "New Parameter Value" for a Parameter Values BB, the parameters tree on the left is used to select parameters from the different structures. For the PV BB, relative paths do not make any sense. Only absolute paths are allowed, and the selection between relative and absolute pahts should be removed, along with the "Local reference point".

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):


src/MoBi.UI/Views/SelectReferenceView.cs
```diff
@@ -64,6 +64,12 @@ public bool EnableMultiSelect
          get => _treeView.OptionsSelection.MultiSelect;
       }
 
+      public void SetLayoutItemLocalisationVisible(bool visible) 
+         => layoutItemLocalisation.Visibility = visible ? LayoutVisibility.Always : LayoutVisibility.Never;
+
+      public void SetRelativePathRadioButtonEnabled(bool enabled)
+         => radioGroupReferenceType.Properties.Items.GetItemByValue(ObjectPathType.Relative).Enabled = enabled;
```
rwmcintosh, line 71: There's a static class from DevExpress `LayoutVisibilityConvertor` that handles the boolean conversions like this.

rwmcintosh (approved):

rwmcintosh (commented):
just change the bool -> layoutvisibility

rwmcintosh (approved):

src/MoBi.Presentation/Tasks/Interaction/ParameterValuesTask.cs
```diff
@@ -144,7 +144,8 @@ public IReadOnlyList<ObjectPath> GetNewPaths()
             //and to initialize it with the correct data so that the initial state of the OK button is correct
             modalPresenter.Encapsulate(referenceAtParamValuePresenter);
             referenceAtParamValuePresenter.Init(null, new List<IObjectBase>(), null);
-
+            referenceAtParamValuePresenter.SetLayoutItemLocalisationVisible(false);
+            referenceAtParamValuePresenter.SetRelativePathRadioButtonEnabled(false);
```
Yuri05: Does it make sense to group both new properties into just one?

src/MoBi.Presentation/Presenter/CreateObjectPathsFromReferencesPresenter.cs
```diff
@@ -14,6 +14,8 @@ public interface ICreateObjectPathsFromReferencesPresenter : IPresenter<ICreateO
       IReadOnlyList<ObjectPath> GetAllSelections();
       void AddSelection();
       Size? ModalSize { get; }
+      void SetLayoutItemLocalisationVisible(bool visible);
+      void SetRelativePathRadioButtonEnabled(bool enabled);
```
msevestre: I am wondering about naming properties after the ux element that they represent. I think the fact that it’s a radio button is irrelevant. The presenter should worry about the function being shown or hidden, not how the view is displaying it. In fact, I have no idea what this code is doing without looking at the UI and understanding exactly how it’s organized 

Also agreed with Juri. Why do we need two properties ?

PR #2122, opened 2025-09-30
2067 snapshot based runners seem to only use a single core v13 mapper

### LINKED ISSUE #2067: Snapshot based runners seem to only use a single core
When fixing [this issue](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3229) for population simulations. It seemed like running individual simulations was always using a single core and never truly parallelizing async/await.

For that issue, the population runner does parallelize all the individual simulations, but that doesn't mean that all separate individual simulations are also parallelized. We have to make another adjustment somewhere.

benjaperez1983:
Fixes #2067

# Description

Execute multiple simulations when loaded from snapshot.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):


src/MoBi.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -105,12 +112,13 @@ private async Task<ModelProject> mapToModel(SnapshotProject projectSnapshot, Pro
...
 
+      List<Task<MoBiSimulation>> tasks = [];
       foreach (var simulationSnapshot in projectSnapshot.Simulations)
       {
          try
          {
-            var simulation = await _simulationMapper.MapToModel(simulationSnapshot, simulationContext);
-            addSimulations(project, simulation);
+            tasks.Add(_simulationMapper.MapToModel(simulationSnapshot, simulationContext));
+
```
rwmcintosh: I think this should revert to the original code. We don't want to run the construction of simulations in parallel, only the actual simulation run.

src/MoBi.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -119,6 +127,47 @@ private async Task<ModelProject> mapToModel(SnapshotProject projectSnapshot, Pro
...
+               _logger.AddException(ex);
+            }
+            finally
+            {
+               gate.Release();
+            }
+         });
+
+         await Task.WhenAll(runTasks);
+      }
```
rwmcintosh: Were you able to investigate `IConcurrentSimulationRunner` from core to see if it can be used? It is very similar to this code.

benjaperez1983: I did, and also consulted some Microsoft Documentation.

I used SemaphoreSlim + Task.WhenAll instead of the Paralel.Foreach because our workload is async (RunSimulationAsync). 
IConcurrentSimulationRunner is built on sync Parallel.ForEach (wrapped in Task.Run), which would force blocking on each async call.

The correct approach would be to use Parallel.ForEachAsync for this, which would cover our case, but unfortunately it is not available on this framework (.NET 6+).

https://learn.microsoft.com/en-us/training/modules/implement-asynchronous-tasks/5-run-asynchronous-tasks-parallel
https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.parallel.foreachasync?view=net-9.0

The issue is here:

return Task.Run(() => Parallel.ForEach(data, createParallelOptions(cancellationToken, numberOfCoresToUse),
   datum =>
   {
      cancellationToken.ThrowIfCancellationRequested();
      action(datum, cancellationToken);   // <-- ISSUE HERE, this is not an ASYNC Method or task
      // For sims we'd need: _simulationRunner.RunSimulationAsync(sim)
      // but Parallel.ForEach can't await. We must either:
      //  1) fire-and-forget (BUG: ForEach finishes before tasks do, exceptions lost), or
      //  2) block with .GetAwaiter().GetResult() --> Blocks the Thread, deadlocks risk.
   }), cancellationToken);
   
So the real issue is the loop body call (action(...)) inside Parallel.ForEach. 
With _simulationRunner.RunSimulationAsync(sim) there’s no way to await without either not waiting or blocking.
SemaphoreSlim + Task.WhenAll avoids both of these issues.

Let me know your Thoughts about this.

src/MoBi.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -105,12 +112,15 @@ private async Task<ModelProject> mapToModel(SnapshotProject projectSnapshot, Pro
          NumberOfSimulationsLoaded = 0
       };
 
+      var sims = new List<MoBiSimulation>();
```
rwmcintosh: I guess we could get away without this variable. We only use it to tally the simulations, but we are going to run all the simulations in the project after this step so we could just get it from there.

src/MoBi.Core/Snapshots/Mappers/ProjectMapper.cs
```diff
@@ -119,6 +129,31 @@ private async Task<ModelProject> mapToModel(SnapshotProject projectSnapshot, Pro
...
+            {
+               _logger.AddException(ex);
+            }
+            finally
+            {
+               gate.Release();
+            }
+         });
+
+         await Task.WhenAll(runTasks);
```
rwmcintosh: Could you move it to a method to improve the readability there?

rwmcintosh (approved):
I would just make a couple small changes.

msevestre:
>IConcurrentSimulationRunner is built on sync Parallel.ForEach (wrapped in Task.Run), which would force blocking on each async call.

I don't understand what you mean. Parallel.ForEach blocks on each call?

msevestre:
>The correct approach would be to use Parallel.ForEachAsync for this, which would cover our case, but unfortunately it is not available on this framework (.NET 6+).

That's an interesting question. We are moving to .NET 8 now. Should we move our components when needed or is the implication of this way too high?

rwmcintosh:
> >The correct approach would be to use Parallel.ForEachAsync for this, which would cover our case, but unfortunately it is not available on this framework (.NET 6+).
> 
> 
> 
> That's an interesting question. We are moving to .NET 8 now. Should we move our components when needed or is the implication of this way too high?

At least testing here limits the reach to mobi. @benjaperez1983 can you try it? Do all other assemblies need to be upgraded at the same time?

src/MoBi.Core/Snapshots/Mappers/ProjectMapper.cs
Yuri05: Agree

benjaperez1983:
> IConcurrentSimulationRunner

For what I`ve read about Parallel.ForEach is a **synchronous** API. 
It runs a **sync delegate** on multiple threads and the call doesn’t return until all iterations finish. 
Because the delegate is sync, if I need to call an async method like RunSimulationAsync, I have only two choices:

1) Block the async call so it looks sync:

Parallel.ForEach(sims, opt, sim =>
{
  _simulationRunner.RunSimulationAsync(sim).GetAwaiter().GetResult(); // blocking a worker thread
});


which “forces blocking” (what I meant): each iteration ties up a thread while the async operation is running.

2) Fire-and-forget inside the body (don’t wait), which is incorrect because Parallel.ForEach would finish before the tasks do, and exceptions/cancellation are lost.

Task.Run(() => Parallel.ForEach(...)) doesn’t make the work asynchronous, it just runs our synchronous loop on a thread-pool thread. While it’s running, threads are still blocked.

benjaperez1983:
Converted this to draft, to prevent any merging. Since once I implemented the Parallel.ForEachAsync, I can see other changes need to be made at Core.
So I`d rather discuss them with @rwmcintosh  or all the team before merging.

benjaperez1983:
depends on https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/pull/2659

rwmcintosh (changes requested):
The document submodule change should be reverted

benjaperez1983:
> The document submodule change should be reverted

Good Catch! already reverted it to V13

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -354,7 +354,12 @@ private void addPredefinedMoleculesForImporter(MetaDataCategory metaDataCategory
 
       private IEnumerable<IContainer> allMolecules()
       {
-         return _buildingBlockRepository.MoleculeBlockCollection.SelectMany(buildingBlock => { return buildingBlock.Select(builder => builder); }).DistinctBy(builder => builder.Name);
+         return System.Linq.Enumerable.DistinctBy(
+            _buildingBlockRepository.MoleculeBlockCollection.SelectMany(bb => bb),
+            b => b.Name,
+            StringComparer.Ordinal);
```
rwmcintosh: that should go back to the extension method syntax rather than the static method syntax right? Only change is who is providing the extension?

rwmcintosh (approved):

msevestre (approved):
loooks good. Just one change with the extgension methods not called properly

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
benjaperez1983: Yeah I specified the namespace because of the ambiguous call. 
But reverting it.

PR #2147, opened 2025-10-20
2125 order of module building blocks in simulation

### LINKED ISSUE #2125: Order of module building blocks in simulation
The icon order should match between module explorer and simulation explorer.

Module Explorer:
Image

Simulation Explorer
Image

benjaperez1983:
Fixes #2125

# Description

Order the Module building blocks on the Simulation Explorer as they are on the Module Explorer

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [ ] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
@@ -152,8 +152,7 @@ namespace MoBi.Presentation.Presenter.Main
             var o1 = orderFor(node1.Tag as BuildingBlock);
             var o2 = orderFor(node2.Tag as BuildingBlock);
 
-            var byRank = o1.CompareTo(o2);
-            if (byRank != 0) return byRank;
+            return o1.CompareTo(o2);
          }
 
          return nameComparison(node1, node2);
--- a/src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
@@ -1,6 +1,11 @@
+using System;
+using System.Collections.Generic;
+using System.Drawing;
 using MoBi.Core.Domain.Model;
 using MoBi.Core.Events;
+using MoBi.Presentation.Nodes;
 using OSPSuite.Core.Domain;
+using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Presentation.Core;
 using OSPSuite.Presentation.Nodes;
@@ -8,13 +13,12 @@ using OSPSuite.Presentation.Presenters;
 using OSPSuite.Presentation.Presenters.Classifications;
 using OSPSuite.Presentation.Presenters.ContextMenus;
 using OSPSuite.Presentation.Presenters.Main;
+using OSPSuite.Presentation.Presenters.Nodes;
 using OSPSuite.Presentation.Regions;
 using OSPSuite.Presentation.Services;
 using OSPSuite.Presentation.Views;
 using OSPSuite.Utility.Events;
 using OSPSuite.Utility.Extensions;
-using System.Collections.Generic;
-using System.Drawing;
 using ITreeNodeFactory = MoBi.Presentation.Nodes.ITreeNodeFactory;
 
 namespace MoBi.Presentation.Presenter.Main
@@ -25,6 +29,8 @@ namespace MoBi.Presentation.Presenter.Main
       IListener<SimulationRunFinishedEvent>
 
    {
+      int OrderingComparisonFor(ITreeNode<IWithName> node1, ITreeNode<IWithName> node2);
+      int TryGetOrderingComparison(object parentTag, ITreeNode<IWithName> node1, ITreeNode<IWithName> node2);
    }
 
    public abstract class ExplorerPresenter<TView, TPresenter> : AbstractExplorerPresenter<TView, TPresenter>, IExplorerPresenter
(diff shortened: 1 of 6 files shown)


src/MoBi.UI/Views/SimulationExplorerView.cs
```diff
@@ -32,11 +38,16 @@ private void compareNodeValues(object sender, CompareNodeValuesEventArgs e)
          if (e.Node1.Level == 0)
             e.Result = 0;
 
+         //we need to sort the module nodes exactly as we do on the ModuleExplorerView.cs
+         if (nodeIsModuleNode(e.Node1.ParentNode))
+            e.Result = _moduleExplorerPresenter.OrderingComparisonFor(e.Node1.Tag as ITreeNode<IWithName>, e.Node2.Tag as ITreeNode<IWithName>);
```
rwmcintosh: A better way to do this is to pull up the `OrderingComparisonFor` from the `ModuleExplorerPresenter` into the `ExplorerPresenter` that is common to both. 

Make sure this ordering is only used for the appropriate nodes in both cases.

rwmcintosh (changes requested):

src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
```diff
@@ -41,6 +41,18 @@ public abstract class ExplorerPresenter<TView, TPresenter> : AbstractExplorerPre
       protected readonly IMoBiContext _context;
       private readonly IMultipleTreeNodeContextMenuFactory _multipleTreeNodeContextMenuFactory;
 
+      private static readonly IReadOnlyDictionary<Type, int> _buildingBlockOrder = new Dictionary<Type, int>
```
rwmcintosh: If you use Cache, you can specify how to handle keys not present

```
private static readonly Cache<Type, int> _buildingBlockOrder = new Cache<Type, int>(onMissingKey: x => int.MaxValue)
```

src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
```diff
@@ -167,5 +185,16 @@ private static int nameComparison(ITreeNode<IWithName> a, ITreeNode<IWithName> b
...
+
+      private static int orderFor(BuildingBlock buildingBlock)
+      {
+         if (buildingBlock == null)
+            return int.MaxValue;
+
+         if (_buildingBlockOrder.TryGetValue(buildingBlock.GetType(), out var order))
+            return order;
+
+         return int.MaxValue;
```
rwmcintosh: Then you can just go

```
         if (buildingBlock == null)
            return int.MaxValue;

         return _buildingBlockOrder[buildingBlock.GetType()];
```

because if the key is not found, the `int.Max` is returned

rwmcintosh (changes requested):
One tip to use the Cache because it has some missing key features that you want here.

Also, we did not have tests covering the ordering of building blocks before, but since we are being deliberate about the order now, it makes sense to add them.

benjaperez1983:
> One tip to use the Cache because it has some missing key features that you want here.
> 
> Also, we did not have tests covering the ordering of building blocks before, but since we are being deliberate about the order now, it makes sense to add them.

Agreed, added some for different scenarios. Such as Mixed BBs or even Same BB which might not make sense but better be certain that this will not break anything.

rwmcintosh (approved):

src/MoBi.UI/Views/SimulationExplorerView.cs
```diff
@@ -33,10 +38,15 @@ private void compareNodeValues(object sender, CompareNodeValuesEventArgs e)
             e.Result = 0;
 
          //we do not want to sort the items under the simulation node (i.e. no children). Otherwise, Nodes are sorted alphabetically
-         else if (nodeIsSimulationNode(e.Node1.ParentNode))
+         if (nodeIsSimulationNode(e.Node1.ParentNode))
             e.Result = 0;
+         else if(nodeIsModuleConfigurationNode(e.Node1.ParentNode))
+            e.Result = _simulationExplorerPresenter.OrderingComparisonForModules(e.Node1.Tag as ITreeNode<IWithName>, e.Node2.Tag as ITreeNode<IWithName>);
```
msevestre, line 44: I thinl that the casting should be done in the presenter not in the view. 
What is the type of Tag anyways?

src/MoBi.Presentation/Presenter/Main/ExplorerPresenter.cs
```diff
@@ -106,5 +126,73 @@ public virtual void Handle(SimulationRunFinishedEvent eventToHandle)
       {
          _view.Enabled = true;
       }
+
+      public virtual int OrderingComparisonForModules(ITreeNode<IWithName> node1, ITreeNode<IWithName> node2)
```
msevestre, line 130: @rwmcintosh @benjaperez1983 There is a lot of code just for the sake of sorting. Would it belong into another service somehow? From the ExplorerPresneter, it would be nice to say (_sortService.Sort(node1, ode2) and be done with it. Like we do with the ContextMenu./ Give me the contextmeny for my node and basta

rwmcintosh: Yeah, that can easily be a service. Same amount of code though.

src/MoBi.UI/Views/SimulationExplorerView.cs
rwmcintosh: > What is the type of Tag anyways?

`object`

msevestre:
Yeah it's not about amount of code but more about do. I need to ever look
at this again once implemented. Anyways does not have to be done now but
maybe to keep in mind for another time

Kind Regards,

Michael


--


*Michael Sevestre*

*Design2Code Inc.*


Email ***@***.***

Website www.design2code.ca

Sent via Superhuman ***@***.***>

PR #2174, opened 2025-11-13
1964 parameters from reactions are not listed in the possible referenced objects fix

### LINKED ISSUE #1964: Parameters from Reactions are not listed in the "Possible referenced objects"
[HallowModel_2017b_121.zip](https://github.com/user-attachments/files/21295433/HallowModel_2017b_121.zip)

See attached project. Some reactions have global parameters, e.g. `Sodium_Intake|Na_intake_rate`.

These parameters are not shown in any list of possible referenced objects when creating a formula:

Image

Expected behavior: reactions are shown in the root node, and can be expanded to show global parameters, similarly like in the simulations tree view:

Image

Same behavior in v11.

benjaperez1983:
Fixes #1964

# Description

Merge by module

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

image

# Questions (if appropriate):


rwmcintosh (changes requested):
Undo the submodule changes

src/MoBi.Presentation/Presenter/SelectReferenceAtParameterPresenter.cs
```diff
@@ -36,7 +36,7 @@ protected override void AddSpecificInitialObjects()
          if(_addTime)
             AddTimeReference();
          AddSpatialStructures();
-         AddReactions();
```
rwmcintosh, line 34: I would just do one more thing before merging this. Can you check if this edition of the reactions should be used at all? I think all places where we are using `AddReactions` should be replaced with the new one.

I think it should be replaced everywhere. Eg, in this view where the tree view is built for molecule builder. 

Image

In other words, I think the old version `AddReactions` should be removed and this version used instead. And then I would just rename it `AddReactions`

rwmcintosh (approved):

src/MoBi.Presentation/Presenter/SelectReferencePresenterBase.cs
```diff
@@ -447,12 +459,8 @@ private IEnumerable<MoleculeBuilder> allMolecules
 
       protected void AddReactions()
       {
-         _view.AddNodes(getReactions.MapAllUsing(_referenceMapper));
-      }
-
-      private IEnumerable<ReactionBuilder> getReactions
-      {
-         get { return _buildingBlockRepository.ReactionBlockCollection.SelectMany(x => x).OrderBy(x => x.Name); }
+         var nodes = _buildingBlockRepository.ReactionBlockCollection.Select(x => _referenceMapper.MapFrom(x).WithText(x.DisplayName));
```
msevestre: @benjaperez1983 @rwmcintosh should not we always set the text to be the display name  in the referenceMapper? Why are we doing sthg special just for reactions?

msevestre (changes requested):
question for me. Feels like the fix is in the wrong place

rwmcintosh:
> question for me. Feels like the fix is in the wrong place

Well, the fix is not only the name, previously these were not grouped at all and all reactions from all modules were in a flat list.

src/MoBi.Presentation/Presenter/SelectReferencePresenterBase.cs
benjaperez1983: I Believe this is done as it is for AddSpatialStructores on Line 267:

"var nodes = spatialStructures.Select(x => _referenceMapper.MapFrom(x).WithText(x.DisplayName));"

Is there any particular reason why not to do it this way given then?

msevestre: >Is there any particular reason why not to do it this way given then?

what do you mean?

rwmcintosh: He's suggesting we move the naming into the `IObjectBaseDTOToReferenceNodeMapper`. See the `MapFrom` and add a check if the target is a building block, then we should use `DisplayName` always.

```
      public ITreeNode MapFrom(ObjectBaseDTO objectBaseDTO)
      {
         var node = new ReferenceNode(objectBaseDTO)
         {
            Icon = objectBaseDTO.Icon,
            GetChildren = x => _getChildren(x).Select(MapFrom).ToList()
         };

         if (objectBaseDTO.ObjectBase is BuildingBlock buildingBlock)
            node.Text = buildingBlock.DisplayName;
         
         return node;
      }
```

msevestre: yes that's what I am suggesting but what is @benjaperez1983 suggesting lol. If he suggests the same than yes

rwmcintosh: There's another location in this file where the `WithText` extension is used. @benjaperez1983  They can both be removed if the mapper uses the `DisplayName` internally.

benjaperez1983: Exactly, it is either we do it for every case, or for none, to be consistent.
Will do it for all cases then and resolve the case.

benjaperez1983:
@msevestre I believe all the comments are resolved. 
Whenever possible please take a look. 
Thanks

PR #2176, opened 2025-11-17
Fixes #2166 Simulation cannot be closed due to invalid output intervals

### LINKED ISSUE #2166: Simulation cannot be closed due to invalid output intervals
Simulation view becomes inactive and cannot be closed due to invalid OutputIntervals

how to reproduce:
- enter an invalid output interval, e.g. 48h to 24h
- ignore the red crosses indicating the error and run the simulation by pressing F5
- view will switch from Parameters tab to Results tab and error message will pop up
- returning back to parameters tab is not possible. neither is closing the simulation

now 3 possibilities:
1) try to close the simulation by e.g. "close all documents" --> watch the blue circle of death for several minutes --> simulation view is still blocked
2) close project (not MoBi) --> the simulation view will stay --> reopening project doesn't help
3) save project, exit mobi and reopen project --> all good, change the output interval

rwmcintosh:
Fixes #2166

# Description
There's some rework of the validation and some additional validation done. Also, we do not show results tab on a simulation unless the simulation was completed successfully.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditOutputSchemaPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditOutputSchemaPresenter.cs
@@ -83,6 +83,8 @@ namespace MoBi.Presentation.Presenter
 
       public void SetParameterUnit(IParameterDTO parameterDTO, Unit newUnit, OutputIntervalDTO outputIntervalDTO)
       {
+         // The UI validation of interval start and end does not evaluate for unit changes.
+         // This is a programmatic validation of the proposed new unit
          if (!intervalIsValidInNewUnit(parameterDTO, newUnit, outputIntervalDTO))
          {
             displayIntervalErrorMessage();
--- a/src/MoBi.Presentation/Presenter/EditOutputSchemaPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditOutputSchemaPresenter.cs
@@ -1,4 +1,5 @@
-﻿using OSPSuite.Assets;
+﻿using MoBi.Assets;
+using OSPSuite.Assets;
 using OSPSuite.Utility.Collections;
 using OSPSuite.Utility.Extensions;
 using MoBi.Core.Domain.Model;
@@ -9,6 +10,7 @@ using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.Core.Events;
+using OSPSuite.Core.Services;
 using OSPSuite.Presentation.DTO;
 using OSPSuite.Presentation.Presenters;
 using OSPSuite.Utility.Events;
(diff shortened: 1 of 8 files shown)


src/MoBi.Presentation/Presenter/EditSimulationPresenter.cs
```diff
@@ -230,7 +230,7 @@ public void Handle(SimulationRunFinishedEvent eventToHandle)
 
          _view.SetParametersTabEnabled(true);
 
-         if (!_view.ShowsResults)
+         if (!_view.ShowsResults && eventToHandle.Success)
```
rwmcintosh, line 233: I added this first to only prevent the tab from switching when the simulation fails. The symptom was that the validation error is being shown in the parameter screen, and you cannot switch the tabs when that happens, but you can programmatically switch. When that happens you cannot switch back.

src/MoBi.UI/Views/EditOutputSchemaView.cs
```diff
@@ -33,7 +33,7 @@ public EditOutputSchemaView()
 
       public override void InitializeBinding()
       {
-         _gridViewBinder = new GridViewBinder<OutputIntervalDTO>(gridViewIntervals) { BindingMode = BindingMode.OneWay, ValidationMode = ValidationMode.LeavingRow };
+         _gridViewBinder = new GridViewBinder<OutputIntervalDTO>(gridViewIntervals) { BindingMode = BindingMode.OneWay };
```
rwmcintosh, line 36: It's a bit less convenient without `LeavingRow`, but with it, you can create the invalid state, and we should prevent the parameter from being changed to create an invalid output interval.

src/MoBi.Presentation/Presenter/EditOutputSchemaPresenter.cs
```diff
@@ -77,14 +81,60 @@ public void RemoveOutputInterval(OutputIntervalDTO outputIntervalDTO)
          bindToView();
       }
 
-      public void SetParameterUnit(IParameterDTO parameterDTO, Unit newUnit)
+      public void SetParameterUnit(IParameterDTO parameterDTO, Unit newUnit, OutputIntervalDTO outputIntervalDTO)
```
rwmcintosh, line 84: The UI validation doesn't work as well with the combobox, so in the presenter we test for invalid unit changes and create a dialog to inform the user.

src/MoBi.UI/Views/EditOutputSchemaView.cs
Yuri05: With this change, what will be the validation mode? LeavingCell?

rwmcintosh: Yeah, the standard, when you leave the cell

Yuri05: Meaning that we will need to edit start/end time always in the "correct" order.
E.g. if I add a second time interval and would like to change it from [0h .. 24h] to e.g. [25h .. 50h] - I would need to change the **end time** first, and then a start time. Little bit annoying, but I can live with it.

Image

rwmcintosh: Yes, that's right. A little less convenient, but better than finding this other bug where the user could not do anything but kill the process.

Yuri05: Agree

Yuri05 (approved):

PR #2220, opened 2026-01-14
Update builds to 12 3

rwmcintosh:


.github/workflows/build-and-test_12.3.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 env:
   MAJOR: 12
-  MINOR: 2
+  MINOR: 3
   RUN: ${{ github.run_number }}
```
Yuri05: the RUN environment variable is not used and the line can be removed

.github/workflows/build-and-test.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           dotnet restore
 
       - name: Build
-        run: dotnet build MoBi.sln /p:Version=12.2.9999
+        run: dotnet build MoBi.sln /p:Version=12.3.9999
```
Yuri05, line 28: why not using `MAJOR` and `MINOR` from the environment here?

.github/workflows/build-and-test_12.3.yml
rwmcintosh: I'm removing all of them

.github/workflows/build-and-test.yml
rwmcintosh: removing in favor of just putting 12.3 all over. That way we can search for all instances more easily

Yuri05 (approved):

PR #2236, opened 2026-01-20
Fixes #2186 Rename reaction on load should update formula paths

### LINKED ISSUE #2186: Rename reaction on load should update formula paths
If you load a reaction and a reaction with the same name is present, renaming is forced,  also for the formulas with the same names, but the paths in the formulas are not renamed accordingly

rwmcintosh:
Fixes #2186 Rename reaction on load should update formula paths

# Description
When loading an object from PKML and a rename is forced, the contained formulas should update paths where the path contains the original name.

I did not limit this to only acting on reactions even though the ticket specifically refers to reactions.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [x] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

### DIFF
--- a/src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
+++ b/src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
@@ -266,16 +266,8 @@ namespace MoBi.Presentation.Tasks.Interaction
          if (string.Equals(oldName, newName))
             return;
 
-         switch (objectBase)
-         {
-            case IUsingFormula usingFormula:
-               if(usingFormula.Formula != null)
-                  correctFormulaPathsForRename(oldName, newName, usingFormula.Formula);
-               break;
-            case IFormula formula:
-               formula.ObjectPaths.Each(x => x.Replace(oldName, newName));
-               break;
-         }
+         if (objectBase is IUsingFormula usingFormula) 
+            usingFormula.Formula?.ObjectPaths.Each(x => x.Replace(oldName, newName));
 
          if (!(objectBase is IContainer container))
             return;
--- a/src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
+++ b/src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
@@ -266,22 +266,24 @@ namespace MoBi.Presentation.Tasks.Interaction
          if (string.Equals(oldName, newName))
             return;
 
-         if (objectBase is IUsingFormula usingFormula) 
-            usingFormula.Formula?.ObjectPaths.Each(x => x.Replace(oldName, newName));
+         var objects = objectsToCorrect(objectBase);
 
-         if (!(objectBase is IContainer container))
-            return;
+         objects.Each(x =>
+         {
+            if (x is IUsingFormula usingFormula)
+               usingFormula.Formula?.ObjectPaths.Each(path => path.Replace(oldName, newName));
+         });
+      }
 
+      private static List<IObjectBase> objectsToCorrect(IObjectBase objectBase)
+      {
          var children = new List<IObjectBase>();
 
-         // Correct child parameters of the container
-         children.AddRange(container.GetAllChildren<IParameter>());
-
-         // Recurse containers to check their parameters
-         children.AddRange(container.GetAllChildren<IContainer>());
-
-         // Distinct to remove duplicates (e.g., parameters that are also containers)
-         children.Distinct().Each(x => correctFormulaPathsForRename(oldName, newName, x));
+         if (objectBase is IContainer container)
+            children.AddRange(container.GetAllChildrenAndSelf<IEntity>());
+         else
+            children.Add(objectBase);
+         return children;
       }
 
       /// <summary>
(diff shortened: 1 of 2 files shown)


src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
```diff
@@ -252,15 +254,54 @@ public virtual IMoBiCommand AddTo(TChild childToAdd, TParent parent, IBuildingBl
...
+         }
+
+         if (!(objectBase is IContainer container))
+            return;
+
+         // Correct child parameters of the container
+         container.GetAllChildren<IParameter>().Each(x => correctFormulaPathsForRename(oldName, newName, x));
+
+         // Recurse containers to check their parameters
+         container.GetAllChildren<IContainer>().Each(x => correctFormulaPathsForRename(oldName, newName, x));
```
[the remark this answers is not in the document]
rwmcintosh: Implemented

src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
```diff
@@ -252,15 +254,54 @@ public virtual IMoBiCommand AddTo(TChild childToAdd, TParent parent, IBuildingBl
...
+      /// <param name="objectBase">The object to recurse</param>
+      private void correctFormulaPathsForRename(string oldName, string newName, IObjectBase objectBase)
+      {
+         if (string.Equals(oldName, newName))
+            return;
+
+         switch (objectBase)
+         {
+            case IUsingFormula usingFormula:
+               correctFormulaPathsForRename(oldName, newName, usingFormula.Formula);
```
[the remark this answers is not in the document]
rwmcintosh: implemented

src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
```diff
@@ -252,15 +254,60 @@ public virtual IMoBiCommand AddTo(TChild childToAdd, TParent parent, IBuildingBl
...
+
+         switch (objectBase)
+         {
+            case IUsingFormula usingFormula:
+               if(usingFormula.Formula != null)
+                  correctFormulaPathsForRename(oldName, newName, usingFormula.Formula);
+               break;
+            case IFormula formula:
+               formula.ObjectPaths.Each(x => x.Replace(oldName, newName));
+               break;
```
Yuri05: I think this case is wrong.
Why should renaming the **formula** also rename its object paths???

Yuri05: It's correct if triggered from the previous case (formula of renamed IUsingFormula). 
So I think smthg like below should be enough:

```
switch (objectBase)
         {
            case IUsingFormula usingFormula:
               if(usingFormula.Formula != null)
                  formula.ObjectPaths.Each(x => x.Replace(oldName, newName));
               break;
        }
```

Yuri05: Or do we have a use case, where we add just a formula which **does not belong to any IUsingFormula** and the formula is being renamed during add and we must rename its object paths?

rwmcintosh: Well, this is just part of the recursion. So your suggestion would just remove one level. But functionally there's no difference.

Terminating conditions of the recursion are if the old and new name are the same, if the objectBase is a Formula and when the object base is not a container.

I'll implement it that way

src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
```diff
@@ -252,15 +254,60 @@ public virtual IMoBiCommand AddTo(TChild childToAdd, TParent parent, IBuildingBl
...
+         var children = new List<IObjectBase>();
+
+         // Correct child parameters of the container
+         children.AddRange(container.GetAllChildren<IParameter>());
+
+         // Recurse containers to check their parameters
+         children.AddRange(container.GetAllChildren<IContainer>());
+
+         // Distinct to remove duplicates (e.g., parameters that are also containers)
+         children.Distinct().Each(x => correctFormulaPathsForRename(oldName, newName, x));
```
Yuri05: Too much recursion here.
If we have object hierarchy A|B|C|D then `correctFormulaPathsForRename(D)` will be triggered 4 times!
(`GetAllChildren()` returns both direct and indirect children of a container).
Also why this distinguishing between parameters and containers?

I think it should be just something like
```
children.AddRange(container.GetChildren()); 
children.Each(x => correctFormulaPathsForRename(oldName, newName, x));
```

Or do I miss something (again)? :)

rwmcintosh: ah, yes, in that case it's too many recursive calls. In fact recursion should be removed here and in the tags correction

src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
```diff
@@ -252,28 +254,67 @@ public virtual IMoBiCommand AddTo(TChild childToAdd, TParent parent, IBuildingBl
...
+            switch (x)
+            {
+               case TransportBuilder transportBuilder:
+                  correctTransportBuilderTags(oldName, newName, transportBuilder);
+                  break;
+               case ApplicationBuilder applicationBuilder:
+                  correctApplicationBuilderMolecules(oldName, newName, applicationBuilder);
+                  break;
+            }
+         });
```
Yuri05, line 317: even better now :)
Just the summary of correctPaths and correct Tags are now outdated (Remove "Recursively")

Yuri05 (approved):

src/MoBi.Presentation/Tasks/Interaction/InteractionTaskFor.cs
```diff
@@ -252,28 +254,67 @@ public virtual IMoBiCommand AddTo(TChild childToAdd, TParent parent, IBuildingBl
...
+
+         if (objectBase is IContainer container)
+            children.AddRange(container.GetAllChildrenAndSelf<IEntity>());
+         else
+            children.Add(objectBase);
+         return children;
+      }
+
+      /// <summary>
+      /// Updates name tags in all sub-containers from the specified old name to the new name before it is added to the project
```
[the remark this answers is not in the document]
rwmcintosh: Corrected

PR #2289, opened 2026-03-16
2280-renaming-molecules-to-existing-molecules-in-the-project-validati…

### LINKED ISSUE #2280: Renaming molecules to existing molecules in the project validation fails
It should be possible to rename molecules so long as they are uniquely named in the module. It should not be required that they are uniquely named in the project

In the molecules BB select a molecule and rename to the name of a molecule in another module. It should be allowed, but currently not.

benjaperez1983:
Fixes #2280

# Description

Renaming molecules to existing moleculse inside the project validation.

## Type of change

# How Has This Been Tested?

- [ ] Integration tests
- [x] Unit tests
- [ ] Manual tests 
- [ ] No tests required

# Reviewer checklist

# Questions (if appropriate):

## Summary by CodeRabbit

* **Bug Fixes**
  * Forbidden-name checks now respect module boundaries so identical names in different modules don't conflict.
  * Initial-condition names are correctly scoped to their module and are no longer incorrectly excluded.
  * Molecule-related forbidden-name collection now considers names defined with the molecule's building block, improving conflict detection.

* **Tests**
  * Added and updated tests covering module-scoped naming for molecules, reactions, and parameters to validate the corrected behavior.

### DIFF
--- a/src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
+++ b/src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
@@ -63,9 +63,15 @@ namespace MoBi.Core.Services
          var nameHash = new HashSet<string>();
          addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInProject());
          addNamesToHash(nameHash, allParametersFromMoleculesInProject());
-         addNamesToHash(nameHash, allMoleculeNamesFromInitialConditionsInModule(moleculeBuilder));
+         addNamesToHash(nameHash, allMoleculeNamesFromBuildingBlock(moleculeBuilder));
          addNamesToHash(nameHash, allReactionNamesFromProject());
          return nameHash;
+      }
+
+      private static IReadOnlyList<string> allMoleculeNamesFromBuildingBlock(MoleculeBuilder builder)
+      {
+         var buildingBlock = builder.BuildingBlock as MoleculeBuildingBlock;
+         return buildingBlock == null ? Enumerable.Empty<string>().ToList() : buildingBlock.AllNames();
       }
 
       public IEnumerable<string> For(ReactionBuilder moleculeBuilder)
@@ -122,7 +128,7 @@ namespace MoBi.Core.Services
 
       private IEnumerable<string> allReactionNamesFromProject() => allNamesFrom<MoBiReactionBuildingBlock, ReactionBuilder>(_buildingBlockRepository.ReactionBlockCollection);
 
-      private IEnumerable<string> allNamesFrom<TBuildingBlock, TBuilder>(IEnumerable<TBuildingBlock> buildingBlock) where TBuildingBlock : IBuildingBlock<TBuilder> where TBuilder : class, IBuilder => 
+      private IEnumerable<string> allNamesFrom<TBuildingBlock, TBuilder>(IEnumerable<TBuildingBlock> buildingBlock) where TBuildingBlock : IBuildingBlock<TBuilder> where TBuilder : class, IBuilder =>
          buildingBlock.SelectMany(x => x.All()).Select(x => x.Name);
 
       private IEnumerable<string> allLocalParametersFromMoleculesInProject() => allParametersFromMoleculesInProject(ParameterBuildMode.Local);
@@ -140,61 +146,6 @@ namespace MoBi.Core.Services
          var nameHash = new HashSet<string>();
          _buildingBlockRepository.MoleculeBlockCollection
             .SelectMany(x => x.All()).Each(m => addNamesToHash(nameHash, allParameterNamesFrom(m, x => x.BuildMode == buildMode)));
-
-         return nameHash;
-      }
-
-      /// <summary>
-      ///    Retrieves all the molecule names from start values where a molecule is named like the molecule builder.
-      ///    The MoleculeBuilders Name is not returned
-      /// </summary>
-      /// <param name="moleculeBuilder">The molecule builder for which forbidden names are retrieved.</param>
-      /// <returns>molecule names that are forbidden for the moleculebuilder</returns>
-      /// <remarks>
-      ///    This is necessary when a molecule was removed and another molecule should be renamed. It
-      ///    should not be possible to rename to the removed named to prevent double definitions of InitialConditions
-      /// </remarks>
-      private IEnumerable<string> allMoleculeNamesFromInitialConditions(MoleculeBuilder moleculeBuilder)
-      {
-         var nameHash = new HashSet<string>();
-         // We need to retrieve Names from here if a removed Molecule is still in MSV
-         // to prevent double definitions in InitialConditions
-         getAllInitialConditionsFromBuildingBlocksFor(moleculeBuilder)
-            .Select(x => x.MoleculeName).Distinct()
-            .Where(x => !x.Equals(moleculeBuilder.Name))
-            .Each(x => nameHash.Add(x));
-
-         return nameHash;
-      }
-
-      private IEnumerable<InitialCondition> getAllInitialConditionsFromBuildingBlocksFor(MoleculeBuilder builder)
-      {
-         var builderName = builder.Name;
-         return _buildingBlockRepository.InitialConditionBlockCollection
-            .Where(x => x.Any(msv => msv.MoleculeName.Equals(builderName)))
-            .SelectMany(x => x.All());
-      }
-
-      /// <summary>
-      ///    Retrieves molecule names from initial conditions scoped to the molecule builder's module only.
-      ///    Same-type entities (molecules) merge across modules during model construction,
-      ///    so only initial conditions within the same module should be checked.
-      /// </summary>
-      private IEnumerable<string> allMoleculeNamesFromInitialConditionsInModule(MoleculeBuilder moleculeBuilder)
-      {
-         var module = moleculeBuilder.BuildingBlock?.Module;
-         if (module == null)
-            return allMoleculeNamesFromInitialConditions(moleculeBuilder);
-
-         var nameHash = new HashSet<string>();
-         var builderName = moleculeBuilder.Name;
-
-         module.InitialConditionsCollection
-            .Where(x => x.Any(ic => ic.MoleculeName.Equals(builderName)))
-            .SelectMany(x => x)
-            .Select(x => x.MoleculeName).Distinct()
-            .Where(x => !x.Equals(builderName))
-            .Each(x => nameHash.Add(x));
 
          return nameHash;
       }
--- a/src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
+++ b/src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
@@ -1,13 +1,12 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using OSPSuite.Utility.Extensions;
-using OSPSuite.Utility.Visitor;
 using MoBi.Core.Domain.Model;
-using MoBi.Core.Domain.Repository;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Repositories;
+using OSPSuite.Utility.Extensions;
+using OSPSuite.Utility.Visitor;
 using IBuildingBlockRepository = MoBi.Core.Domain.Repository.IBuildingBlockRepository;
 
 namespace MoBi.Core.Services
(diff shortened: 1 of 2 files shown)


src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -61,10 +61,23 @@ public ForbiddenNamesRetriever(IBuildingBlockRepository buildingBlockRepository,
...
+         else
+         {
+            addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInProject());
+            addNamesToHash(nameHash, allParametersFromMoleculesInProject());
+            addNamesToHash(nameHash, allMoleculeNamesFromInitialConditions(moleculeBuilder));
+            addNamesToHash(nameHash, allReactionNamesFromProject());
+         }
+
          return nameHash;
       }
```
Yuri05, line 68: What are the use cases where `module `is `null `here?
Just asking myself if the whole `else`-part is required at all.

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -61,10 +61,23 @@ public ForbiddenNamesRetriever(IBuildingBlockRepository buildingBlockRepository,
...
+            addNamesToHash(nameHash, allMoleculeNamesFromInitialConditionsInModule(moleculeBuilder, module));
+            addNamesToHash(nameHash, allReactionNamesFromModule(module));
+         }
+         else
+         {
+            addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInProject());
+            addNamesToHash(nameHash, allParametersFromMoleculesInProject());
+            addNamesToHash(nameHash, allMoleculeNamesFromInitialConditions(moleculeBuilder));
+            addNamesToHash(nameHash, allReactionNamesFromProject());
+         }
```
rwmcintosh: I think we still want to prevent names across the whole project in general, but, the specific case of a like-named molecule is special because during model construction the two molecules will be merged meaning we expect it. That's not the case when you have a molecule that has the same name as a spatial structure for example.

So, to generalize this, I think the intent should be to keep the project wide prohibition on _like-named_ entities, but for _like-typed_ entities, we want to allow _like-names_ because they will be merged if they collide in the model construction.

rwmcintosh (changes requested):

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -61,10 +61,23 @@ public ForbiddenNamesRetriever(IBuildingBlockRepository buildingBlockRepository,
...
-         addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInProject());
-         addNamesToHash(nameHash, allParametersFromMoleculesInProject());
-         addNamesToHash(nameHash, allMoleculeNamesFromInitialConditions(moleculeBuilder));
-         addNamesToHash(nameHash, allReactionNamesFromProject());
+         var module = moleculeBuilder.BuildingBlock?.Module;
+
+         if (module != null)
+         {
+            addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInModule(module));
+            addNamesToHash(nameHash, allParametersFromMoleculesInModule(module));
```
Yuri05: ~AFAICS we trying to prevent here that molecule is named as any parameter of the spatial structure or molecule-parameter? What was the rationale behind this?~ Clarified above.

On the other side, we don't prevent that a molecule is named as a top-container of the spatial structure (or e.g. `Events`). 
So I can name my molecule e.g. Neighborhoods and create a simulation with this. The molecule is then merged with the corresponding container in the simulation (which produce the weird views below).
Funny enough - such a simulation can be even run (at least in the simple case without reactions, transporters etc. - did not check what would happen in these cases).

So we should at least prohibit that a molecule is named as a top container of the spatial structure or `Events`.

Image

Image

Image

rwmcintosh: I think we want to leave all the name prohibition as-is except -> we want to allow like-type entities to have a like-name because they would be merged in model construction.

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -61,10 +61,23 @@ public ForbiddenNamesRetriever(IBuildingBlockRepository buildingBlockRepository,
...
+            addNamesToHash(nameHash, allMoleculeNamesFromInitialConditionsInModule(moleculeBuilder, module));
```
Yuri05: We should also add the names of existing **molecule builders** in the module.
E.g. if we added a molecule before and did not extend the ICs yet - the name of that molecule will be missed in the forbidden list.

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
Yuri05: I still don't understand the use case: when do we have the situation that we add/rename a molecule and have `module = null` ?

rwmcintosh: No, I'm not replying to that comment. I'm saying in general we don't want to have this naming limitation done in this way. We want to search the project, not the module, with the exception of the specific case of like-typed entity. In this exact case, we can use the old code and just not prevent molecule names.

When we rename a molecule, existing molecule names are not prohibited.

benjaperez1983: Good catch... Let me re-work this.

rwmcintosh: ok, maybe one further clarification using molecules as the example

We want to make sure that we don't allow rename a molecule to the same name as another molecule in the same building block (which means it will be in the same module)

But we do want to allow molecules to have the same name to molecules in other building blocks (and modules).

For all other name checks we want to keep them checking the whole project because when editing a name in a building block you don't know how it will be combined with other modules.

We still want to prevent molecules from being named the same as any parameter in the project for example because a user could combine modules in such a way that the molecule and the parameter end up in the same container. The model construction would fail in that case.

Yuri05: > We want to make sure that we don't allow rename a molecule to the same name as another molecule in the same building block (which means it will be in the same module)

Agree.

> But we do want to allow molecules to have the same name to molecules in other building blocks (and modules).

Agree.

>For all other name checks we want to keep them checking the whole project because when editing a name in a building block you don't know how it will be combined with other modules.

This was my initial question (and still unclear for me): how can it happen that during editing of a molecule, it does not belong to any module?

>We still want to prevent molecules from being named the same as any parameter in the project for example because a user could combine modules in such a way that the molecule and the parameter end up in the same container. The model construction would fail in that case.

Agree - but why only parameters? Container names should be checked in the same manner then, no?

rwmcintosh: I think it's not possible for the module to not be present. I think it must be.

For wider consideration on how we handle name prohibition -> yeah, let's consider this. I think for the purposes of this PR we want to allow the like-type to be named the same. Maybe it's only an issue for Molecules.

Yuri05: > I think it's not possible for the module to not be present. I think it must be.

if so, the part 
```
         else
         {
            addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInProject());
            addNamesToHash(nameHash, allParametersFromMoleculesInProject());
            addNamesToHash(nameHash, allMoleculeNamesFromInitialConditions(moleculeBuilder));
            addNamesToHash(nameHash, allReactionNamesFromProject());
         }
```
can be safely removed

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
Yuri05: rename the argument from `moleculeBuilder `to `reactionBuilder`

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
rwmcintosh: yeah, but if we might not have this at all because I think there's a bunch of stuff to redo. But if we have this section again, there's no need to check for module exists.

My request for changes is unrelated to yours.

Yuri05: > yeah, but if we might not have this at all because I think there's a bunch of stuff to redo. But if we have this section again, there's no need to check for module exists.

Fine with this. My bigger concerns are
* https://github.com/Open-Systems-Pharmacology/MoBi/pull/2289#pullrequestreview-3953636867
* Checking for parameters, but not for containers (which then can result in the weird merges in the simulation https://github.com/Open-Systems-Pharmacology/MoBi/pull/2289#pullrequestreview-3953616917) (applies also for Reactions - and then containers should also check for molecule/reaction names)

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -63,7 +63,7 @@ public IEnumerable<string> For(MoleculeBuilder moleculeBuilder)
          var nameHash = new HashSet<string>();
          addNamesToHash(nameHash, allParameterNamesFromSpatialStructureInProject());
          addNamesToHash(nameHash, allParametersFromMoleculesInProject());
-         addNamesToHash(nameHash, allMoleculeNamesFromInitialConditions(moleculeBuilder));
+         addNamesToHash(nameHash, allMoleculeNamesFromInitialConditionsInModule(moleculeBuilder));
```
Yuri05: I still think that the molecule names should be fetched from the **molecule builders** of the module and not from its initial conditions. Or at least ICs AND molecule builders
* We could have a molecule in the module which is not yet included into ICs - its name will be missed in the forbidden list
* We could have molecule names in the IC, which do not belong to any molecule in the module anymore (e.g. if I delete or rename a molecule without changing the ICs). Should this molecule name than appear in the forbidden names list? I think no. @PavelBal 

Btw. (not related to this PR): it would be good if the user would get a prompt for deleting all corresponding IC entries when deleting a molecule (similar to the rename prompt when renaming)

Yuri05: I created a separate issue for container names check:
https://github.com/Open-Systems-Pharmacology/MoBi/issues/2297
I think this should be done (in the future PRs) as well to avoid weird situations with merging of molecules into containers (even if the model construction does not crash and simulation runs)

PavelBal: > We could have molecule names in the IC, which do not belong to any molecule in the module anymore (e.g. if I delete or rename a molecule without changing the ICs). Should this molecule name than appear in the forbidden names list? I think no. @PavelBal

No... Molecules are only created when they are also present in the "Molecules" BB, not in the IC BB.

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
rwmcintosh: I would open new issues for specific items here. We have a lot of cases to consider.

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
rwmcintosh: They have to be present in both IC and Molecules right?

So if we have in the same module:

- IC with the same name, but not a molecule -> allow the rename
- No IC with the same name, but there's a molecule -> do not allow the rename
- Both -> do not allow
- Neither -> allow

So then it boils down to having nothing to do with the IC's anymore. Or what?

Yuri05: >So then it boils down to having nothing to do with the IC's anymore.

IMO: yes. Only with the molecules.

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
rwmcintosh: done

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
rwmcintosh: I had to draw it out in bullets before I got it. finally

src/MoBi.Core/Services/ForbiddenNamesRetriever.cs
```diff
@@ -63,12 +62,18 @@ public IEnumerable<string> For(MoleculeBuilder moleculeBuilder)
...
          addNamesToHash(nameHash, allReactionNamesFromProject());
          return nameHash;
       }
 
-      public IEnumerable<string> For(ReactionBuilder moleculeBuilder)
+      private static IReadOnlyList<string> allMoleculeNamesFromBuildingBlock(MoleculeBuilder builder)
+      {
+         var buildingBlock = builder.BuildingBlock as MoleculeBuildingBlock;
+         return buildingBlock == null ? Enumerable.Empty<string>().ToList() : buildingBlock.AllNames();
+      }
```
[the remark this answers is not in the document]
rwmcintosh: it's forbidden to rename a molecule to itself so this is ok.

Yuri05 (approved):

rwmcintosh (approved):

PR #2384, opened 2026-05-08
Fixes #2383 Align with new IParameterFactory.CreateDistributedParameter contract

### LINKED ISSUE #2383: Update PathAndValueEntityToDistributedParameterMapper to align with new IParameterFactory.CreateDistributedParameter contract
## Context

In OSPSuite.Core ([issue #2838](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2838)), `IParameterFactory.CreateDistributedParameter` is being updated so that the returned `IDistributedParameter` is fully formed: in addition to the distribution formula, it now pre-creates the sub-parameters referenced by the formula (e.g. `Mean`, `Deviation`/`GeometricDeviation`/`Minimum`/`Maximum`) plus `Percentile`. This makes the factory API honest — a "distributed parameter" returned by it is actually operational instead of half-built.

This change is required to fix [PK-Sim #3511](https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3511): a distributed parameter that is overwritten in an individual but not present in the spatial structure must surface in the simulation as distributed (with the overwritten value applied as a fixed value), not as a plain constant.

## MoBi impact

[`PathAndValueEntityToDistributedParameterMapper`](https://github.com/Open-Systems-Pharmacology/MoBi/blob/develop/src/MoBi.Core/Mappers/PathAndValueEntityToDistributedParameterMapper.cs) calls the same factory and then iterates caller-supplied sub-parameters, calling `distributedParameter.Add(...)` for each. After the Core change, the factory will already have created sub-parameters with the same names, and `IContainer.Add` throws `NotUniqueNameException` on duplicates.

## Required change

Update `PathAndValueEntityToDistributedParameterMapper.MapFrom` so that when adding a caller-supplied sub-parameter, any existing same-named child (placeholder created by the factory) is removed first, or the existing child is updated in place. Either approach should leave the mapper producing the same final structure as today (one child per caller-supplied sub-parameter, with the caller's name/value/dimension/formula/displayUnit).

## Done when

- `PathAndValueEntityToDistributedParameterMapper` no longer throws against the new factory behavior.
- Existing MoBi tests still pass.
- A test confirms that calling the mapper with sub-parameter names that overlap the factory's pre-created sub-parameters works (no duplicate children, caller's values win).

msevestre:
## Summary

- OSPSuite.Core's `IParameterFactory.CreateDistributedParameter` is being updated to return a fully-formed distributed parameter — it now pre-creates the sub-parameters referenced by the distribution formula (Mean, Deviation, Percentile, ...) and applies the optional value. See Open-Systems-Pharmacology/OSPSuite.Core#2857.
- This PR adapts MoBi's three callers of the mapper / factory to that new contract:
  - `PathAndValueEntityToDistributedParameterMapper` updates an existing factory-created sub-parameter's value when the caller provides one with a matching name, instead of `Add`-ing a duplicate (which would throw `NotUniqueNameException`).
  - `IndividualParameterToParameterDTOMapper` and `SpatialStructureContentExporter` no longer assign a `ConstantFormula(value)` after the mapper — the OSPSuite.Core mapper now applies the value internally. `IFormulaFactory` is no longer needed in either class.

## Test plan

- [x] New unit test `When_creating_a_distributed_parameter_whose_factory_already_pre_created_overlapping_sub_parameters` covers the duplicate sub-parameter-name case.
- [x] All MoBi unit/integration tests pass: 2215 passed (the obsolete `the_formula_factory_creates_a_formula_from_the_value` test on `IndividualParameterToParameterDTOMapperSpecs` was removed; the corresponding behavior is now covered in OSPSuite.Core).

## Release Notes

* **Bug Fixes**
  * Fixed issue where duplicate sub-parameter names could be created in distributed parameters; now correctly updates existing sub-parameters instead.
  * Improved parameter value handling and formula management in mapping operations.

* **Refactor**
  * Simplified internal dependency structure in mappers and export operations for better code organization.

### DIFF
--- a/src/MoBi.Core/Mappers/PathAndValueEntityToDistributedParameterMapper.cs
+++ b/src/MoBi.Core/Mappers/PathAndValueEntityToDistributedParameterMapper.cs
@@ -1,7 +1,7 @@
-﻿using OSPSuite.Core.Domain;
+﻿using System.Collections.Generic;
+using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Formulas;
-using System.Collections.Generic;
 using OSPSuite.Utility.Extensions;
 
 namespace MoBi.Core.Mappers
@@ -32,6 +32,16 @@ namespace MoBi.Core.Mappers
       {
          subParameters.Each(subParameter =>
          {
+            //the factory pre-creates the sub-parameters referenced by the distribution formula and a Percentile.
+            //If the caller provides a sub-parameter with the same name, just update the existing one's value.
+            var existing = distributedParameter.GetSingleChildByName<IParameter>(subParameter.Name);
+            if (existing != null)
+            {
+               if (subParameter.Value.HasValue)
+                  existing.Value = subParameter.Value.Value;
+               return;
+            }
+
             distributedParameter.Add(_parameterFactory.CreateParameter(subParameter.Name, subParameter.Value, subParameter.Dimension, formula: subParameter.Formula, displayUnit: subParameter.DisplayUnit));
          });
       }
--- a/src/MoBi.Presentation/Mappers/IndividualParameterToParameterDTOMapper.cs
+++ b/src/MoBi.Presentation/Mappers/IndividualParameterToParameterDTOMapper.cs
@@ -18,14 +18,12 @@ namespace MoBi.Presentation.Mappers
    {
       private readonly IParameterToParameterDTOMapper _parameterToParameterDTOMapper;
       private readonly IParameterValueToParameterMapper _parameterValueToParameterMapper;
-      private readonly IFormulaFactory _formulaFactory;
       private readonly ICloneManagerForBuildingBlock _cloneManager;
 
-      public IndividualParameterToParameterDTOMapper(IParameterToParameterDTOMapper parameterToParameterDTOMapper, IParameterValueToParameterMapper parameterValueToParameterMapper, IFormulaFactory formulaFactory, ICloneManagerForBuildingBlock cloneManager)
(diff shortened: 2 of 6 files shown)


src/MoBi.Core/Mappers/PathAndValueEntityToDistributedParameterMapper.cs
```diff
@@ -32,6 +32,16 @@ private void addSubParameters<TBuilder>(IReadOnlyList<TBuilder> subParameters, I
       {
          subParameters.Each(subParameter =>
          {
+            //the factory pre-creates the sub-parameters referenced by the distribution formula and a Percentile.
+            //If the caller provides a sub-parameter with the same name, just update the existing one's value.
+            var existing = distributedParameter.GetSingleChildByName<IParameter>(subParameter.Name);
+            if (existing != null)
```
msevestre, line 38: @rwmcintosh need help here. Hard to say for me if this is equivalent. I think it is. All tests are passing but I am not 100%sure as I do not know all the ins and outs of this code

msevestre (commented):
@rwmcintosh I had to adjust some code part that I do not really understand from a big picutre POV. The chnages in core are CREATING the parameters so we need to update MoBi with this new knowledge

src/MoBi.Core/Mappers/PathAndValueEntityToDistributedParameterMapper.cs
rwmcintosh: Yeah, seems right to me. You are just ensuring after parameter creation that the value from the PathAndValueEntity overrides what is in the distribution, for example `Mean` from the PathAndValue will win rather than the default created by the `parameterFactory`?

rwmcintosh (approved):

src/MoBi.Core/Mappers/PathAndValueEntityToDistributedParameterMapper.cs
msevestre: yep exactly

PR #2393, opened 2026-05-12
Fixes #2391 Show formula and paths for transport in simulation

### LINKED ISSUE #2391: Show formula and paths for transport in simulation
Can resolved formula and paths be shown in the simulation for transport similarly as for reactions?

Reactions

Image

rwmcintosh:
Fixes #2391

## Summary
Adds an `EditTransportInSimulationPresenter` / `EditTransportInSimulationView` pair so that clicking a `Transport` in the simulation tree shows its resolved rate formula, source/target container paths, moved molecule, dimension, and a read-only parameter values tab — mirroring the existing reaction display.

Shared logic between the reaction and transport in-simulation presenters is lifted into a generic abstract base `EditProcessInSimulationPresenter<TView, TPresenter, TProcess> where TProcess : Process`. The two views remain separate concrete classes (the layouts diverge — stoichiometric label vs. source/target/molecule/dimension fields). The factory `EditInSimulationPresenterFactory.PresenterFor` gets a new `Transport` clause between the `Reaction` and the generic `IContainer` fallback.

Drive-by: changed the read-only name field on the reaction view and the formula string on `EditExplicitFormulaView` from `Enabled = false` to `Properties.ReadOnly = true`, so the text remains selectable/copy-friendly instead of greyed out.

## Test plan
- [x] `dotnet build` succeeds for `MoBi.Presentation`, `MoBi.UI`, and `MoBi.Tests`
- [x] Full `MoBi.Presentation` test namespace passes (1340 passed, 1 skipped, 0 failed)
- [x] New `EditTransportInSimulationPresenterSpecs` mirrors the reaction specs and passes
- [x] `ShowPresenterFactorySpecs` extended with a Transport routing assertion
- [x] Manual smoke: load a simulation with passive transports, click a transport, confirm rate, source path, target path, molecule, dimension, description and read-only parameter values all show
- [x] Manual smoke: regression check that reactions still render identically

image

## Summary by CodeRabbit

* **New Features**
  * Added support for editing Transport processes during simulations with parameter configuration and kinetic formula viewing.

* **Improvements**
  * Modified read-only field behavior in formula editors and transport views to use read-only mode instead of disabling controls.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Open-Systems-Pharmacology/MoBi/pull/2393)

### DIFF
--- a/src/MoBi.Presentation/Presenter/EditProcessInSimulationPresenter.cs
+++ b/src/MoBi.Presentation/Presenter/EditProcessInSimulationPresenter.cs
@@ -1,0 +1,82 @@
+using System;
+using System.Collections.Generic;
+using MoBi.Presentation.Presenter.BasePresenter;
+using MoBi.Presentation.Views;
+using OSPSuite.Core.Domain;
+using OSPSuite.Presentation.Presenters;
+using OSPSuite.Utility.Events;
+
+namespace MoBi.Presentation.Presenter
+{
+   public abstract class EditProcessInSimulationPresenter<TView, TPresenter, TProcess>
+      : AbstractEditPresenter<TView, TPresenter, TProcess>, IEditInSimulationPresenter
+      where TView : class, IEditProcessInSimulationView<TPresenter>
+      where TPresenter : IEditPresenter
+      where TProcess : Process
+   {
+      protected TProcess _process;
+      private readonly IEditParametersInContainerPresenter _editParametersInContainerPresenter;
+      private readonly IFormulaPresenterCache _formulaPresenterCache;
+      private IEditTypedFormulaPresenter _formulaPresenter;
+
+      public TrackableSimulation TrackableSimulation
+      {
+         get;
+         set
+         {
+            field = value;
+            _editParametersInContainerPresenter.EnableSimulationTracking(value);
+         }
+      }
+
+      protected EditProcessInSimulationPresenter(TView view,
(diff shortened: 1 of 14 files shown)


src/MoBi.Presentation/Presenter/EditProcessInSimulationPresenter.cs
rwmcintosh, line 1: pulled up from EditReactionInSimulationPresenter

src/MoBi.UI/Views/EditTransportInSimulationView.cs
```diff
@@ -0,0 +1,73 @@
...
+         tabParameters.Show();
+      }
+
+      public override void InitializeResources()
+      {
+         base.InitializeResources();
+         layoutControlDescription.Text = AppConstants.Captions.Description.FormatForLabel();
+         layoutControlKinetic.Text = AppConstants.Captions.Kinetic.FormatForLabel();
+         layoutControlKinetic.TextLocation = Locations.Top;
+         layoutControlKinetic.MinSize = new Size(3 * OSPSuite.UI.UIConstants.Size.BUTTON_HEIGHT, 2 * OSPSuite.UI.UIConstants.Size.BUTTON_WIDTH);
```
msevestre, line 59: that's very funky here. Why these weird sizing?

msevestre (approved):

PR #2434, opened 2026-07-13
Fixes #2429 Show all reaction parameters in 'Possible referenced objects'

### LINKED ISSUE #2429: "Possible referenced objects" for reactions only shows the last created parameter of a reaction
1) Open a reactions BB
2) Create a new reaction
3) In the reaction, create >1 parameters

Image

4) In the "Properties" tab of this reaction, list "Possible referenced objects" only shows the parameter that was created last, not all parameters of the reaction

Image

rwmcintosh:
Fixes #2429
Adding a node to the tree when the parent was not already expanded meant that the lazy-loading would not run when the tree is eventually expanded. What we want to do is add the node if the tree has already been expanded to that level once and all the children are present.

We can't use `HasChildren` because it could be a node that was fully expanded, but has no children.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tree updates when adding reaction parameters.
  * Ensured newly added parameters appear immediately when the reaction’s children have already been loaded, including the first parameter added to an empty reaction.

* **Tests**
  * Added coverage for parameter additions before and after reaction expansion.

### DIFF
--- a/src/MoBi.Presentation/Nodes/HierarchicalStructureNode.cs
+++ b/src/MoBi.Presentation/Nodes/HierarchicalStructureNode.cs
@@ -17,6 +17,13 @@ namespace MoBi.Presentation.Nodes
          _childrenLoaded = false;
       }
 
+      /// <summary>
+      ///    Returns <c>true</c> once the children have been enumerated (lazily loaded) into this node, even if that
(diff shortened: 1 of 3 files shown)


rwmcintosh:
@coderabbitai full review please

Yuri05 (approved):

PR #2435, opened 2026-07-14
Fixes #2428 Cannot select Global parameter type for a molecule parameter

### LINKED ISSUE #2428: Cannot select "Global" type for new parameters in a Molecule
1. Open a simulation (e.g., the Aciclovir example simulation)
2. Open the Molecules BB
3. Select a molecule (e.g., Aciclovir)
4. Select "New parameter" and change the "Parameter Type" from "Local" to "Global"

Image

5) "Parameter type" is now empty:

Image

When clicking "OK", the parameter is properly created as "Global", so its probably only a UI problem.

rwmcintosh:
## Problem

When creating a new parameter in a Molecule (or Reaction) building block and changing **Parameter type** from `Local` to `Global`, the combo box goes blank instead of showing `Global`. The parameter is still created correctly as `Global` on OK — the defect is purely in the UI display.

Fixes #2428

## Summary by CodeRabbit

* **Bug Fixes**
  * Editing a parameter now refreshes the existing view content rather than fully re-rendering it, avoiding unnecessary list rebinding.
  * Changes to other parameters no longer trigger an unintended refresh while editing is in progress.
* **Tests**
  * Updated presenter specs to validate the new refresh behavior and to ensure refresh is not called when the edited parameter is not the one that changed.

### DIFF
--- a/rakefile.rb
+++ b/rakefile.rb
@@ -111,7 +111,7 @@ def copy_templates_files(source_dir)
 end
 
 def relative_src_dir_for(configuration)
-   File.join('src', 'MoBi', 'bin', configuration, 'net8.0-windows')
+   File.join('src', 'MoBi', 'bin', configuration, 'net10.0-windows')
 end
 
 def src_dir_for(configuration)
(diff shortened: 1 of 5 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #2438, opened 2026-07-15
Fixes #2432 Create new MoleculeProperties parameters as Global

### LINKED ISSUE #2432: Creating a new parameter in `MoleculeProperties` - alwas set to "global" type
When creating a new `MoleculeProperties` parameter in the spatial structure, there is no field for changing the global/local option, and no column showing the property:

Image

However, if I enable the "Type" column, I see that the new parameter is created as "local", while all other parameters are "global":

Image

So the new parameter behaves differently from the other "MoleculeProperties".

In the simulation itself, the parameter is created in `<MOLECULE>|parameter`, so it behaves like a GLOBAL parameter (event though it is marked as "local"). But when I try to reference it in a formula, it fails because it is being resolved as a LOCAL parameter.

New `MoleculeProperties` parameters should be created as **Global**, consistent with all other `MoleculeProperties` parameters.

---

_The second problem originally reported here (`MoleculeProperties` parameters only being created for non-stationary molecules) has been split out into #2437._

rwmcintosh:
New parameters added to a `MoleculeProperties` container in the spatial structure were created with `Local` build mode, while all other `MoleculeProperties` parameters are `Global`. As reported in #2432, this makes the parameter land at the global `<MOLECULE>|parameter` path but be resolved as local when referenced in a formula, causing the reference to fail.

`DefaultParameterBuildMode` now returns `Global` for `MoleculeProperties` containers, covering both parameter-creation flows. Added a spec asserting the build mode is `Global`.

Fixes #2432

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected project and test references so the MoBi.R project loads and builds reliably.
  * Global molecule properties now use the appropriate global parameter build mode, while local properties retain local behavior.
  * Resolved a project task setup issue affecting simulation and parameter-identification workflows.

* **Tests**
  * Added coverage for parameter creation in global and local molecule properties containers.


src/MoBi.Core/Domain/Extensions/ContainerExtensions.cs
```diff
@@ -25,7 +25,8 @@ public static ParameterBuildMode DefaultParameterBuildMode(this IContainer conta
       {
          var needsGlobalParameter = container.IsAnImplementationOf<TransporterMoleculeContainer>() ||
                                     container.IsAnImplementationOf<ReactionBuilder>() ||
-                                    container.IsAnImplementationOf<InteractionContainer>();
+                                    container.IsAnImplementationOf<InteractionContainer>() ||
+                                    container.IsMoleculeProperties();
```
Yuri05: what about new parameters in the **local** `MoleculeProperties `containers? They must be of type `Local`.

Image

rwmcintosh: Would this be enough of an indicator
```
return container.IsMoleculeProperties() && container.ContainerType == ContainerType.Molecule;
```

Seems this container from `SpatialStructure.GlobalMoleculeDependentProperties` is always created with this type.

Yuri05 (approved):

PR #2444, opened 2026-07-20
Fixes #2443 Restore simulation association on PI snapshot import

### LINKED ISSUE #2443: Loading snapshot with a parameter identification throws NullReferenceException
Loading a project snapshot that contains a parameter identification fails with `Object reference not set to an instance of an object`.

**Cause:** In `MoBi.Core.Snapshots.Mappers.ProjectMapper.mapToModel`, parameter identifications are mapped *before* simulations are added to the project. The PI mapper resolves each linked parameter's simulation via `MoBiProject().Simulations.FindByName(...)`, which returns `null` (hence the `Cannot find simulation` warnings). Those `null` `ParameterSelection` entries are then passed to `IdentificationParameterFactory.CreateFor`, where `.Where(x => x.Parameter != null)` dereferences `null` and throws.

PK-Sim's `ProjectMapper` loads simulations *before* parameter identifications; MoBi has the two blocks reversed.

**Fix:** Move the PI mapping to after the simulation-loading loop in MoBi's `ProjectMapper`, matching PK-Sim. Optionally, guard `IdentificationParameterFactory.CreateFor` (OSPSuite.Core) against `null` selections so a genuinely-missing simulation degrades to a warning instead of crashing.

rwmcintosh:
Fixes #2443

Loading a project snapshot containing a parameter identification dropped the PI's simulation association (and threw a `NullReferenceException`).

Two causes, both fixed here:
- **`ProjectMapper`** mapped parameter identifications *before* simulations were added to the project, so name lookups failed. PIs are now mapped after simulations load, matching PK-Sim.
- **`MoBiProject.All<T>()`** only enumerated building blocks, but simulations live in a separate list. The Core PI mapper resolves the PI's `Simulations` via `project.All<ISimulation>()`, which was always empty in MoBi. `All<T>()` now includes simulations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved project item retrieval so simulations are included consistently alongside other project components.
  * Adjusted project loading so parameter identifications are restored after simulation setup, improving snapshot loading consistency.

### DIFF
--- a/src/MoBi.Core/Domain/Model/MoBiProject.cs
+++ b/src/MoBi.Core/Domain/Model/MoBiProject.cs
@@ -39,6 +39,9 @@ public class MoBiProject : Project
       parameterIdentification.IsLoaded = true;
    }
 
+   /// <summary>
+   /// Returns all simulations or building blocks matching <typeparamref name="T"/>
+   /// </summary>
    public override IReadOnlyCollection<T> All<T>() => get<T>();
 
    public IEnumerable<CurveChart> Charts => _charts;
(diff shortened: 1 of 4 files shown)


rwmcintosh:
@coderabbitai when you cannot review a PR due to rate limit that should not mean the PR has a failing check. We have other reviewers.

Yuri05 (approved):

PR #2452, opened 2026-07-24
Fixes #1908 Defer simulation model construction until the model is accessed

### LINKED ISSUE #1908: Load projects without loading simulations
In R, we don't want to automatically load all simulations in a project. We may not want any of them, but only the modules and building blocks

rwmcintosh:
## What it does
Simulations load as lightweight **shells**: everything except the `IModel` is deserialized when the project opens, and the model is constructed **on first access to `Model`**, from the compressed content kept in a new project‑scoped `ISimulationContentRepository`. This takes model construction — the dominant cost of opening a project — off the load path when simulations aren't touched.

## Performance
`my project` (4 simulations), opened via MoBi.R: eager **~15.3 s → ~4.9 s** deferred (~3×).

## Tests
All green locally: MoBi.Tests unit **2460**, non‑SBML integration **94** (incl. new lazy‑load + save round‑trip specs), MoBi.R.Tests **227**.

## Summary by CodeRabbit

* **New Features**
  * Simulations can now load their models on demand, improving project loading performance.
  * Accessing a simulation model automatically loads and registers it when needed.
  * Project saves preserve deferred simulation models without unnecessarily loading them.

* **Bug Fixes**
  * Unregistered simulations are now safely ignored during removal.
  * Clearing the application context also clears retained simulation content.

* **Tests**
  * Added coverage for deferred loading, model access, project saving, and simulation content handling.


src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -83,12 +90,58 @@ public class MoBiSimulation : ModelCoreSimulation, IMoBiSimulation
...
+      get
+      {
+         // A shell was loaded without a model. Materialize it on first access.
+         if (!SuppressModelMaterialization && !_modelMaterialized && _lazyModelLoader != null)
+         {
+            // Guard against re-entrancy: registering the materialized model re-reads Model via AcceptVisitor.
+            _modelMaterialized = true;
+            var loader = _lazyModelLoader;
+            _lazyModelLoader = null;
+            loader();
```
rwmcintosh, line 133: hydrated on first access

src/MoBi.Core/Serialization/ORM/Mappers/ProjectToProjectMetaDataMapper.cs
```diff
@@ -62,7 +65,16 @@ private SimulationMetaData mapFrom(IMoBiSimulation simulation)
...
             return simulationMetaData;
 
-         serialize(simulationMetaData, simulation);
+         // The simulation has changes but its model was never materialized: the model is therefore unchanged, so
+         // reuse its retained bytes rather than building the model just to re-serialize it.
+         if (!simulation.IsModelLoaded && _simulationContentRepository.Contains(simulation.Id))
+         {
+            simulationMetaData.Discriminator = simulation.GetType().Name;
+            simulationMetaData.Content.Data = _serializationService.SerializeSimulationReusingModel(simulation, _simulationContentRepository.ContentFor(simulation.Id));
+         }
```
rwmcintosh, line 74: saving a simulation that's been changed, but not the model

src/MoBi.Core/Serialization/ORM/Mappers/ProjectMetaDataToProjectMapper.cs
```diff
@@ -136,7 +144,17 @@ private void addObservedDataToSerializationContext(SerializationContext serializ
...
+         var content = simulationMetaData.Content;
+         var simulation = _serializationService.DeserializeSimulationDeferringModel(content.Data, _project, serializationContext);
+
+         // A shell was returned (its model was deferred). Keep the compressed content so the model can be
+         // constructed on first access, and wire the lazy trigger.
+         if (!simulation.IsModelLoaded)
+         {
+            _simulationContentRepository.Register(simulation.Id, content.Data);
+            _simulationModelLoader.ConfigureLazyModelLoad(simulation);
+         }
```
rwmcintosh, line 156: A simulation gets a lazy load action that runs when Model get is accessed. That happens at deserialization time, but only if the simulation is the current version. If it's not, the current version, then it's deserialized eagerly to apply project conversions.

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
Yuri05: Why do you need to replace the initial __lazyModelLoader_ with _null_?
Isn't setting `_modelMaterialized = true` enough?

src/MoBi.Core/Serialization/ORM/Mappers/ProjectToProjectMetaDataMapper.cs
```diff
@@ -62,7 +65,16 @@ private SimulationMetaData mapFrom(IMoBiSimulation simulation)
          if (!simulation.HasChanged)
             return simulationMetaData;
 
-         serialize(simulationMetaData, simulation);
+         // The simulation has changes but its model was never materialized: the model is therefore unchanged, so
+         // reuse its retained bytes rather than building the model just to re-serialize it.
```
Yuri05, line 69: What kind of changes does this include? Just something like simulation description?

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
rwmcintosh: Yeah, it would be enough. But also that just frees up the reference. It should never be needed again after this.

src/MoBi.Core/Serialization/ORM/Mappers/ProjectToProjectMetaDataMapper.cs
rwmcintosh: Yeah, a property that doesn't touch the model. Name actually does, but you get the idea... simple properties that are not tied to the model

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
Yuri05: OKI

rwmcintosh:
The OSPSuite.Core prerequisite is now open: Open-Systems-Pharmacology/OSPSuite.Core#2922 — makes `ModelCoreSimulation.Model` `virtual` (Core issue #2921).

Once that merges and a Core version is released, the `OSPSuite.*` references here get repointed off the local build to that release and this comes out of draft. Until then CI can't restore the pinned local Core package.

Yuri05:
@coderabbitai full review

Yuri05:
@rwmcintosh Pls check the 🐰 comments :)

src/MoBi.Core/Domain/Repository/SimulationContentRepository.cs
```diff
@@ -0,0 +1,46 @@
...
+
+   public class SimulationContentRepository : ISimulationContentRepository
+   {
+      private readonly Cache<string, byte[]> _contentById = new Cache<string, byte[]>(onMissingKey: _ => null);
+
+      public void Register(string simulationId, byte[] content) => _contentById[simulationId] = content;
+
+      public bool Contains(string simulationId) => _contentById.Contains(simulationId);
+
+      public byte[] ContentFor(string simulationId) => _contentById[simulationId];
```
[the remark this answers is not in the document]
rwmcintosh: We don't need to be so defensive. The bytes are not going to be manipulated

src/MoBi.Core/Domain/Model/MoBiSimulation.cs
```diff
@@ -83,12 +90,58 @@ public class MoBiSimulation : ModelCoreSimulation, IMoBiSimulation
...
+         }
+
+         return base.Model;
+      }
```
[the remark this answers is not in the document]
rwmcintosh: implemented

Yuri05 (approved):

PR #2453, opened 2026-07-24
Fixes #2436 Prepend the Events container to event-group parameter value paths

### LINKED ISSUE #2436: Path element “Events” is missing in the path after adding a new parameter
Example:
In the attached MoBi file, there is a Module “IV2” with an events building  block (event group loaded via pkml from the original PK-Sim application in Module "Sim1") and a parameter building block. 
The parameter set “PV2000_afterAddingParameter” contains the parameter “Dose” after adding it by right click “New parameter value” and selecting the dose parameter from the ProtocolSchemaItem within IV2-Events. Looks like this (after adjusting the parameter value and selecting the correct dimension):  

Image

However, this parameter does not work since it misses the path element “Events” as path element 0 and each of the original path elements are in fact on lever deeper. I.e., the path should look like this (as in parameters set “PV2000_correctPath” within Module IV2 of the attached MoBi file): 

Image

Seems that for events, right click “New parameter value” is not producing the correct path.

MoBi file: 

[MissingPathElementExample.zip](https://github.com/user-attachments/files/30052341/MissingPathElementExample.zip)

rwmcintosh:
## Problem

Adding a parameter value via right-click **New parameter value** and selecting a parameter from an **events** building block (e.g. the `Dose` parameter of a `ProtocolSchemaItem`) produced a path that is missing the leading `Events` element, so the parameter value did not resolve against the model.

Reproduced from the [issue's example project](https://github.com/Open-Systems-Pharmacology/MoBi/issues/2436):

Fixes #2436

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected reference paths for parameters selected within event groups.
  * Event-group parameter paths now include the appropriate Events container, ensuring selections resolve correctly.

### DIFF
--- a/src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
+++ b/src/MoBi.Presentation/Presenter/SelectReferenceAtParameterValuePresenter.cs
@@ -130,7 +130,22 @@ namespace MoBi.Presentation.Presenter
             return returnPath;
          }
 
-         return CreatePathFor(item);
+         var itemPath = CreatePathFor(item);
+         prependEventsContainerIfMissing(item, itemPath);
+         return itemPath;
+      }
+
+      private static void prependEventsContainerIfMissing(IEntity item, ObjectPath objectPath)
+      {
+         // Entities selected from an event group building block are rooted at their top
+         // EventGroupBuilder, but in the built model these live under the "Events" top container.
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #2454, opened 2026-07-24
Coverage: replace dotcover.xml with coverage.runsettings (net10)

rwmcintosh:
## Problem
The scheduled coverage job fails on net10. Root causes:
1. dotCover 2026.2 CLI changed (kebab-case flags, `--flag value` not `--flag=value`, XML config removed) — silently ignored the old args and hung.
2. More fundamentally, dotCover 2026.2.0 (the only version that can instrument .NET 10) **cannot generate a report** — the console runner hangs/crashes in report generation ([DCVR-13230](https://youtrack.jetbrains.com/projects/DCVR/issues/DCVR-13230), open). Older versions cannot instrument net10 at all.

## Fix
Switch coverage to `dotnet test` with the **Microsoft Code Coverage** collector (supports net10, emits cobertura for codecov), driven by the shared action ([Workflows PR](https://github.com/Open-Systems-Pharmacology/Workflows/pull/247)). This repo:
- replaces `dotcover.xml` with `coverage.runsettings` (module/source exclusions preserved),
- drops the obsolete `xml-configuration` input from the coverage workflow.

Test projects already reference Microsoft.NET.Test.Sdk + NUnit3TestAdapter, so no csproj changes are needed. Verified locally end-to-end on a net10 NUnit project.

---
**Update — UI exclusion added.** A second commit excludes the UI layer from coverage (Views/WinForms cannot be exercised by the test infrastructure), replacing the per-repo UI *Views source* exclude with a whole *UI module* exclude. This is on top of the test-assembly exclusion, so coverage reflects testable product code. Verified on net10 that the `*.UI.dll` ModulePath exclude drops the UI assembly from the report.


coverage.runsettings
```diff
@@ -0,0 +1,29 @@
...
+  <DataCollectionRunSettings>
+    <DataCollectors>
+      <DataCollector friendlyName="Code Coverage" uri="datacollector://Microsoft/CodeCoverage/2.0">
+        <Configuration>
+          <Format>cobertura</Format>
+          <CodeCoverage>
+            <ModulePaths>
+              <Exclude>
+                <ModulePath>.*FakeItEasy.*</ModulePath>
+                <ModulePath>.*\.Tests\.dll$</ModulePath>
```
Yuri05, line 14: What is excluded by `.*\.Tests\.dll$` ?

rwmcintosh: The test code. It should not be included in stats about coverage. Always 100%

rwmcintosh: We used to count that coverage, but it's never too late.

Yuri05: Makes sense :)

Yuri05 (approved):

PR #2467, opened 2026-07-30
Fixes #2466 Restore the application molecule builder editor

### LINKED ISSUE #2466: Cannot create an Application Molecule Builder: no component for ICreatePresenter<ApplicationMoleculeBuilder>
Regression from #2412 (commit e20a4344f, "Remove dead code").

rwmcintosh:
This presenter/view cluster is only ever resolved through the ICreatePresenter<> registration convention, so it had no static references and #2412 removed it as dead code. Creating an application molecule builder has thrown ever since, from both the tree context menu and the application editor grid.

Fixes #2466


rwmcintosh:
Straight re-add of previously removed code

PR #2502, opened 2026-08-20
Fixes #2313 Cannot create simulation if PV contains a formula based on the parameter in the individual

### LINKED ISSUE #2313: Cannot create simulation if PV contains a formula based on the parameter in the individual
Happens both in 12.2 and 13. 

[bug_PV_Individual.zip](https://github.com/user-attachments/files/26377308/bug_PV_Individual.zip)
In the project attached, there is an entry in PV which is defined by the formula:

In this formula, `AeffVariabilityFactor` refers to the individual parameter `Organism|Lumen|Effective surface area variability factor` :

Image

When trying to create a new simulation with the only module and the only individual in this project - it fails because the parameter `Organism|Lumen|Effective surface area variability factor` could not be found.

Object Type | Object Name | Building Block Type | Building Block Name | Message | Origin
-- | -- | -- | -- | -- | --
Parameter Value | Effective surface area | Parameter Values Building Block | S1 - Parameter Values | Parameter 'Effective surface area' with path 'S2\|Organism\|Lumen\|Caecum\|Effective surface area' references an entity with path 'S1\|Organism\|Lumen\|Effective surface area variability factor' that cannot be found | Simulation

Another problem: in the PV, the formula list is empty, ~so not sure where the (only) formula in the PV refers to:~
EDIT: seems to be just a refresh issue; after closing/reopening the project the formula appears in the list

To recreate the problem from scratch:
* remove the entry from PV
* Update the simulation S1 from Building blocks
* In the simulation S1:
  * Overwrite the formula of `Organism|Lumen|Caecum|Effective surface area` with a const value
    
  * Reset the value of of `Organism|Lumen|Caecum|Effective surface area` to its original formula by clicking on the Reset button
    
* Commit the simulation S1 to building blocks

rwmcintosh:
Fixes #2313 

When a simulation parameter is committed to a Parameter Values / Initial Conditions building block, its formula references were cloned verbatim from the built model, keeping the simulation-root prefix (e.g. `S1|Organism|...`). On a later build `TopContainerPathReplacer` only re-roots paths starting with a top-container name, so the stale `S1` was never rewritten and the reference failed to resolve in the new simulation.

Fix: strip the model-root (simulation) name from the committed formula references (inverse of `TopContainerPathReplacer`) so the building block stores model-relative paths. Covers both ParameterValue and InitialCondition commits.

Note: projects already saved with an `S1`-rooted formula in a building block's FormulaCache are not auto-healed; delete the affected PV and its cached formula, then re-commit.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed formulas added from simulation quantities so absolute references correctly omit the simulation model root.
  - Preserved relative formula references when adding path/value entities.

- **Tests**
  - Added coverage for both absolute references requiring cleanup and relative references that should remain unchanged.

### DIFF
--- a/src/MoBi.Core/Commands/AddPathAndValueEntityFromQuantityInSimulationCommand.cs
+++ b/src/MoBi.Core/Commands/AddPathAndValueEntityFromQuantityInSimulationCommand.cs
@@ -3,6 +3,7 @@ using MoBi.Core.Domain.Model;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
+using OSPSuite.Utility.Extensions;
 using MoBi.Assets;
 using MoBi.Core.Domain.Services;
 using MoBi.Core.Helper;
@@ -41,6 +42,7 @@ namespace MoBi.Core.Commands
             return;
 
          var pathAndValueEntity = CreateNewEntity(context);
+         removeSimulationRootFromFormulaReferences(pathAndValueEntity);
          ObjectType = _objectTypeResolver.TypeFor(pathAndValueEntity);
          Description = AppConstants.Commands.AddedPathAndValueEntity(pathAndValueEntity, _buildingBlock.Name, ObjectType);
          _buildingBlock.Add(pathAndValueEntity);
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

Issue #2485, opened 2026-08-13
Malformed snapshot when importing PK-Sim simulations that share the same individual BB


PavelBal:
1. Create two simulations in PK-Sim that use the same individual
2. Import them in MoBi. When importing the second simulation, the individual in MoBi is automatically renamed:

<img width="254" height="114" alt="Image" src="https://github.com/user-attachments/assets/f866a05a-5103-4b72-85da-4d0fad38e908" />

But when exporting the project to snapshot, the two individuals have the same name:

```
  "IndividualBuildingBlockSnapshots": [
    {
      "PKSimSnapshot": {
        "Name": "Walker 1979 Volunteer 6",
        "Seed": 182030328,
        "OriginData": {
          "CalculationMethods": [
            "SurfaceAreaPlsInt_VAR1",
            "Body surface area - Mosteller"
          ],
          "Species": "Human",
          "Population": "European_ICRP_2002",
          "Gender": "MALE",
          "Age": {
            "Value": 27.0,
            "Unit": "year(s)"
          },
          "Weight": {
            "Value": 68.0,
            "Unit": "kg"
          },
          "ValueOrigin": {
            "Source": "Publication",
            "Method": "InVivo",
            "Description": "Walker 1979"
          }
        }
      },
      "UpdatedValues": []
    },
    {
      "PKSimSnapshot": {
        "Name": "Walker 1979 Volunteer 6",
        "Seed": 182030328,
        "OriginData": {
          "CalculationMethods": [
            "SurfaceAreaPlsInt_VAR1",
            "Body surface area - Mosteller"
          ],
          "Species": "Human",
          "Population": "European_ICRP_2002",
          "Gender": "MALE",
          "Age": {
            "Value": 27.0,
            "Unit": "year(s)"
          },
          "Weight": {
            "Value": 68.0,
            "Unit": "kg"
          },
          "ValueOrigin": {
            "Source": "Publication",
            "Method": "InVivo",
            "Description": "Walker 1979"
          }
        }
      },
      "UpdatedValues": []
    }
  ]
}
```

Trying to load this snapshot in MoBi throws an error:

Application:
MoBi® 13.0.36

Sequence contains more than one matching element

Stack trace:
```
at System.Linq.ThrowHelper.ThrowMoreThanOneMatchException()
   at System.Linq.Enumerable.TryGetSingle[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
[...]

PR #2503, opened 2026-08-20
Fix duplicate module name suffix when importing a colliding simulation

### LINKED ISSUE #1763: If I rename the imported simulation to e.g. **xxx**, the module is then renamed not to **xxx** but to **xxx 1**
Not a big deal, but if I rename the imported simulation to e.g. **xxx**, the module is then renamed not to **xxx** but to **xxx 1**
I would not change it now in V12 - but maybe in V12.1?
grafik

_Originally posted by @Yuri05 in https://github.com/Open-Systems-Pharmacology/MoBi/issues/1554#issuecomment-2590980075_

rwmcintosh:
Fixes https://github.com/Open-Systems-Pharmacology/MoBi/issues/1763

Importing a second PK-Sim simulation whose name and module both collide produced a doubled suffix (e.g. `…table 1 1` instead of `…table 1`): the module was renamed once for its own collision and again by the simulation-follow prefix `Replace`.

Fix consolidates module renaming into a single pass after the simulation name is finalized. Cosmetic only — names were always unique. Unrelated to #2485.

Added a regression spec (verified failing before the fix).


rwmcintosh:
@coderabbitai review please

rwmcintosh:
Marking as draft: #2500 will introduce a breaking test that overlaps with this area. Once #2500 merges, merge `develop` into this branch and reconcile before marking ready for review.

Yuri05 (approved):

rwmcintosh:
#2500 is merged. Merged `develop` into this branch (merge commit 8d5ef673) — no conflicts. Full MoBi.Tests suite passes (2644 passed, 0 failed), so the anticipated test conflict didn't materialize. Marking ready for review.

PR #2506, opened 2026-08-21
Fail a qualification run when errors are logged

### LINKED ISSUE #2505: Fail a qualification run when errors are logged
Follow-up to Open-Systems-Pharmacology/QualificationRunner#144.

During a qualification run, some conditions that degrade a simulation are only logged, so the CLI still exits `0` and the QualificationRunner reports success even though the results may be wrong. Logged severity currently has **no effect on control flow** — only a thrown exception changes the CLI exit code.

### Fail the run when an error is logged (enforcement gate)
Add an error-counting `ILoggerProvider` to `MoBi.CLI`, registered on the shared `ILoggerCreator` **only for the qualification command**. After `RunBatchAsync`, if any `Error`/`Critical` message was logged, set a non-zero exit code. QualificationRunner already treats a non-zero exit as an invalid project, so no change is needed there.

Files:
- `src/MoBi.CLI/ErrorCountingLoggerProvider.cs` (new)
- `src/MoBi.CLI/Program.cs`

This mirrors the same gate being added to PK-Sim. The result-changing snapshot-mapping reclassifications from QualificationRunner#144 live in `PKSim.Core` (the compound/simulation mappers), so MoBi only needs the general CLI gate as the safety net.

A branch with this change (`qualification-fail-on-logged-errors`) has been prepared and builds clean (`MoBi.CLI`: 0 warnings, 0 errors); a PR will follow.

rwmcintosh:
Fixes #2505

Logged errors during a qualification run now set a non-zero exit code so the QualificationRunner stops instead of reporting success with wrong results.

Adds an error-counting logger provider to `MoBi.CLI`, registered for the qualification command; any `Error`/`Critical` message logged during the run fails it.

Mirror of PK-Sim Log error return code.

### DIFF
--- a/src/MoBi.CLI/ErrorCountingLoggerProvider.cs
+++ b/src/MoBi.CLI/ErrorCountingLoggerProvider.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Threading;
+using Microsoft.Extensions.Logging;
+
+namespace MoBi.CLI
+{
+   public class ErrorCountingLoggerProvider : ILoggerProvider
+   {
+      private int _errorCount;
+
+      public bool HasErrors => _errorCount > 0;
+
+      public ILogger CreateLogger(string categoryName) => new ErrorCountingLogger(this);
+
+      private void increment() => Interlocked.Increment(ref _errorCount);
+
+      public void Dispose()
+      {
+      }
+
+      private class ErrorCountingLogger : ILogger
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

msevestre (approved):
I am ok with the code duplication with PKSim here.

Issue #2513, opened 2026-08-25
Rework the "Warnings" tab


PavelBal:
We have so many different types of warnings shown in this tab, that it would make sense to either

1) introduce a category for warnings so they can be filtered by it, or
2) Have different tabs for different types of warnings.

PR #2515, opened 2026-08-25
Fixes #2511 Add tooltip to "Warn for non-finite quantities" option

### LINKED ISSUE #2511: Add tooltip to "Warn for non-finite quantities"
Image

rwmcintosh:
## Summary

Fixes #2511. The **Warn for non-finite quantities** option on the *General* tab of the Options dialog gave no hint about what it checks or when it runs. It now has a tooltip:

> Show a warning for each quantity whose value is NaN or infinite at time t=0.
> The check is performed when a simulation is created and each time it is run.

The wording follows what the setting actually drives — `SimulationQuantityValueWarningTask.WarnForNonFiniteQuantities` in Core walks all `IQuantity` children of the model root and raises a warning for each NaN and each infinite value. It is invoked from two places: `ModelConstructor` during simulation creation, and `SimulationRunner` before each run.

### DIFF
--- a/src/MoBi.Assets/ToolTips.cs
+++ b/src/MoBi.Assets/ToolTips.cs
@@ -20,7 +20,7 @@ namespace MoBi.Assets
       public static readonly string AddToProject = "Save as a new building block in project.";
       public static readonly string ResetParameterToolTip = "Reset parameter to default";
       public static readonly string SaveSimulationSettingsToolTip = "Save current simulations settings as template into project or user profile. These settings will be used as default settings for each newly created simulation.";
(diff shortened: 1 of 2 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

Yuri05:
@rwmcintosh @PavelBal @msevestre LGFM
However, be aware that e.g. non-floating molecules (enzymes, metabolites, complexes) are **always** created with some global parameters set to NaN - so the warning will be always triggered here. Not sure what's the best way to deal with this.
image

PavelBal:
@Yuri05 Yes I think we should rework how (and where) we show warnings https://github.com/Open-Systems-Pharmacology/MoBi/issues/2513

Issue #2519, opened 2026-08-25
Add a --cores option to MoBi.CLI to control the degree of parallelism


rwmcintosh:
**TL;DR** — MoBi.CLI always uses `ProcessorCount − 1` cores for its parallel work (running simulations during snapshot loads, and model construction once #2424 lands) with no way to override it. Add a `--cores <n>` option so callers can control it — in particular the QualificationRunner, which spawns one MoBi.CLI process per MoBi project and will pass `--cores 1` so that parallelism is managed at the process level.

### Summary

The CLI host hardcodes its degree of parallelism at bootstrap (`MoBi.CLI.Core\MinimalImplementations\CoreUserSettings.cs:14`):

```csharp
public int MaximumNumberOfCoresToUse { get; set; } = Math.Max(Environment.ProcessorCount - 1, 1);
```

No MoBi.CLI verb exposes an option to change it. This one value drives:

- the parallel simulation runs during snapshot/project loads (`ProjectMapper.runParallelSimulations`),
- parallel model construction during snapshot loads (#2424),
- the batch simulation update (#2189, UI host uses the user setting instead).

When the QualificationRunner processes MoBi projects it spawns MoBi.CLI processes, so the same nested oversubscription applies as on the PK-Sim side: N concurrent processes × (cores − 1) workers each.

### Proposed work

- Add a `--cores <n>` option (n ≥ 1) to the shared CLI command base so every verb accepts it.
- When provided, set `ICoreUserSettings.MaximumNumberOfCoresToUse` after container initialization, before the runner starts.
- When absent, keep the current default (`ProcessorCount − 1`).

### Note on MoBi.R

`MoBi.R` registers this same `CoreUserSettings` class as a singleton (`RRegister.cs`), so R sessions inherit the `ProcessorCount − 1` default with no way to change it either. An R-facing setter is a natural follow-up but out of scope here.

### Design decision (agreed 2026-08-25)

Parallelism gets exactly one owner per level: the QualificationRunner manages process-level fan-out and passes `--cores 1` to each child; a standalone CLI invocation owns the whole machine by default. See the companion issues below.

### Companion issues

- PK-Sim side: Open-Systems-Pharmacology/PK-Sim#3711
- QualificationRunner side: Open-Systems-Pharmacology/QualificationRunner#189
- Motivation/measurements: Open-Systems-Pharmacology/PK-Sim#3710

PR #2520, opened 2026-08-25
Fixes #2424 Parallelize simulation model construction during snapshot load

### LINKED ISSUE #2424: Parallelize simulation model construction during snapshot/project load
### Summary

When a project is loaded from a snapshot, each simulation's model is **rebuilt sequentially**. In `ProjectMapper.MapToModel`:

```csharp
foreach (var x in projectSnapshot.Simulations)
{
   try
   {
      var simulation = await _simulationMapper.MapToModel(x, simulationContext);  // one at a time
      addSimulations(project, simulation);
      simulationContext.NumberOfSimulationsLoaded++;
   }
   catch (Exception e) { _logger.AddException(e); }
}
```

and `SimulationMapper.MapToModel` builds each one via `_simulationFactory.CreateSimulationAndValidate(...)` → `CreateModelAndValidate` → `_modelConstructor.CreateModelFrom`. For projects with many simulations this is the same per-simulation model-construction cost that motivated the parallel batch update in #2189 — but it runs serially here.

### Why this isn't already covered

The parallel path added in #2189 (`SimulationUpdateTask.ConfigureSimulationsInParallel`) is reachable **only** through the tree context-menu `UpdateSimulationUICommand` (single + multi select). The snapshot/project loading path does not use it.

Note: snapshot loading already parallelizes the simulation **run** (`ProjectMapper.runParallelSimulations` → `Parallel.ForEachAsync`), but only the solving — never the model **construction**.

### Proposed work

Apply the #2189 parallelism to the model-construction loop in `ProjectMapper.MapToModel`. The enablers are already in place:
- the thread-safe `ModelConstructor` (Open-Systems-Pharmacology/OSPSuite.Core#2887, the same one that made #2189 possible), and
- `IConcurrencyManager` (already used elsewhere for parallel runs).

### Things the parallel version must handle

The current `foreach` carries shared state that a concurrent version has to deal with:
- **`simulationContext`** mutable progress state — `NumberOfSimulationsLoaded` counter and the per-simulation `LoadingSimulation(name, N, M)` progress logging.
- **`addSimulations(project, …)`** mutates the project's simulation collection — needs to be thread-safe or marshaled.
- **`SimulationMapper.MapToModel` does more than construct the model** — charts, output mappings, output selections, parameter and scale-divisor updates — some of which touch shared context (`SnapshotContextWithSimulation`, `SimulationAnalysisContext`, observed data).
- **Per-simulation error isolation** — the current loop catches and logs per-simulation exceptions and continues; the parallel version must preserve that (a single bad simulation should not abort the whole load), mirroring the worker-level isolation used in `ConfigureSimulationsInParallel`.

### Context

Spun off from review of #2276 (the #2189 UI batch-update feature).

rwmcintosh:
Fixes #2424

> Requires OSPSuite.Core **13.0.165** (thread-safe `ModelConstructor` chain: Open-Systems-Pharmacology/OSPSuite.Core#2959, Open-Systems-Pharmacology/OSPSuite.Core#2965) — already referenced by this branch. Based on the merged #2522 prerequisite.

### What changes

- The per-simulation loop in `ProjectMapper.mapToModel` maps the first simulation alone (warming lazily initialized services), then the rest under `Parallel.ForEachAsync` bounded by `MaximumNumberOfCoresToUse`. Results are collected per index and committed to the project sequentially in snapshot order, preserving per-simulation error isolation. Because the same private `mapToModel` serves `MapToModel` and `MapToModelAndExportInputs`, this covers the UI snapshot load, the CLI `SnapshotRunner`, the qualification runner and MoBi.R in one change.
- The snapshot `SimulationMapper` suppresses `ShowProgress` only while the model is constructed (the core progress stream would interleave across parallel constructions) and restores the captured value on the configuration the simulation keeps — a snapshot-loaded simulation behaves like one loaded from a project file. The load reports its own progress: "Loading N simulations…", then "Simulation 'X' loaded (k/N)" per completion.
- Failures are named and accounted: a simulation that cannot be loaded logs the failing simulation, a snapshot referencing a module not in the project fails with a `MoBiException` naming the module (instead of an incidental `NullReferenceException`), and a load that lost simulations logs an "Only k/N simulations loaded" warning.
- `SimulationFactory` resolves its per-operation `ICloneManagerForBuildingBlock` per call instead of injecting a shared instance — the clone manager holds per-operation formula caches, and this was the last stateful dependency preventing one factory instance from serving parallel workers (same pattern as the existing `SimulationUpdateTask` worker isolation).
- `ServerGarbageCollection` is enabled for MoBi.exe and MoBi.CLI — parallel model construction is allocation-heavy and workstation GC stalls it. Measurements (PK-Sim side, same core construction path): Open-Systems-Pharmacology/PK-Sim#3710 — pure 60-simulation load 670 s sequential → 121 s parallel with server GC, and 9.2× vs the shipped release in the GUI.

### Behavior change

A qualification run whose snapshots silently lose a simulation now exits non-zero: the per-simulation failure logs an Error-level line naming the simulation, which the qualification verb's error counting picks up. Before this PR such a run reported success (an `OSPSuiteException`-class failure logged only a warning). The `snap` verb's exit code is unchanged.

### Tests

- Unit specs pin the progress suppression from both sides: the flag is `false` as seen by the factory during construction and the captured value (including `false`) is restored on the configuration the simulation keeps. Verified by mutation — removing either half fails its spec.
- A missing-module snapshot throws with the module name (unit spec), and the ProjectMapper failure spec asserts the surviving simulations keep the snapshot order, the log names the simulation and the missing module, and the partial-load warning fires — at `MaxDegreeOfParallelism = 4` so the parallel path actually runs.
- An integration spec loads four simulations sharing one 2,700-parameter module at DOP 4 and DOP 1 and compares the resulting models (entity paths, formula names and strings, dimensions, parameter values) plus entity-id uniqueness.
- All existing ProjectMapper integration specs (including the qualification input-export fixtures), SnapshotTask, CLI SnapshotRunner, SimulationFactory and DimensionValidator suites pass.

Related: `--cores` control for qualification workflows is #2519 / Open-Systems-Pharmacology/QualificationRunner#189 (children will run `--cores 1`).

### DIFF
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1503,6 +1503,8 @@ namespace MoBi.Assets
          public const string CreatingSimulation = "Creating...";
          public const string ConfiguringSimulation = "Configuring...";
          public const string UpdatingSimulation = "Updating...";
+         public static string LoadingSimulationsMessage(int simulationCount) => $"Loading {simulationCount} {"simulation".PluralizeIf(simulationCount)}...";
+         public static string SimulationsLoadedMessage(int loadedCount, int simulationCount) => $"{loadedCount}/{simulationCount} {"simulation".PluralizeIf(simulationCount)} loaded";
          public static readonly string AddParameter = "Add Parameter ";
          public static readonly string AddMolecule = "Add Molecule";
          public static readonly string LoadParameter = "Load Parameter ";
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1270,6 +1270,7 @@ namespace MoBi.Assets
          public static readonly string ErrorInFormula = "Error in Formula";
          public static readonly string AppliedMoleculeNotInProject = "Applied Molecule is not in Project";
          public static readonly string CouldNotCreateSimulation = "Unable to create the simulation. Please check warnings or errors in the notification view.";
+         public static string CannotLoadSimulation(string simulationName) => $"Cannot load simulation '{simulationName}'";
          public static readonly string TemplateShouldContainAtLeastOneCurve = "Template should contain at least one curve.";
          public static readonly string FrameworkExceptionOccurred = "An exception occurred";
          public static readonly string MergingSpatialStructuresIsNotSupported = "Merging Spatial Structures is not supported.";
@@ -1504,7 +1505,7 @@ namespace MoBi.Assets
          public const string ConfiguringSimulation = "Configuring...";
          public const string UpdatingSimulation = "Updating...";
          public static string LoadingSimulationsMessage(int simulationCount) => $"Loading {simulationCount} {"simulation".PluralizeIf(simulationCount)}...";
-         public static string SimulationsLoadedMessage(int loadedCount, int simulationCount) => $"{loadedCount}/{simulationCount} {"simulation".PluralizeIf(simulationCount)} loaded";
+         public static string SimulationsLoadedMessage(string simulationName, int loadedCount, int simulationCount) => $"Simulation '{simulationName}' loaded ({loadedCount}/{simulationCount})";
          public static readonly string AddParameter = "Add Parameter ";
          public static readonly string AddMolecule = "Add Molecule";
          public static readonly string LoadParameter = "Load Parameter ";
--- a/src/MoBi.Assets/AppConstants.cs
+++ b/src/MoBi.Assets/AppConstants.cs
@@ -1271,6 +1271,7 @@ namespace MoBi.Assets
          public static readonly string AppliedMoleculeNotInProject = "Applied Molecule is not in Project";
          public static readonly string CouldNotCreateSimulation = "Unable to create the simulation. Please check warnings or errors in the notification view.";
          public static string CannotLoadSimulation(string simulationName) => $"Cannot load simulation '{simulationName}'";
+         public static string ModuleNotFound(string moduleName) => $"Module '{moduleName}' not found in project";
          public static readonly string TemplateShouldContainAtLeastOneCurve = "Template should contain at least one curve.";
          public static readonly string FrameworkExceptionOccurred = "An exception occurred";
          public static readonly string MergingSpatialStructuresIsNotSupported = "Merging Spatial Structures is not supported.";
@@ -1506,6 +1507,7 @@ namespace MoBi.Assets
          public const string UpdatingSimulation = "Updating...";
          public static string LoadingSimulationsMessage(int simulationCount) => $"Loading {simulationCount} {"simulation".PluralizeIf(simulationCount)}...";
          public static string SimulationsLoadedMessage(string simulationName, int loadedCount, int simulationCount) => $"Simulation '{simulationName}' loaded ({loadedCount}/{simulationCount})";
+         public static string OnlySomeSimulationsLoadedMessage(int loadedCount, int simulationCount) => $"Only {loadedCount}/{simulationCount} {"simulation".PluralizeIf(simulationCount)} loaded";
          public static readonly string AddParameter = "Add Parameter ";
          public static readonly string AddMolecule = "Add Molecule";
          public static readonly string LoadParameter = "Load Parameter ";
--- a/src/MoBi.Assets/MoBi.Assets.csproj
+++ b/src/MoBi.Assets/MoBi.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" /> 
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.165" /> 
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.165" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/MoBi.Assets/MoBi.Assets.csproj
+++ b/src/MoBi.Assets/MoBi.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" /> 
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.161" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" /> 
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/MoBi.Assets/MoBi.Assets.csproj
+++ b/src/MoBi.Assets/MoBi.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.160" /> 
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.160" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.160" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" /> 
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/MoBi.Assets/MoBi.Assets.csproj
+++ b/src/MoBi.Assets/MoBi.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.160" /> 
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.160" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.160" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.164" /> 
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/MoBi.Assets/ToolTips.cs
+++ b/src/MoBi.Assets/ToolTips.cs
@@ -20,6 +20,7 @@ namespace MoBi.Assets
       public static readonly string AddToProject = "Save as a new building block in project.";
       public static readonly string ResetParameterToolTip = "Reset parameter to default";
       public static readonly string SaveSimulationSettingsToolTip = "Save current simulations settings as template into project or user profile. These settings will be used as default settings for each newly created simulation.";
+      public static readonly string WarnForNonFiniteQuantities = "Show a warning for each quantity whose value is NaN or infinite at time t=0.\nThe check is performed when a simulation is created and each time it is run.";
 
       public static class BuildingBlockMolecule
       {
--- a/src/MoBi.BatchTool/MoBi.BatchTool.csproj
+++ b/src/MoBi.BatchTool/MoBi.BatchTool.csproj
@@ -40,7 +40,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="DevExpress.Win.Design" Version="25.2.7" Condition="'$(ExcludeDesigner)' != 'true'" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.160" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
     <PackageReference Include="OSPSuite.SimModelSolver_CVODES" Version="5.0.0.53" />
--- a/src/MoBi.BatchTool/MoBi.BatchTool.csproj
+++ b/src/MoBi.BatchTool/MoBi.BatchTool.csproj
@@ -40,7 +40,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="DevExpress.Win.Design" Version="25.2.7" Condition="'$(ExcludeDesigner)' != 'true'" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.165" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
     <PackageReference Include="OSPSuite.SimModelSolver_CVODES" Version="5.0.0.53" />
--- a/src/MoBi.BatchTool/MoBi.BatchTool.csproj
+++ b/src/MoBi.BatchTool/MoBi.BatchTool.csproj
@@ -40,7 +40,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="DevExpress.Win.Design" Version="25.2.7" Condition="'$(ExcludeDesigner)' != 'true'" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
     <PackageReference Include="OSPSuite.SimModelSolver_CVODES" Version="5.0.0.53" />
--- a/src/MoBi.BatchTool/MoBi.BatchTool.csproj
+++ b/src/MoBi.BatchTool/MoBi.BatchTool.csproj
@@ -40,7 +40,7 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="DevExpress.Win.Design" Version="25.2.7" Condition="'$(ExcludeDesigner)' != 'true'" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.160" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
     <PackageReference Include="OSPSuite.FuncParser" Version="5.0.0.3" />
     <PackageReference Include="OSPSuite.SimModel" Version="5.0.0.80" />
     <PackageReference Include="OSPSuite.SimModelSolver_CVODES" Version="5.0.0.53" />
--- a/src/MoBi.CLI.Core/MoBi.CLI.Core.csproj
+++ b/src/MoBi.CLI.Core/MoBi.CLI.Core.csproj
@@ -16,13 +16,13 @@
 
 
 	<ItemGroup>
-      <PackageReference Include="OSPSuite.Assets" Version="13.0.160" />
-      <PackageReference Include="OSPSuite.Infrastructure.Autofac" Version="13.0.160" />
+      <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
+      <PackageReference Include="OSPSuite.Infrastructure.Autofac" Version="13.0.164" />
       <PackageReference Include="OSPSuite.Utility" Version="5.0.0.11" />
-      <PackageReference Include="OSPSuite.Core" Version="13.0.160" />
-      <PackageReference Include="OSPSuite.R" Version="13.0.160" />
-      <PackageReference Include="OSPSuite.Infrastructure" Version="13.0.160" />
-      <PackageReference Include="OSPSuite.Infrastructure.Castle" Version="13.0.160" />
+      <PackageReference Include="OSPSuite.Core" Version="13.0.164" />
+      <PackageReference Include="OSPSuite.R" Version="13.0.164" />
+      <PackageReference Include="OSPSuite.Infrastructure" Version="13.0.164" />
+      <PackageReference Include="OSPSuite.Infrastructure.Castle" Version="13.0.164" />
 	</ItemGroup>
 	<ItemGroup>
       <ProjectReference Include="..\MoBi.Assets\MoBi.Assets.csproj" />
--- a/src/MoBi.CLI.Core/MoBi.CLI.Core.csproj
+++ b/src/MoBi.CLI.Core/MoBi.CLI.Core.csproj
@@ -16,13 +16,13 @@
 
 
 	<ItemGroup>
-      <PackageReference Include="OSPSuite.Assets" Version="13.0.164" />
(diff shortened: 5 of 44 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05:
@rwmcintosh Please check :)

<details><summary><b>Review by Perplexity</b></summary>

Issues found, highest impact first.

### 1. `ShowProgress = false` leaks into the persisted simulation configuration

`SimulationMapper.MapToModel` sets the flag on the configuration *before* `CreateSimulationAndValidate`, and that method clones exactly that configuration onto the simulation (`cloneManager.Clone(configurationReferencingBuildingBlocks)`). `SimulationConfiguration.UpdatePropertiesFrom` → `CopyPropertiesFrom` copies `ShowProgress` ([[OSPSuite.Core SimulationConfiguration](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Builder/SimulationConfiguration.cs)](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Builder/SimulationConfiguration.cs)), so every snapshot-loaded simulation carries `ShowProgress = false` for the rest of the session. Consequences:

- The interactive update path does `CreateFromProjectTemplatesBasedOn(simulationToUpdate.Configuration)` → `CopyPropertiesFrom` → `false`. So the "Updating…" heavy-work run of a snapshot-loaded simulation gets **no** core progress — the PR description claims that path keeps it.
- `SimulationConfigurationDiffBuilder` compares `ShowProgress`, so comparing a snapshot-loaded simulation with a normally created one reports a spurious difference.
- The flag is *not* in `SimulationConfigurationXmlSerializer.PerformMapping`, so it silently flips back to `true` after a project save/reload — the behaviour differs between "loaded from snapshot" and "loaded from .mbp3".

Suggested fix: keep the suppression local to construction (set it on the configuration handed to the model constructor, or restore the previous value before the clone) rather than on the object that becomes `simulation.Configuration`.

### 2. The new spec never exercises the parallel path

`concern_for_ProjectMapper` builds the mapper with `_coreUserSettings = A.Fake<ICoreUserSettings>()`, so `MaximumNumberOfCoresToUse` returns `0` and `Math.Max(1, 0)` makes `MaxDegreeOfParallelism = 1`. `When_mapping_a_snapshot_with_multiple_simulations_to_project` therefore runs strictly sequentially, where `Parallel.ForEachAsync` preserves order regardless of the collect-then-commit logic — the ordering assertion is vacuous and would still pass if the new code were broken. Stub the setting (`A.CallTo(() => _coreUserSettings.MaximumNumberOfCoresToUse).Returns(4)`), and add a case where a middle simulation fails to actually cover the "order preserved + per-simulation isolation" claims.

### 3. Events are still published from worker threads with no marshalling

`SimulationFactory.showWarnings` publishes `ShowValidationResultsEvent` from whichever thread constructs, and now up to N threads do so concurrently. The handler is `NotificationPresenter.Handle` → `updateNotificationWith`, which does a non-atomic `_allNotifications = new NotifyList<…>(_allNotifications)` read-modify-write and then `_view.BindTo(...)` — lost notifications plus a DevExpress grid update off the UI thread. `QuantityValueInSimulationChangeTracker` has the same shape: `updateParameters`/`updateScaleDivisors` run inside the parallel body and publish `SimulationStatusChangedEvent` per change. [[Issue #2424](https://github.com/Open-Systems-Pharmacology/MoBi/issues/2424)](https://github.com/Open-Systems-Pharmacology/MoBi/issues/2424) lists this explicitly ("no marshalling guard") and the PR doesn't address it. The commit loop is already sequential — collecting validation results per index and publishing there would be a cheap fix.

### 4. Silent sequential fallback hides a misconfigured core count

`Math.Max(1, _userSettings.MaximumNumberOfCoresToUse)` degrades to fully sequential whenever the setting is `0`/unset. `MoBi.BatchTool`'s `BatchUserSettings.MaximumNumberOfCoresToUse` is an unset `int`, and the test fake is the same case — a host that forgets the setting silently loses the whole feature with no log line. Consider falling back to `Environment.ProcessorCount - 1` when the value is `<= 0`. The identical `ParallelOptions` block now appears twice in `ProjectMapper` (lines ~149 and ~208) and should be extracted.

### 5. Diagnosability regression in the progress/error reporting

The old message carried the simulation name (`Captions.LoadingSimulation(snapshot.Name, k, N)`); the replacement only reports counts, and the failure branch logs `_logger.AddException(e)` with nothing identifying which simulation failed — on a 60-simulation load that makes triage much harder. The `k/N` messages can also appear out of order since they are logged from the workers. Include `snapshots[index].Name` in both the completion message and the exception context.

### 6. Building blocks are shared across concurrently constructed configurations

`IModelConstructor.CreateModelFrom` documents concurrency as safe only "provided each call receives its own `SimulationConfiguration` whose building blocks are not shared mutably with another concurrent call" ([[ModelConstructor](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Services/ModelConstructor.cs)](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Domain/Services/ModelConstructor.cs)). But `SimulationConfigurationMapper` hands every configuration the same project instances (`ModuleByName`, `IndividualsCollection.FindByName`, `ExpressionProfileCollection.FindByName`), and `SimulationBuilder.performMerge` only clones builders when `RequiresBaseClone` is true — in the single-module case the project's own builders go straight into the builder caches, and the merge path explicitly "marks all entities in the source". Nothing in this PR or its tests demonstrates that two simulations built in parallel from one shared module produce the same model as sequential builds; that comparison (single-module and multi-module-merge) is the test worth adding, since it's the invariant the whole change rests on.

### 7. Smaller points

- `System.Threading.Interlocked` is fully qualified inline instead of adding the `using`, unlike the rest of the file.
- `SimulationContext.NumberOfSimulationsToLoad`/`NumberOfSimulationsLoaded` default to `1`, not `0`, and MoBi now never sets them — leaving misleading state on a Core type shared with PK-Sim rather than removing the MoBi usage cleanly.
- `SimulationUpdateTask.ConfigureSimulationsInParallel` — the other parallel construction path — still doesn't set `ShowProgress = false`, so the interleaving rationale is applied inconsistently. Its per-worker resolve block also remains, though the issue analysis said it becomes deletable once the clone manager is per-call.
- The `ParallelOptions` carries no `CancellationToken` and the body discards the one it receives, so a CLI/qualification cancel can't interrupt construction.
- Server GC was added to MoBi.exe and MoBi.CLI but not MoBi.BatchTool; also note the qualification runner spawns multiple MoBi.CLI children, each of which will now reserve server-GC heaps per core — worth checking peak memory for concurrent children even with `--cores 1`.
</details>

rwmcintosh:
1. Correct — fixed. ShowProgress is now disabled only while the model is constructed; the configuration cloned onto the created simulation gets the default back. A snapshot-loaded simulation therefore updates with core progress, compares clean in the diff builder and behaves like one loaded from a project file. Guarded by a new observation asserting the loaded simulations don't keep the suppression.

2. Correct — fixed. The mapper specs now stub MaximumNumberOfCoresToUse = 4 so the parallel path actually runs, and a new spec breaks a middle simulation's module reference and asserts the surviving simulations keep the snapshot order and the failure is logged with the simulation name.

3. Not applicable — the marshalling guard lives in the EventPublisher itself. Every publication from a non-UI thread is Post-ed to the registered SynchronizationContext; the GUI registers a WindowsFormsSynchronizationContext created on the UI thread, so all handlers — NotificationPresenter included — run serialized on the UI message loop. The read-modify-write is never concurrent and BindTo never runs off the UI thread. The only listeners of ShowValidationResultsEvent/SimulationStatusChangedEvent are presenters in MoBi.Presentation, so the headless hosts have no subscribers. This is also why the pre-existing fire-and-forget dimension-validation publications (which publish from a thread-pool continuation on develop today) have always been safe.

4. The fallback is unreachable in every real host. The GUI defaults MaximumNumberOfCoresToUse to ProcessorCount − 1, the CLI does the same, and MoBi.R reuses the CLI implementation; the BatchTool never loads snapshots. Falling back to ProcessorCount − 1 on 0 would silently enable full parallel construction in a host that never opted in — sequential is the safer degradation. The duplicated ParallelOptions block is extracted into a shared helper.

5. Correct — fixed. The completion message now names the simulation (Simulation 'X' loaded (3/60)) and a failed load logs an error naming the simulation, matching the PK-Sim loop. The k/N lines report completions from workers, so their ordering is not deterministic by design — each line is self-describing.

6. Partially correct — and the requested test was added. "Marking all entities in the source" writes into the SimulationBuilder's private _builderSources dictionary; no shared object is mutated. All merge property-writes target clones (the RequiresBaseClone path), and the single-module path only reads the project's builders. Sharing project instances across parallel constructions is also not new — ConfigureSimulationsInParallel (#2276) has done it since batch update. But agreed that nothing demonstrated the invariant: a new integration spec loads four simulations sharing one module (2,700+ parameters) at DOP 4 and DOP 1 and compares the resulting model structures, using the same entity-path/formula-name fingerprint as the Core concurrency test.

7.

Interlocked using added (also on the PK-Sim branch).
The SimulationContext counters can only leave Core once the PK-Sim loop PR merges (PK-Sim develop still writes them) — filed https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/issues/2966.
ConfigureSimulationsInParallel is intentionally out of scope. Note that adding ShowProgress = false there without the restore would introduce exactly the leak from point 1 (its ClonedConfiguration becomes the simulation's). The per-worker resolves are still required: SimulationConfigurationFactory constructor-injects the stateful clone manager.
There is no cancellation source in the mapping path to wire — the sequential code could not be cancelled either; the qualification runner cancels at process level.
The BatchTool never reaches the snapshot load path. The concurrent-children memory point is fair: on net10, DATAS is on by default with server GC so idle heaps shrink, and process-level parallelism is what #2519 / https://github.com/Open-Systems-Pharmacology/QualificationRunner/issues/189 will manage.

Yuri05:
@rwmcintosh Next :)

<details><summary><b>Review by Perplexity No. 2</b></summary>

## Re-review after 357b447 (+ Core 13.0.165)

Verified against the updated branch. Points 1, 2, 5 and the `Interlocked`/duplicate-`ParallelOptions` nits are genuinely fixed. On point 3 you are right and I was wrong: `EventPublisher.PublishEvent` snapshots listeners under lock and `Post`s to the registered `SynchronizationContext` when the publishing thread isn't the captured UI thread ([[OSPSuite.Utility EventPublisher](https://github.com/Open-Systems-Pharmacology/OSPSuite.Utility/blob/develop/src/OSPSuite.Utility/Events/EventPublisher.cs)](https://github.com/Open-Systems-Pharmacology/OSPSuite.Utility/blob/develop/src/OSPSuite.Utility/Events/EventPublisher.cs)), and `MoBi.UI.ApplicationStartup` registers the `WindowsFormsSynchronizationContext` plus the UI thread — so `NotificationPresenter` is never entered concurrently. Withdrawn. Point 4 (fallback) and the out-of-scope items I accept as answered.

Remaining issues, all in the new code.

### 1. The suppression itself is still untested

The only new assertion is the post-condition (`ShowProgress` is `true` on the loaded simulations). Delete `configuration.ShowProgress = false;` and every spec still passes, while the interleaved progress stream the PR exists to fix comes back. As written, the `false` … `true` pair also reads as a no-op to anyone who doesn't read the comments. A spec that captures the flag as seen by `IModelConstructor.CreateModelFrom` (fake constructor, assert `configuration.ShowProgress` is `false` at construction time and `true` afterwards) would lock both halves of the invariant, and is the one that actually guards the feature.

### 2. The parallel/sequential fingerprint is weaker than the risk it covers

`fingerprintOf` compares sorted `EntityPath|Formula.Name` pairs. That catches misattached entities, but not the failure modes a shared-builder race would most plausibly produce: wrong `IParameter.Value` / `Formula.ToString()`, wrong dimension, or duplicate ids (ids are deliberately excluded). Adding the value/formula-string and a `GroupBy(id).Count()` uniqueness check costs nothing on the same traversal.

Related: one execution at DOP 4 with three concurrent constructions is a smoke test, not a race detector — it will pass on a broken build most runs. And the baseline is the same `Parallel.ForEachAsync` at DOP 1, not the pre-PR sequential path, so any order-dependence inside the loop body itself is invisible to the comparison.

Also in that spec: the two projects are loaded back-to-back without closing the first, so both live in the shared `MoBiContext`/`WithIdRepository` at once (double memory on a 2,700-parameter model ×8, and `_project` silently ends up being `_sequentialProject`), and each run leaks a `Guid`-named JSON into `%TEMP%` that is never deleted.

### 3. The restore hardcodes `true` instead of the previous value

```csharp
mobiSimulation.Configuration.ShowProgress = true;
```

Today `SimulationConfigurationMapper` never maps the flag, so the default is `true` and this is correct by coincidence. Capturing it (`var showProgress = configuration.ShowProgress;` … restore `showProgress`) makes the pair self-evidently a save/restore and survives the flag becoming part of the snapshot. Same block: `simulation as MoBiSimulation` is still dereferenced unguarded on the next line — pre-existing, but the new line moved it one statement earlier.

### 4. The new failure spec relies on an incidental `NullReferenceException`

Setting `Module = "does-not-exist"` makes `context.MoBiProject().ModuleByName(...)` return `null`, and the failure is then a `NullReferenceException` thrown inside `new ModuleConfiguration(module)` (the ctor dereferences `module.InitialConditionsCollection`). The spec asserts only that some error mentioning `simulation-2` was logged, so it would pass for any accidental crash anywhere in the mapping — it doesn't pin "module not found". More importantly it documents the production behaviour: a snapshot referencing a missing module now logs a raw NRE stack next to "Cannot load simulation 'x'". Since the loop swallows per-simulation failures, that stack is all the user gets. A guarded lookup throwing `MoBiException($"Module '{snapshot.Module}' not found")` in `ModuleConfigurationMapper` would make both the log and the test meaningful.

### 5. Log-category inconsistency and double entries per failure

`_logger.AddError(CannotLoadSimulation(...))` omits the `projectSnapshot.Name` category that every other line in this method passes, so the failure lands in a different log category than the "Loading…"/"loaded" lines it belongs with. The new assertion uses `A<string>._` for the category, so it won't catch it. Each failure also produces two entries (`AddException` + `AddError`); folding the name into the exception log, or logging the exception at debug, would keep one user-facing line.

### 6. Pre-existing, but now more visible: a partial load still reports success

Not introduced here — `develop` already caught per-simulation exceptions — but with the new named error it becomes obvious that `SnapshotRunner.createProjectFromSnapshotFile` saves the `.mbp3` and exits 0 for a project that is silently missing simulations; the run only fails when the whole project load throws. A final "3/4 simulations loaded" warning (or a non-zero exit for the CLI snapshot mode) would close it, and the loop already has the counter to do it.

### 7. Minor

- `parallelOptions()` silently yields DOP 1 whenever the setting is 0; an `AddDebug($"Constructing simulations with {dop} core(s)")` makes the degradation diagnosable without changing the (reasonable) safe-default behaviour.
- `ProjectLoadSpecs.cs` still ends without a trailing newline, and the new `using`s are inserted out of the file's existing grouping (`Microsoft.Extensions.Logging` after the `MoBi.Core.*` block in `ProjectMapperSpecs.cs`, `MoBi.HelpersForTests` before `MoBi.Core.Domain.Model`).
- `snapshotWithSimulationCopies(..., numberOfCopies: 3)` producing four simulations via `for (var i = 2; i <= numberOfCopies + 1; i++)` is easy to misread; `numberOfSimulations` as the parameter would match the assertion.
</details>

rwmcintosh:
1. Correct — the suppression is now pinned from both sides. A new unit spec for the snapshot SimulationMapper captures the flag as seen by the factory at construction time: it asserts false while the model is constructed and restored on the configuration the simulation keeps. Verified by mutation — deleting the suppression fails the first observation, deleting the restore fails the second (plus the integration guard).

2. The cheap additions are in; the framing I'd defend. The fingerprint now carries the formula string, dimension, and parameter value per entity, and a new observation checks entity-id uniqueness. On the rest: one DOP-4 execution is an equivalence test by design, not a race detector — that duty lives in Core's dedicated 8-way concurrency fingerprint test; and the DOP-1 baseline through the same loop isolates concurrency as the only variable (the pre-PR loop isn't on this branch to run side-by-side, and the commit order is pinned by the order assertions). One correction: the two projects do not coexist in the shared repository — NewProject() calls Clear() → ObjectRepository.Clear() before the second load; the first project survives only as a spec-held object graph, which the comparison requires. The temp-file leak was real — the spec deletes it in Cleanup().

3. Fixed — the mapper captures and restores the flag, so the pair reads as a save/restore and survives the flag ever becoming part of the snapshot. The as MoBiSimulation deref stays unguarded: the factory contract returns a MoBiSimulation, and the codebase style avoids defensive guards on guaranteed casts.

4. Fixed as suggested. ModuleConfigurationMapper now throws MoBiException "Module 'x' not found in project" on a missing module, with a unit spec pinning the throw. The failure spec asserts the log names the missing module — note it checks the Warning level, because AddException deliberately logs OSPSuiteExceptions as warnings with the clean message and no stack trace.

5. Category fixed — both catch-block lines now carry the project category. The two entries per failure are intentional: the exception detail for diagnosis plus one readable line naming the simulation, the same shape as the PK-Sim loop.

6. The warning is in — a load that lost simulations now logs "Only k/N simulations loaded" under the project category, asserted in the failure spec (and mirrored in the PK-Sim loop branch). On the exit code: the qualification verb already exits non-zero for this case today — Program wires the ErrorCountingLoggerProvider for qualification runs, and the per-simulation failure logs at Error level, which flips the return code. The snap verb's exit-0 on partial loss predates this PR (the same accounting family as Open-Systems-Pharmacology/PK-Sim#3666); if its return code should change, that's a separate issue and PR rather than a regression fix here.

7. All three taken — parallelOptions() logs the degree of parallelism at debug level, the using orders and trailing newline are fixed, and the parameter is now numberOfSimulations with a matching loop bound.

Yuri05:
@rwmcintosh 
<details><summary><b>Review by Perplexity No. 3</b></summary>

### Re-review after 29c56b4

Verified in source. Points 1, 3, 4, 5, 6 and the three nits are fixed as described, and the two claims I'd have pushed back on check out: `MoBiException` is an `OSPSuiteException` and not one of the excluded types, so `IsInfoException` is true and `AddException` logs it as a warning with the clean message ([[OSPSuite.Core LoggerExtensions](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Services/LoggerExtensions.cs)](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Services/LoggerExtensions.cs), [[ExceptionExtensions](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Extensions/ExceptionExtensions.cs)](https://github.com/Open-Systems-Pharmacology/OSPSuite.Core/blob/develop/src/OSPSuite.Core/Extensions/ExceptionExtensions.cs)); and `SnapshotTask` calls `_moBiContext.NewProject()` → `Clear()` → `ObjectRepository.Clear()` per load, so the two projects don't coexist in the repository. Withdrawn.

What's left is small.

### 1. Qualification runs now exit non-zero where they previously passed

`Program.cs:52` wires `ErrorCountingLoggerProvider` for `QualificationRunCommand` only, and the new `_logger.AddError(CannotLoadSimulation(...))` is the first Error-level line on this path: before this PR a simulation that failed with a `MoBiException` produced `AddException` → **warning** only, so the qualification verb returned 0. Any pipeline whose snapshots have been quietly losing a simulation will start failing. That's very likely the behaviour you want, but it is a behaviour change that isn't mentioned in the description and deserves a line in the PR text (or the release notes).

### 2. The capture/restore introduced in this commit isn't what the spec pins

`should_restore_the_progress_flag_on_the_configuration_kept_by_the_simulation` starts from a default `SimulationConfiguration` and asserts `ShouldBeTrue()` — identical result for the previous hardcoded `= true`. Reverting `showProgress` back to a literal `true` passes both new observations. A second case (or a parameterised one) starting from `_configuration.ShowProgress = false` and asserting it stays `false` is what actually pins the save/restore you just wrote.

### 3. `parallelOptions()` logs without a category and twice per load

The same commit standardised every log line in `ProjectMapper` on the `projectSnapshot.Name` category, but the new `_logger.AddDebug($"Parallel operations limited to {n} core(s)")` passes none, so it lands in a different category than the load it describes. It also fires from both call sites with nothing distinguishing construction from the simulation run, and a method named `parallelOptions()` now has a logging side effect. `parallelOptions(string category, string phase)` — or logging at the call sites — keeps the factory pure and the lines attributable.

### 4. `Cleanup()` deletes a file owned by `GlobalContext`

The temp snapshot is created once in `GlobalContext` but deleted in `Cleanup()`, which runs after every observation (five in that class); `ProjectTaskIntegrationTests` in the same folder uses `GlobalCleanup()` for exactly this. Harmless today only because `File.Delete` ignores a missing path.

### 5. The missing-module spec depends on the mapper staying synchronous

`The.Action(() => sut.MapToModel(_snapshot, _context)).ShouldThrowAn<MoBiException>()` works because `ModuleConfigurationMapper.MapToModel` returns `Task.FromResult` and throws before it. If it ever becomes `async`, the throw moves into the returned Task, which the lambda discards — the spec then passes on a broken mapper (and leaves an unobserved faulted task). Awaiting the call in the assertion removes that dependency.

### 6. Two accuracy notes on the failure path

- The spec comment "an OSPSuite exception is logged as a warning with its clean message" holds for `MoBiException`, but `IsInfoException` explicitly excludes `NotFoundException` and `QualificationRunException`, which log at Error with a clean message. Worth narrowing the comment to info exceptions so nobody generalises it.
- What actually reaches the `catch` in `mapSimulationAt` is an `AggregateException`: `SimulationConfigurationMapper` calls `_moduleConfigurationMapper.MapToModels(...).Result`, and `MapTo` is `async`, so the synchronous throw is captured into the task. `ExceptionMessage`/`IsInfoException` unwrap it, so level and text are right — but only `InnerException` (the first failure) is reported, so a simulation with two bad module references names only one of them.
</details>

rwmcintosh:
1. Correct - fixed. My round-2 claim that qualification already exited non-zero was only true for hard exceptions, as you describe. The PR description now carries a Behavior change section; the snap verb is unchanged.

2. Correct - fixed. Reverting the restore to a literal true now fails the new fixture (verified).

3. Correct - fixed.

4. Correct - fixed.

5. Correct - fixed.

6. Both taken: comment narrowed. The first-failure-only reporting is the suite-wide unwrap behavior, not introduced here — leaving aggregate reporting to a broader change.

Yuri05:
## Re-review after 5222eae

Verified all six items in source. The debug lines now carry the project category (`project.Name` is assigned from `projectSnapshot.Name` at the top of `mapToModel`, so both phases land in the same category), `parallelOptions()` is a pure factory again, the missing-module spec awaits through `GetAwaiter().GetResult()` so the throw is observed, the temp file moved to `GlobalCleanup()`, and the new `When_mapping_a_simulation_snapshot_whose_configuration_disables_progress` fixture does pin the restore to the captured value — a literal `true` would fail it. The Behavior change section is in the description.

Only two cosmetic points remain, neither blocking.

### 1. One duplicated observation is vacuous in the new fixture

`should_suppress_the_core_progress_while_the_model_is_constructed` is copied verbatim into both derived fixtures. In `When_mapping_a_simulation_snapshot_whose_configuration_disables_progress` the configuration already starts at `false`, so that observation passes even if the mapper never touches the flag — it asserts nothing there. Either drop it from the second fixture (the restore observation is the one that carries the weight) or move it onto the abstract base, where NUnit runs it once per derived fixture without the copy.

### 2. The construction debug line fires for single-simulation snapshots

`_logger.AddDebug($"Constructing simulations with up to {n} core(s)")` is emitted inside the `Simulations.Length > 0` block, before a `Parallel.ForEachAsync` over `Enumerable.Range(1, 0)` — a one-simulation snapshot logs a parallelism figure for a loop that never runs a worker. Guarding it on `snapshots.Length > 1` keeps the debug log honest.

Yuri05 (approved):

PR #2517, opened 2026-08-25
Fixes #2514 Allow deletion of the MoleculeProperties container

rwmcintosh:
Fixes #2514

Follow-up to #2504. `Delete` was suppressed for every container whose name is "special" (`ContextMenuFor.InitializeWith` returns early for `MoleculeProperties` and `Neighborhoods`), so the only way to remove a `MoleculeProperties` container was to switch its parent to `Logical`.

`Delete` is now offered for any `MoleculeProperties` that has a parent container — under a physical parent, under a logical parent, and inside a neighborhood. It routes through the normal `RemoveCommandForContainer`, so the usual confirmation dialog, diagram cleanup and undo all apply.

## Summary by CodeRabbit

- **New Features**
  - Added a Delete option for eligible molecule-properties containers that have a parent container.
  - Added support for displaying this option across physical, logical, neighborhood, and global spatial-structure contexts.

- **Bug Fixes**
  - Corrected context menus for special-name containers so applicable deletion actions are available.

- **Tests**
  - Added coverage verifying Delete menu behavior for supported container types and scenarios.

PR #2521, opened 2026-08-26
Fixes #2512 Do not create the global MoleculeProperties container by default

### LINKED ISSUE #2512: Do not create `MoleculeProperties` container by default in a new SS-BB
When creating a new Spatial Structure Building Block, the `MoleculeProperties` container is added automatically. This should be avoided, so the `Overwrite` mode does not remove the molecule properties of the previous containers.

Instead, offer an action to add this container to the top level.

rwmcintosh:
A new spatial structure no longer carries an empty global `MoleculeProperties` container, so it stops overwriting the ones defined by previous modules when the module merge behavior is `Overwrite`. The container is no longer created by `SpatialStructureFactory` in Open-Systems-Pharmacology/OSPSuite.Core#2961, and `SpatialStructureMerger` already skips it when it is `null`.

In its place:
- **Add** — `AddMoleculeProperties` on the spatial structure root context menu, offered only while the container is absent.
- **Delete** — on the container's own node, wired to a dedicated command. It plugs into the `AddMenuItemsForSpecialName` hook from #2517, which deliberately left the global container out because the generic remove would have dropped the tree node without removing the data.

Both are undoable (`AddGlobalMoleculePropertiesCommand` / `RemoveGlobalMoleculePropertiesCommand`), and `HierarchicalSpatialStructurePresenter.entityIsInSpatialStructure` now recognises the global container so the tree refreshes on add and remove.

Also fixes two null paths that a spatial structure without the container would have hit: `ParameterResolver.resolveMoleculePropertiesInSpatialStructure` and `HierarchicalSpatialStructurePresenter.InitializeWith`.

Bumps OSPSuite.Core to 13.0.161, which this depends on.

Fixes #2512

## Summary by CodeRabbit

* **New Features**
  * Added the ability to create and remove global molecule-dependent properties for spatial structures.
  * Added context-menu actions for managing these properties, including confirmation before removal.
  * Spatial structure views now display global molecule properties when available.

* **Bug Fixes**
  * Improved handling of spatial structures without global molecule properties, preventing empty or invalid tree entries.
  * Safely handles missing global molecule properties during parameter resolution.

* **Tests**
  * Added coverage for creation, removal, cancellation, duplicate prevention, and spatial structure display behavior.

### DIFF
--- a/src/MoBi.Assets/MoBi.Assets.csproj
+++ b/src/MoBi.Assets/MoBi.Assets.csproj
@@ -22,9 +22,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="OSPSuite.Assets" Version="13.0.160" /> 
-    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.160" />
-    <PackageReference Include="OSPSuite.Core" Version="13.0.160" />
+    <PackageReference Include="OSPSuite.Assets" Version="13.0.161" /> 
+    <PackageReference Include="OSPSuite.Assets.Images" Version="13.0.161" />
+    <PackageReference Include="OSPSuite.Core" Version="13.0.161" />
   </ItemGroup>
 
   <ItemGroup>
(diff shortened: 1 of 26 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #2522, opened 2026-08-26
Fixes #2518 Create the dimension validation state per run

### LINKED ISSUE #2518: DimensionValidator: fire-and-forget validations can overlap on a shared stateful instance
**TL;DR** — Dimension validation runs on a background task that nobody awaits, on a validator instance whose state is shared by every simulation in a snapshot load. So even today, loading a multi-simulation snapshot can attribute warnings to the wrong simulation or throw a `NullReferenceException`, in every host (desktop, CLI, MoBi.R). MoBi.R also carries a Singleton registration of this validator that turns out to be dead code — but a trap. This is a prerequisite for #2424, and fixing it lets `SimulationUpdateTask` return to plain constructor injection.

### The stateful validator

`DimensionValidator` holds per-run state in instance fields (`DimensionValidator.cs:23-27`):

```csharp
private bool _checkDimensions;
private SimulationBuilder _simulationBuilder;
private ValidationResult _result;
```

`Validate` (`:63-82`) wraps itself in a `Task.Run`, assigns both fields, traverses every entity of the model via `AcceptVisitor(this)`, and nulls the fields in a `finally`. Every `Visit` callback funnels into `addNotification` (`:134-141`), which dereferences both fields.

### Defect 1 — validation is fire-and-forget

`SimulationFactory.cs:97-101`:

```csharp
private void validateDimensions(IModel model, SimulationBuilder simulationBuilder)
{
   _dimensionValidator.Validate(model, simulationBuilder)
      .SecureContinueWith(t => showWarnings(t.Result));
}
```

`SecureContinueWith` only attaches continuations — nothing awaits. `CreateModelAndValidate` (`:103-118`) kicks off validation and returns while a pool thread is still walking the model.

### Defect 2 — one validator instance serves an entire snapshot load

The snapshot `SimulationMapper` holds a single `ISimulationFactory` for the whole project load (`SimulationMapper.cs:27`), which holds a single `IDimensionValidator`. Combined with defect 1, the failure needs no parallel loop at all:

1. The sequential loop maps simulation N — its `Validate` starts traversing on a pool thread.
2. The loop proceeds to simulation N+1, builds its model, reaches `validateDimensions` — a second `Validate` starts **on the same instance**.
3. Timing lottery: the second call's `_result = new ValidationResult()` clobbers the first (N's warnings silently land in N+1's result), or the first call's `finally` nulls `_result`/`_simulationBuilder` mid-traversal of the second — `NullReferenceException` in `addNotification` (`_simulationBuilder.BuilderFor(...)` at `:136`).

Related: the continuation publishes `ShowValidationResultsEvent` from the pool thread (`SimulationFactory.cs:136-139`) — in the desktop host, UI subscribers receive it off the UI thread; in the #2189 parallel-update path, N of these can fire concurrently.

The #2189 path avoids the per-instance overlap only because each worker resolves its own `ISimulationFactory` and uses its validator once (`SimulationUpdateTask.cs:145-157`). Parallelizing the snapshot loop (#2424) with the current mapper-held factory would turn the overlap from rare to routine.

### The dead Singleton registration in MoBi.R

`RRegister.cs:64` registers the validator as Singleton:

```csharp
container.Register<IDimensionValidator, DimensionValidator>(LifeStyle.Singleton);
```

But the container resolves the **first** registration for a service, and `CoreRegister`'s assembly scan — which registers `DimensionValidator` transient (it is not in the exclude list, `CoreRegister.cs:41-62`) — runs earlier in the MoBi.R bootstrap (`Bootstrap\ApplicationStartup.cs:23-25`). So this line is silently ineffective. As written it documents a lifestyle that is not in effect, and if a future bootstrap reordering ever made it win, every consumer in the MoBi.R process would share one mutable validator. It should be removed.

### Proposed work

- [ ] Make `Validate` stateless: the visitor-callback signatures cannot take extra parameters, so extract the traversal into a per-call visitor object instantiated inside `Validate` — the same pattern Open-Systems-Pharmacology/OSPSuite.Core#2887 used (e.g. `FormulaOriginVisitor`, "created per call so that the task remains stateless"). Lifestyle then stops mattering.
- [ ] Remove the ineffective Singleton registration in `RRegister.cs`.
- [ ] Decide whether dimension validation should be awaited, so warnings deterministically belong to the simulation that produced them and `ShowValidationResultsEvent` is not published from arbitrary pool threads.

### Follow-on simplification enabled by this fix

`SimulationUpdateTask.configureSimulation` (`:145-157`) resolves `ISimulationConfigurationFactory`, `ISimulationFactory`, and `ICloneManagerForBuildingBlock` per worker, with the comment "resolve services per-instance because some services are not stateless". Tracing that graph, only two leaves are actually stateful: `DimensionValidator` (this issue) and `CloneManagerForBuildingBlock` (per-operation formula caches, `CloneManagerForBuildingBlock.cs:45-47`); `SimulationFactory` and `SimulationConfigurationFactory` themselves hold only readonly fields. With the validator stateless and the clone manager given the per-call-instance treatment (the `DynamicFactory` pattern Core uses for `IModelValidatorFactory`), all three services become safely constructor-injectable — the resolve-per-worker block in `SimulationUpdateTask` can be deleted, and the parallel snapshot loop in #2424 needs no per-worker resolution either.

**Correction (2026-08-27, after the review of #2520):** the clone manager did not get the per-call-instance treatment. What shipped in #2522 is narrower: `SimulationFactory` resolves `ICloneManagerForBuildingBlock` per call inside its own methods, while `SimulationConfigurationFactory` still constructor-injects the stateful instance (used by `CreateFromProjectTemplatesBasedOn`) and `configureSimulation` also uses one directly for `Clone(templateConfiguration)`. The resolve-per-worker block therefore stays — it is what gives each worker its own clone manager instances; only the `ISimulationFactory` resolve became redundant. The parallel snapshot loop in #2424 is unaffected: it reaches the clone manager only through `SimulationFactory`'s per-call resolve.

rwmcintosh:
Fixes #2518

> Requires OSPSuite.Core **13.0.164** (contains Open-Systems-Pharmacology/OSPSuite.Core#2959) — already referenced by this branch.

### What changes

`DimensionValidator` no longer holds per-run state: the visiting state moves into a `ValidationRun` visitor created inside each `Validate` call, so concurrent validations on the same instance stay independent. The ineffective singleton registration of `IDimensionValidator` in `MoBi.R` is removed.

### Why

Dimension validation runs on `Task.Run` and is never awaited (`SimulationFactory.validateDimensions` only attaches continuations via `SecureContinueWith`), while the snapshot `SimulationMapper` holds a single `ISimulationFactory` — and therefore a single `IDimensionValidator` — for an entire project load. So even in today's sequential loop, simulation N+1's validation can begin on the same instance while simulation N's is still traversing: the second `_result = new ValidationResult()` clobbers the first, and the first call's `finally` nulls `_result`/`_simulationBuilder` mid-traversal of the second, producing a `NullReferenceException` in `addNotification` or warnings attributed to the wrong simulation.

The `MoBi.R` registration was silently ineffective anyway — the container resolves the first registration for a service, and `CoreRegister`'s scan (transient) runs earlier in the MoBi.R bootstrap — so it documented a lifestyle that was not in effect.

This is a prerequisite for #2424 (parallel model construction during snapshot load).

### Tests

A new spec runs two validations concurrently on the same validator with a deterministic interleaving and asserts each reports only its own messages. It fails against the previous implementation.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved dimension validation so simultaneous validation operations remain isolated and return the correct messages for each operation.
  - Removed an unnecessary validator registration from the R integration.

- **Chores**
  - Updated OSPSuite-related components across the application and test projects to version 13.0.164.

- **Tests**
  - Added coverage for concurrent dimension validation scenarios to help prevent cross-operation result mixing.

### DIFF
--- a/src/MoBi.Core/Services/DimensionValidator.cs
+++ b/src/MoBi.Core/Services/DimensionValidator.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
--- a/src/MoBi.Core/Services/DimensionValidator.cs
+++ b/src/MoBi.Core/Services/DimensionValidator.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -11,6 +11,7 @@ using OSPSuite.Core.Domain.Formulas;
 using OSPSuite.Core.Domain.UnitSystem;
 using OSPSuite.FuncParser;
 using OSPSuite.Utility.Extensions;
+using OSPSuite.Utility.Visitor;
 using IMoBiCoreUserSettings = MoBi.Core.ICoreUserSettings;
 
 namespace MoBi.Core.Services
@@ -20,11 +21,7 @@ namespace MoBi.Core.Services
       private readonly DimensionParser _dimensionParser;
       private readonly IObjectPathFactory _objectPathFactory;
       private readonly IMoBiCoreUserSettings _userSettings;
-      private bool _checkDimensions;
-      private SimulationBuilder _simulationBuilder;
-      private ValidationResult _result;
       private readonly IDictionary<string, string> _hiddenNotifications;
-      private bool _checkRules;
 
       public DimensionValidator(DimensionParser dimensionParser, IObjectPathFactory objectPathFactory, IMoBiCoreUserSettings userSettings)
       {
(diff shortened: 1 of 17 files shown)


src/MoBi.Core/Services/DimensionValidator.cs
```diff
@@ -64,164 +61,178 @@ public Task<ValidationResult> Validate(IEnumerable<IContainer> containers, Simul
...
-               _result = null;
-            }
+            var validationRun = new ValidationRun(_dimensionParser, _objectPathFactory, _userSettings, _hiddenNotifications, simulationBuilder);
+            containers.Each(c => c.AcceptVisitor(validationRun));
+            return validationRun.Result;
          });
       }
 
-      public void Visit(IEntity entity)
+      private class ValidationRun : IVisitor<IEntity>, IVisitor<IUsingFormula>
```
rwmcintosh, line 70: The validation logic itself is unchanged. This private class is constructed to make the validation state not part of the service, but part of the individual call to `Validate`. Unreachable from parallel calls.

rwmcintosh:
@coderabbitai review please

tests/MoBi.Tests/Core/DimensionValidatorSpecs.cs
```diff
@@ -46,7 +131,7 @@ protected override void Context()
 
          var lengthDimension = new Dimension(new BaseDimensionRepresentation {LengthExponent = 1}, AppConstants.DimensionNames.LENGTH, "cm");
          var molWeightDimension = new Dimension(new BaseDimensionRepresentation { AmountExponent = 1, MassExponent = -1 }, "MW", "mol/g");
-         var areaDimension = new Dimension(new BaseDimensionRepresentation { LengthExponent = 2,}, AppConstants.DimensionNames.AREA, "cm�");
+         var areaDimension = new Dimension(new BaseDimensionRepresentation { LengthExponent = 2,}, AppConstants.DimensionNames.AREA, "cm�");
```
Yuri05: encoding problem?

Yuri05:
s. CodeRabbit suggestion for the test improvement:
image

Yuri05:
This suggestion by CodeRabbit makes sense as well I think:
image

tests/MoBi.Tests/Core/DimensionValidatorSpecs.cs
rwmcintosh: yes. done

rwmcintosh:
All implemented

Yuri05 (approved):

PR #2524, opened 2026-08-28
Harden the parallel snapshot loading (mirror of PK-Sim#3712)

### LINKED ISSUE #2523: Parallel snapshot construction: mirror PK-Sim#3712
Apply the equivalent of Open-Systems-Pharmacology/PK-Sim#3712 (parallel snapshot construction and the review-round hardening: warm-up gating, out-of-memory containment, cancellation logging) to MoBi's snapshot loading.

rwmcintosh:
Fixes #2523

Follow-up to #2520, mirroring the fixes that came out of the review rounds on Open-Systems-Pharmacology/PK-Sim#3712 (merged) where they apply to MoBi.

### What changes

- Snapshot simulations are mapped sequentially until one succeeds, so lazily initialized services are warm before the remaining simulations are constructed in parallel — this also addresses the two post-merge findings on #2520 (a failing first simulation no longer forfeits the warm-up; per-simulation error isolation is kept).
- A simulation that maps to nothing no longer counts as loaded: the project receives only mapped simulations and the "Only k/N simulations loaded" warning fires.
- Out-of-memory failures never degrade the load silently: a shared `Exception.IsOutOfMemory()` extension (iterative, cycle-detecting via a visited set, unwrapping `AggregateException` and inner-exception chains of any width) backs every broad catch on the load path. A load-phase OOM fails the load; a run-phase OOM aborts the runs with a named error but keeps the loaded project, and the analyses then map as in a no-run load.
- Cancelled runs are logged — except the collateral cancellations that follow an OOM abort, which stay quiet.
- The construction/run degree of parallelism and per-simulation construction starts are logged at debug level, all via `AppConstants`.

Not mirrored: everything `StartableWarmup`-related from the PK-Sim reviews — MoBi has no startable warm-up; the sequential-until-first-success loop is its warm-up and the fan-out is already gated on it.

### Tests

- A handshake spec pins sequential-until-first-success: the third mapping must not start while the second — not yet successful — is still running (the guard was verified by mutation).
- Fixtures for the null mapping (warning fires, order preserved), the load-phase OOM (OOM stays in the exception chain, no per-simulation load error), the run-phase OOM (project kept, aborted-runs error logged) and a cancelled run (cancellation logged).
- `ExceptionExtensions` specs cover direct, nested, aggregate (including one 64 siblings wide), null and cyclic exception chains.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved project loading so simulations continue efficiently after initial successful setup, while failed simulations are excluded from progress totals.
  - Added more reliable handling of out-of-memory conditions during simulation loading and execution.
  - Analyses are now handled appropriately when simulation runs are stopped due to insufficient memory.
  - Improved distinction between user cancellations and cancellations caused by memory-related failures.
- **User Experience**
  - Added clearer status messages for simulation construction, execution, cancellation, and core usage.


rwmcintosh:
@coderabbitai review please

rwmcintosh:
@Yuri05 as per the latest changes from PKSim

src/MoBi.Core/Extensions/ExceptionExtensions.cs
Yuri05, line 1: ok for now, but duplicates https://github.com/Open-Systems-Pharmacology/PK-Sim/blob/develop/src/PKSim.Core/Extensions/ExceptionExtensions.cs and should be moved to Core

Yuri05 (approved):

src/MoBi.Core/Extensions/ExceptionExtensions.cs
rwmcintosh: done

PR #2526, opened 2026-08-31
Fixes #1614 Allow selecting the module when committing simulation changes

### LINKED ISSUE #1614: Commit changes to BBs - allow the selection of the module and the PV/IC BB
Follow up on https://github.com/Open-Systems-Pharmacology/MoBi/issues/1578

When commiting, the user should be able to select the module and the BB (if there are multiple, or select "create a new one") where to commit to.

rwmcintosh:
Fixes #1614

When committing simulation changes, the user now selects where the changes go instead of them always being committed to the last module of the simulation configuration.

**Selection dialog**

A single dialog offers:
- the **module** to commit to (one selection for all building block types, so committed changes always stay in the same module),
- the **parameter values building block** of that module, or "New Parameter Values Building Block",
- the **initial conditions building block** of that module, or "New Initial Conditions Building Block".

The building block selection is only shown for types that actually have tracked changes. The last module and its currently used building blocks are preselected, so confirming the defaults keeps the previous behavior. The building block lists refresh when the module selection changes, and always contain the create-new entry, so a selection can never be empty. Canceling aborts the commit. The dialog is skipped when there is no choice to make.

**Sync behavior**

- Committing to the building blocks used by the simulation (or creating new ones while none are selected) is unchanged: the template and simulation building blocks are synchronized, the change tracker is cleared and the simulation is in sync afterwards.
- Committing to a building block that is **not** used by the simulation (another existing building block, or a new one while the simulation already uses one) writes the values to the template building block only. The confirmation message warns that the simulation stays out of sync and needs to be manually configured to use the committed building block, and the change tracker is kept so the simulation remains marked as not synchronized (red).
- Committing to a module whose committed paths are **shadowed** by a building block selected in a later module configuration also warns and keeps the tracker: the later values win when the simulation is created, so the created simulation would not reproduce the committed values.

**Internals**

`AddSelectedBuildingBlockToLastModuleConfigurationCommand`/`RemoveSelectedBuildingBlockFromLastModuleConfigurationCommand` were renamed and generalized to target the selected module configuration, resolved from the module at execute time so undo/redo still works.


rwmcintosh:
@coderabbitai review please

rwmcintosh:
@coderabbitai review latest changes

Yuri05:
@rwmcintosh (I just have too many AI credits left for the remaining 2 hours 😆 )

## Review: PR #2526 — Allow selecting the module when committing simulation changes

### 1. Committing to a non-last module can silently mark the simulation as "in sync" when it isn't
`SimulationCommitTask.CommitSimulationChanges` now clears the change tracker (`ClearOriginalQuantitiesTrackerCommand`) whenever `commitsToUsedParameterValues`/`commitsToUsedInitialConditions` are true, i.e. whenever the target BB is the one selected in the chosen module configuration — regardless of which module was chosen.

The original code deliberately forced the last module (see the removed XML doc: *"The last module has to be used because if you re-create the simulation from the same modules this is the only way you will get the same simulation"*). If the user commits to an earlier module, a later module's parameter values / initial conditions still win when the simulation is rebuilt, so re-creating the simulation does **not** reproduce the committed values — yet the simulation is now flagged green/in-sync. The PR only detects "wrong building block", never "wrong module (shadowed by a later module)". Either the shadowing case must also set `simulationStaysOutOfSync`, or the user must be warned in the dialog/confirmation message.

### 2. `simulationStaysOutOfSync` is all-or-nothing across the two building block types
```csharp
var simulationStaysOutOfSync = parameterChanges.Any() && !commitsToUsedParameterValues(commitTarget) ||
                               moleculeChanges.Any() && !commitsToUsedInitialConditions(commitTarget);
...
if (!simulationStaysOutOfSync)
   macroCommand.Add(new ClearOriginalQuantitiesTrackerCommand(simulationWithChanges));
```
If parameter values go to an unused BB but initial conditions go to the used one (or vice versa), the tracker is not cleared at all. The correctly synchronized molecule changes remain recorded as pending changes and will be committed a second time on the next commit (duplicated writes / an extra "new BB" creation). The tracker clearing needs to be per quantity type, or the limitation should be documented and prevented.

### 3. `First(...)` in the renamed commands can throw instead of degrading
`AddSelectedBuildingBlockToModuleConfigurationCommand` / `RemoveSelectedBuildingBlockFromModuleConfigurationCommand`:
```csharp
private ModuleConfiguration moduleConfiguration =>
   _simulation.Configuration.ModuleConfigurations.First(x => Equals(x.Module, _existingModule));
```
- `_existingModule` is re-resolved in `ModuleContentChangedCommand.RestoreExecutionData` via `context.Get<Module>(_existingModuleId)`. If that returns an instance that is not reference-identical to the one held by `simulation.Configuration.ModuleConfigurations` (e.g. after project reload / simulation re-creation, since `Equals` on `IObjectBase` is reference equality), `First` throws `InvalidOperationException` during undo/redo. `FirstOrDefault` plus a defensive exit, or matching on `Module.Id`, would be safer.
- The property is also evaluated in `GetInverseCommand`, which dereferences `_simulation` and `_existingModule`. Both are nulled in `ClearReferences`, so any invocation after references are cleared throws a `NullReferenceException`. The previous code had the same `_simulation` dependency, but the added `_existingModule` lookup widens the failure surface. Consider capturing the module configuration identity at construction time rather than resolving it lazily.

### 4. Dialog has no validation — an empty selection silently means "create new"
`SelectCommitTargetView.SelectedParameterValues` / `SelectedInitialConditions` return `null` both for the "New ... Building Block" sentinel **and** when nothing is selected in the combo. In `SelectCommitTargetPresenter.bindBuildingBlocksFor`, if `_templateResolverTask.TemplateBuildingBlockFor(...)` returns `null` (template BB missing/renamed), `select(...)` matches nothing and the combo is left unselected. OK is still enabled, and the commit silently falls back to creating new building blocks. There is no `OKButton` enablement/validation logic in the view.

### 5. Redundant/re-entrant binding during `BindModules`
`cmbModule.SelectedIndexChanged += (o, e) => OnEvent(moduleChanged);` is wired in the constructor, so `BindModules` → `select(...)` → `SelectedIndexChanged` → `_presenter.ModuleChanged()` → `bindBuildingBlocksFor(...)` runs before `SelectCommitTargetFor` explicitly calls `bindBuildingBlocksFor(defaultModuleConfiguration)`. The building block combos are populated twice (and `ModuleChanged` fires while the presenter is still mid-initialization). Suppress the event while binding.

### 6. View caption is the designer placeholder
`SelectCommitTargetView.InitializeResources` sets `Text = AppConstants.Captions.SelectCommitTarget` but never sets `Caption`, so the designer literal remains:
```csharp
this.Caption = "SelectCommitTargetView";
```
Every other modal view in `MoBi.UI/Views` (e.g. `CalculateScaleDivisorsView`, `CreatePKSimMoleculeView`) assigns `Caption` from `AppConstants.Captions`. The user will see the raw class name as the dialog header.

### 7. Commands for the "not used" path are not hidden
`updateNotUsedParameterValues` / `updateNotUsedInitialConditions` return the synchronize commands without `.AsHidden()`, unlike every other path in this class where per-quantity synchronize commands are hidden inside the macro command. This will flood the command history with one visible entry per changed quantity.

### 8. Minor
- `SelectCommitTargetPresenter` allocates sentinel `new ParameterValuesBuildingBlock()` / `new InitialConditionsBuildingBlock()` in the constructor via `AppConstants.Captions.NewWindow(...)`, whose format is `"New {0}"` — it is a *window title* helper being reused as a list entry label. A dedicated caption would be clearer.
- `SelectCommitTargetView.select` declares `System.Func<object, bool>` fully qualified while `System` is not imported; add the `using` for consistency with the rest of the file.
- The dialog always offers both a parameter values and an initial conditions target even when only one of the two kinds of changes exists; the unused selection is silently ignored. Hiding/disabling the irrelevant combo would avoid confusion.
- `noBuildingBlockToSelectFor` calls `_templateResolverTask.TemplateModuleFor(...)`, and `bindBuildingBlocksFor` calls it again per module; no null handling if the template module cannot be resolved (deleted template) — would `NullReferenceException` on `templateModule.ParameterValuesCollection`.

Yuri05 (changes requested):
s. https://github.com/Open-Systems-Pharmacology/MoBi/pull/2526#issuecomment-5483969982

rwmcintosh:
@Yuri05 Thanks for the thorough review. Status per point:

**Addressed (968d5981, 221afe0c):**

- **Point 1 (shadowing by later modules):** Implemented as detection rather than re-restricting to the last module: each committed path is checked against the building blocks selected by later module configurations. If any committed path is shadowed, the confirmation warns that the later values win when the simulation is created, and the change tracker is kept so the simulation remains marked as not synchronized.
- **Point 5 (double binding):** The view now binds a `CommitTargetDTO` through a `ScreenBinder`, following the same pattern as the other modal views. Programmatic `BindToSource` doesn't raise `Changed`, so the re-entrant bind is gone structurally rather than suppressed.
- **Point 8 (`System.Func`):** Gone with the rewrite (the raw event helpers were removed entirely).
- **Point 8 (irrelevant combo):** The building block selection is now hidden for a type without tracked changes, and the skip-dialog rule only counts choices for types that have changes.

**Doesn't reproduce:**

- **Point 4 (empty selection / no validation):** There is no empty state to validate against. Each combo always contains at least the "New Parameter Values Building Block" / "New Initial Conditions Building Block" entry, and the preselection always resolves — to the template of the used building block, or to that create-new entry otherwise. Every state the dialog can be in is a valid commit target.
- **Point 6 (caption):** The dialog title is set via `Text` in `InitializeResources`, which runs after `InitializeComponent` and overwrites the designer placeholder — see the screenshot: the header shows "Select Module and Building Blocks".

**Pushback:**

- **Point 2 (per-type tracker clearing):** The mixed case requires tracked molecule-amount changes, and molecule-amount edits never land in `OriginalQuantityValues` — the scenario is not reachable today. If it ever becomes reachable, the consequence is an idempotent re-write on the next commit, not data corruption. Happy to revisit if IC tracking is ever implemented.
- **Point 3 (`First(...)` / lazy resolution):** The command framework guarantees `RestoreExecutionData` runs before the inverse command is built, and `context.Get<Module>(id)` returns the registered instance — the same instance the module configuration holds, so reference equality holds in-session. History reloaded from disk is display-only. The pre-existing code had the identical shape (`.Last()` evaluated on the restored simulation, `_simulation` dereferenced in `GetInverseCommand`).
- **Point 7 (visible sync commands):** Matches the existing convention: the used-building-block update path has always emitted visible per-quantity synchronize commands for the template building block (only the simulation-side duplicates are hidden). The not-used path mirrors exactly that.
- **Point 8 (`NewWindow` as label):** The same helper is already used for the same purpose — the create-new sentinel entry in `InteractionTasksForExtendablePathAndValueEntity`.
- **Point 8 (null template module):** Unguarded `TemplateModuleFor` matches the pre-existing code in this class (`createNewBuildingBlocks`); resolution failure there means the project is already in an inconsistent state.

src/MoBi.Presentation/Tasks/SimulationCommitTask.cs
```diff
@@ -103,6 +112,166 @@ private void showErrorForUntraceableChanges(IMoBiSimulation simulationWithChange
...
+      ///    Returns true when any committed path is also defined in a building block selected in a module configuration
+      ///    coming after the target module. Those values win when the simulation is created, so the created simulation
+      ///    would not reproduce the committed values
+      /// </summary>
+      private bool changesAreShadowedByLaterModule(IMoBiSimulation simulation, CommitTarget commitTarget, IReadOnlyList<(ObjectPath quantityPath, IParameter quantity)> parameterChanges, IReadOnlyList<(ObjectPath quantityPath, MoleculeAmount quantity)> moleculeChanges)
+      {
+         var laterModuleConfigurations = simulation.Configuration.ModuleConfigurations.SkipWhile(x => !Equals(x, commitTarget.ModuleConfiguration)).Skip(1).ToList();
+
+         return parameterChanges.Any(change => laterModuleConfigurations.Any(x => x.SelectedParameterValues?[change.quantityPath] != null)) ||
+                moleculeChanges.Any(change => laterModuleConfigurations.Any(x => x.SelectedInitialConditions?[change.quantityPath] != null));
```
Yuri05, line 156: Following Scenario: simulation was created with 2 modules:
* module **A** on top (= latest module) without **PV**
* module **B** with PV **PV_B**

In the simulation, I change a parameter **Organism|P1** which is defined in the spatial structure of module **A** and commit this change to **PV_B** of module **B**

`changesAreShadowedByLaterModule()` will return `false` in this case. 
But if I recreate the simulation with the existing configuration - **Organism|P1** will become its value again from the spatial structure of module **A** ... or not? 

I'm just not sure if PV of module B "wins" over the spatial structure of module A or not. If **PV_B** wins, then implementation is probably correct. If not, then not.

rwmcintosh: Yes, the PV's are always applied last so the PV_B of module B values will win.

Yuri05 (approved):

PR #2528, opened 2026-08-31
Create multiple simulations in parallel from MoBi.R

### LINKED ISSUE #2516: MoBi.R - create multiple simulations in parallel
`SimulationTask` should accept multiple `MoBi.R.Domain.SimulationRequest` and create simulations in parallel. Then return an array of Simulations.

Here is the implementation in R (currently only for one simulation) https://github.com/Open-Systems-Pharmacology/OSPSuite-R/blob/30f4b56a112dd1c19b3e908aa15aebb9fb50e31e/R/utilities-simulation.R#L10.

@rwmcintosh please check if 13 or 13.1

rwmcintosh:
Fixes #2516

`SimulationTask` now creates multiple simulations from `SimulationRequest`s in parallel, mirroring the parallel snapshot load from #2520/#2524.

### What changes

- `ISimulationTask` exposes a single `CreateSimulationsAndValidateFrom(params SimulationRequest[] requests)` returning one `SimulationCreationResult` per request in the request order. The former single-request method is folded into it as the private per-request step, following the plural-only shape of the other R-facing services (`RunSimulations`, `ImportResultsFromCSV`).
- Each request carries its name in the new `SimulationRequest.SimulationName` — the name only names the created simulation and the model root container, so it belongs on the request. The batch is then one trailing `params` array, the calling convention rSharp already binds for `SimulationRunner.RunSimulations`, working for any batch size including one. The R wrapper sets the property and splats the requests via `do.call`, as `runSimulations` does.
- The batch mirrors the hardened snapshot-load pattern: requests are created sequentially until the first success so lazily initialized services are warm, the rest are constructed in parallel bounded by `MaximumNumberOfCoresToUse`, and a request that cannot be created reports its errors in its result instead of failing the batch. Out-of-memory still propagates; a null requests array throws.

### Tests

- Four requests sharing one module are created in parallel with names and order preserved.
- A middle request with an illegal name fails into its result's `Errors` during the parallel phase while its neighbors succeed.
- A null requests array throws and an empty one returns no results; the single-request fixtures now run through the batch API.

## Summary by CodeRabbit

* **New Features**
  * Added support for creating and validating multiple simulations in a single operation.
  * Simulation names can now be provided with each simulation request.
  * Results are returned in the same order as the submitted requests.
  * Individual simulation errors are reported in results, allowing other simulations to continue processing.

### DIFF
--- a/src/MoBi.R/Services/SimulationTask.cs
+++ b/src/MoBi.R/Services/SimulationTask.cs
@@ -17,7 +17,8 @@ namespace MoBi.R.Services
    {
       /// <summary>
       ///    Creates and validates one simulation per request, named after <see cref="SimulationRequest.SimulationName" />.
-      ///    The results are returned in the request order.
+      ///    The results are returned in the request order. A request that cannot be created reports its errors in its
+      ///    result.
       /// </summary>
       SimulationCreationResult[] CreateSimulationsAndValidateFrom(params SimulationRequest[] requests);
 
@@ -85,7 +86,7 @@ namespace MoBi.R.Services
 
          var results = new SimulationCreationResult[requests.Length];
 
-         SimulationCreationResult createAt(int index)
+         bool createAt(int index)
          {
             try
             {
@@ -96,7 +97,7 @@ namespace MoBi.R.Services
                results[index] = new SimulationCreationResult(null, Enumerable.Empty<string>(), new[] { e.Message });
             }
 
-            return results[index];
+            return results[index].Simulation != null;
          }
 
          //simulations are created sequentially until one succeeds, so that lazily initialized services are
@@ -104,9 +105,9 @@ namespace MoBi.R.Services
          var warmupCount = 0;
          while (warmupCount < requests.Length)
          {
-            var result = createAt(warmupCount);
+            var success = createAt(warmupCount);
             warmupCount++;
-            if (result.Simulation != null)
+            if (success)
                break;
          }
 
--- a/src/MoBi.R/Services/SimulationTask.cs
+++ b/src/MoBi.R/Services/SimulationTask.cs
@@ -17,7 +17,7 @@ namespace MoBi.R.Services
    {
       /// <summary>
       ///    Creates and validates one simulation per request, named after <see cref="SimulationRequest.SimulationName" />.
-      ///    The simulations are created in parallel and the results are returned in the request order.
+      ///    The results are returned in the request order.
       /// </summary>
       SimulationCreationResult[] CreateSimulationsAndValidateFrom(params SimulationRequest[] requests);
 
--- a/src/MoBi.R/Services/SimulationTask.cs
+++ b/src/MoBi.R/Services/SimulationTask.cs
@@ -16,13 +16,8 @@ namespace MoBi.R.Services
    public interface ISimulationTask
    {
       /// <summary>
-      ///    Creates and validates a simulation named after <see cref="SimulationRequest.SimulationName" />.
-      /// </summary>
-      SimulationCreationResult CreateSimulationAndValidateFrom(SimulationRequest request);
-
-      /// <summary>
-      ///    Creates and validates one simulation per request. The simulations are created in parallel and the results are
-      ///    returned in the request order.
+      ///    Creates and validates one simulation per request, named after <see cref="SimulationRequest.SimulationName" />.
+      ///    The simulations are created in parallel and the results are returned in the request order.
       /// </summary>
       SimulationCreationResult[] CreateSimulationsAndValidateFrom(params SimulationRequest[] requests);
 
@@ -66,7 +61,7 @@ namespace MoBi.R.Services
          return allNamedObjects.FindByName(namedObjectToSelect);
       }
 
-      public SimulationCreationResult CreateSimulationAndValidateFrom(SimulationRequest request)
+      private SimulationCreationResult createSimulationAndValidateFrom(SimulationRequest request)
       {
          if (request == null)
             throw new InvalidArgumentException(AppConstants.Exceptions.SimulationRequestCannotBeNull);
@@ -94,7 +89,7 @@ namespace MoBi.R.Services
          {
             try
             {
-               results[index] = CreateSimulationAndValidateFrom(requests[index]);
+               results[index] = createSimulationAndValidateFrom(requests[index]);
             }
             catch (Exception e) when (!e.IsOutOfMemory())
             {
--- a/src/MoBi.R/Services/SimulationTask.cs
+++ b/src/MoBi.R/Services/SimulationTask.cs
@@ -1,12 +1,14 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading.Tasks;
 using MoBi.Assets;
 using MoBi.R.Domain;
 using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 using OSPSuite.Core.Domain.Services;
 using OSPSuite.Core.Extensions;
+using ICoreUserSettings = MoBi.Core.ICoreUserSettings;
 using ModuleConfiguration = MoBi.R.Domain.ModuleConfiguration;
 
 namespace MoBi.R.Services
@@ -14,6 +16,12 @@ namespace MoBi.R.Services
    public interface ISimulationTask
    {
       SimulationCreationResult CreateSimulationAndValidateFrom(string simulationName, SimulationRequest request);
+
+      /// <summary>
+      ///    Creates and validates one simulation per request, named after <see cref="SimulationRequest.SimulationName" />.
+      ///    The simulations are created in parallel and the results are returned in the request order.
+      /// </summary>
+      SimulationCreationResult[] CreateSimulationsAndValidateFrom(params SimulationRequest[] requests);
 
       ModuleConfiguration CreateModuleConfiguration(Module module,
          string selectedParameterValues = null,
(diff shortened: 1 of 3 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05:
@rwmcintosh 

# Review — PR #2528: Create multiple simulations in parallel from MoBi.R

## Issues

### 1. Shared building blocks are read concurrently — likely thread-safety problem
`CreateSimulationsAndValidateFrom` runs `_simulationFactory.CreateSimulationFrom` for several requests at once, and the intended usage (explicitly exercised by `when_creating_multiple_simulations`, where four requests share the same `_moduleConfiguration`) has multiple threads traversing/cloning the *same* `Module` and building-block object graphs simultaneously. Neither `ISimulationFactory`, the underlying clone managers, nor the `ObjectBaseFactory`/id generation are documented as thread-safe, and the warm-up loop only addresses lazy service initialization, not re-entrancy. The parallel snapshot load in `ProjectMapper` this is modelled after maps *distinct* snapshots, so it is not the same situation. This needs to be established (or the shared inputs cloned once per request up front) before the parallelism is safe; a passing 4-simulation test does not demonstrate absence of a race.

### 2. Breaking public API change with no migration path in this repo
`SimulationCreationResult CreateSimulationAndValidateFrom(string, SimulationRequest)` is removed from `ISimulationTask`, which is an R-facing API exposed via `Api.GetSimulationTask()`. The PR description says "The R wrapper sets the property and splats the requests via `do.call`", but no R wrapper change is in this diff, so the corresponding change in the R package is untracked. At minimum the coordinated change should be linked, and the removal should be called out as breaking rather than described as folding the method in.

### 3. Silent change from throwing to error-in-result for single-request callers
Previously an invalid configuration threw `InvalidOperationException` (see the rewritten `should_throw_expected_exception` → `should_return_the_errors_instead_of_a_simulation`). Now every failure is swallowed into `SimulationCreationResult.Errors`. Existing single-simulation callers that relied on the exception will now silently receive a result with `Simulation == null` and keep going. This is a behavioural break independent of the batching and deserves an explicit note (and, ideally, preserving throw-on-failure when a single request is passed, or documenting the new contract in the R wrapper).

### 4. Inconsistent null handling
A null `requests` array throws `InvalidArgumentException`, but a null *element* inside the array no longer throws: `createSimulationAndValidateFrom` throws `InvalidArgumentException`, `createAt` catches it, and the caller gets an error-result. The two null cases should behave the same way. There is also no test for a null element.

Related: the null-array check reuses `AppConstants.Exceptions.SimulationRequestCannotBeNull`, whose wording refers to a single request; it will read incorrectly for the array case.

### 5. `SimulationRequest.SimulationName` is unvalidated
The name moved from a required method parameter to an optional settable property, so a caller that forgets to set it now passes `null` into `CreateSimulationFrom`. There is no validation and no test covering a null/empty name; the resulting failure will be an opaque message inside `Errors` rather than a clear argument error.

### 6. Diagnostic information is lost when a request fails
The catch block records only `e.Message`:
```csharp
results[index] = new SimulationCreationResult(null, Enumerable.Empty<string>(), new[] { e.Message });
```
Inner-exception detail and stack trace are discarded, and any warnings already accumulated are replaced with an empty collection. For a batch API where failures are no longer thrown, this is the only diagnostic the caller gets — consider including inner exception messages, and logging the exception as `ProjectMapper` does via `_logger.AddException`.

### 7. `createAt` has a confusing dual contract
It both writes `results[index]` as a side effect and returns the same value, and the return value is used only by the warm-up loop (`result.Simulation != null`). The `Parallel.For` call discards it. Returning a `bool` success flag (as `mapSimulationAt` does in `ProjectMapper`) would make the warm-up intent clearer.

### 8. Test gaps
- No test that `MaximumNumberOfCoresToUse` is honoured, whereas the equivalent snapshot work is covered in `ProjectMapperSpecs`/`ProjectLoadSpecs` (`A.CallTo(() => userSettings.MaximumNumberOfCoresToUse).Returns(...)`). The new `_userSettings` dependency is effectively untested.
- No test for out-of-memory propagation, despite it being called out as intended behaviour in the description.
- `should_create_all_simulations_from_the_shared_module` asserts `FirstOrDefault(...)?.Module.ShouldNotBeNull()` — the null-conditional makes the assertion vacuous when no module configuration matches `"Module1"`. It should assert the `FirstOrDefault` result is non-null.
- `when_creating_multiple_simulations_and_one_cannot_be_created` relies on the failing request being at index 1 to be handled in the parallel phase, which is only true because index 0 succeeds first. That coupling is implicit; a comment or an explicit assertion on ordering would make the intent robust.

### 9. Missing newline at end of file
`src/MoBi.R/Services/SimulationTask.cs` ends without a trailing newline (`\ No newline at end of file` in the diff).

rwmcintosh:
@Yuri05 point by point:

**Point 1 — thread safety of shared building blocks.** The premise that the parallel snapshot load maps distinct object graphs doesn't hold: `ModuleConfigurationMapper` resolves each simulation's modules from the project by name, so parallel snapshot constructions already share the same `Module` instances. `When_loading_a_snapshot_with_simulations_sharing_modules_in_parallel` (ProjectLoadSpecs) pins exactly this — four simulations sharing one module built in parallel, compared structurally against a sequential load. The re-entrancy work it relies on landed in #2518 (per-run dimension-validation state) and #2520/#2524 (per-call clone managers in both factories, each pinned by a spec). This PR goes through the same `CreateSimulationAndValidate` path, and its own spec exercises four requests sharing one module.

**Point 2 — breaking change / migration.** The MoBi.R API is unreleased, so there is no released consumer to migrate. The OSPSuite-R wrapper update is the immediate next piece of work and is tracked in Open-Systems-Pharmacology/OSPSuite-R#2024 (referenced from #2516).

**Point 3 — throw vs error-in-result.** Intentional: with a single batch API there is one per-request contract — a request that cannot be created reports through `Errors`; only a malformed call (null array) throws. The R wrapper raises `stop()` from `Errors`, so R-visible behavior is unchanged. The contract is now stated in the XML doc.

**Point 4 — null handling.** Also intentional: a null array is a malformed call, a null element is one bad request that must not fail its siblings — the same isolation the snapshot load applies per simulation. A spec for the null element was added.

**Point 5 — name validation.** The name is validated where it always was, in `SimulationFactory.CreateSimulationFrom`: a missing name yields "Simulation name is required" in `Errors`, not an opaque failure. A spec now pins that.

**Point 6 — diagnostics.** The catch in the batch method only ever sees the pre-factory throws (null request, name validation), whose messages are complete on their own. Construction failures are caught inside `SimulationFactory` exactly as before this PR, and no warnings exist on the exception path to be discarded.

**Point 7 — accepted:** `createAt` now returns a success flag, as `mapSimulationAt` does.

**Point 8 —** the vacuous assertion is fixed, and specs for the missing name, the null element and out-of-memory propagation were added. A core-count test is out of scope: nothing in this PR sets `MaximumNumberOfCoresToUse`. The failing-request spec is constructed so the middle request lands in the parallel phase; that is the spec's construction, not a hidden coupling needing commentary.

**Point 9 —** trailing newlines at EOF are not a convention in this repo.

src/MoBi.R/Services/SimulationTask.cs
```diff
@@ -53,21 +62,64 @@ private T selectByName<T>(IReadOnlyList<T> allNamedObjects, string namedObjectTo
...
+         while (warmupCount < requests.Length)
+         {
+            var success = createAt(warmupCount);
+            warmupCount++;
+            if (success)
+               break;
+         }
+
+         if (requests.Length > warmupCount)
+            Parallel.For(warmupCount, requests.Length, parallelOptions(), index => createAt(index));
```
Yuri05, line 115: shouldn't it be 
```
if (requests.Length > warmupCount + 1)
   Parallel.For(warmupCount + 1, requests.Length, parallelOptions(), index => createAt(index));
```
?
simulation at warmupCount was already created in the line 108 above.

src/MoBi.R/Services/SimulationTask.cs
Yuri05, line 1: AFAI understand, the logic is: if at least one of the simulations could not be created for any reason other than **OutOfMemory** - the whole process stops with error.
But if simulation(s) could not be created due to **OutOfMemory** - their creation result is null and all other simulations are created and returned to the caller.
What's the logic behind this strategy?

src/MoBi.R/Services/SimulationTask.cs
rwmcintosh: no, I don't think so, In a walkthrough
```
createAt(0), 
warmupCount = 1, 
break; 
Parallel.For(1, 4) //creates the rest
```

src/MoBi.R/Services/SimulationTask.cs
rwmcintosh: not quite

OOM means abort the whole call by not handling the OOM exception

All others are caught and result carries the error, continue with the rest of the requests.

The idea is that you cannot recover OOM by continuing in the next request because OOM is not related to the specific simulation, and will probably still throw on the next one in the list.

src/MoBi.R/Services/SimulationTask.cs
Yuri05: yep, you are right, it's incremented after creation and before the break

src/MoBi.R/Services/SimulationTask.cs
Yuri05: ah, true it's `catch (Exception e) when (!e.IsOutOfMemory())`
(and I have read it as `catch (Exception e) when (e.IsOutOfMemory()`) lol

Yuri05: makes totally sense then :)

Yuri05 (approved):

PR #2529, opened 2026-09-01
Fixes #825 Apply the molecule's molecular weight to imported observed data

### LINKED ISSUE #825: Choose molecular weight when reading observed data
When reading observed data in MoBi, although the mapping to a molecule works, the molecular weight of the Molecules building block is not used. Therefore, for plotting etc it is needed to enter the molecular weight manually (or as an extra column in the data). It would be great if the molecular weight is automatically taken from the building block. If there are several molecules with the same name in different Molecule building blocks, the user should be able to select which one to use.

rwmcintosh:
Fixes #825

Assigning a molecule in the observed data importer never transferred that molecule's molecular weight to the imported data set, so it had to be typed in by hand or carried as an extra column.

**Cause.** Core reads the MW from the *value* of the `Molecule` meta data category's `ListOfValues` and parses it as a number in g/mol. MoBi stored `name -> name` there, so the parse always failed and the MW was silently left unset. PK-Sim stores `name -> MolWeight` and works.

Not a recent regression: the lookup was added in 60e88cea0 and removed three days later in 3de65b346. `git tag --contains` returns the same tags for both commits, so no released version ever had it.

**Change.** Store the molecule's `Molecular weight` parameter converted into g/mol, which is the unit Core converts back from — the value reaches Core without a unit attached, so passing the display value through would send `148` for a 148 kDa antibody. Molecules with a `NaN` weight contribute an empty string and are skipped as before.

`predefinedValuesForCategory` now returns the list's keys rather than its values, otherwise the meta data editor's Molecule dropdown would start offering weights instead of molecule names.

Ambiguity when several modules define the same molecule name is out of scope here and tracked in #2530.

**Testing.** `MoBi.Tests` green (2735 passed, 0 failed). Four new observations cover the g/mol value, the `NaN` case, the dropdown keys and `PredefinedValuesFor`. Verified by hand in MoBi against an ADC project: `ADC` (148 kDa) yields 148000 g/mol, `nAb` yields 147220, and the small-molecule payload `Toxophore` yields 780.

## New Features

- Observed-data import metadata now includes molecular weights for predefined molecules.
- Molecular weights are converted to each parameter’s default unit, including g/mol where applicable.
- Predefined molecule metadata is available for import selection and display.

## Bug Fixes

- Missing or invalid molecular-weight values now appear as blank instead of causing import issues.

### DIFF
--- a/src/MoBi.Presentation/Tasks/ObservedDataTask.cs
+++ b/src/MoBi.Presentation/Tasks/ObservedDataTask.cs
@@ -360,7 +360,18 @@ namespace MoBi.Presentation.Tasks
             return;
 
          metaDataCategory.ShouldListOfValuesBeIncluded = true;
-         allMolecules().OrderBy(molecule => molecule.Name).Each(molecule => addInfoToCategory(metaDataCategory, molecule));
+         allMolecules().OrderBy(molecule => molecule.Name)
+            .Each(molecule => addInfoToCategory(metaDataCategory, molecule, molWeightInDefaultUnitFor(molecule)));
+      }
+
+      private static string molWeightInDefaultUnitFor(IContainer molecule)
+      {
+         var molWeightParameter = molecule.Parameter(Parameters.MOL_WEIGHT);
+         if (molWeightParameter == null || double.IsNaN(molWeightParameter.Value))
+            return string.Empty;
+
+         var dimension = molWeightParameter.Dimension;
+         return dimension.BaseUnitValueToUnitValue(dimension.DefaultUnit, molWeightParameter.Value).ToString();
       }
 
       private IEnumerable<IContainer> allMolecules()
@@ -451,9 +462,9 @@ namespace MoBi.Presentation.Tasks
          }
       }
 
-      private void addInfoToCategory(MetaDataCategory metaDataCategory, IContainer container)
+      private void addInfoToCategory(MetaDataCategory metaDataCategory, IContainer container, string displayValue = null)
       {
-         metaDataCategory.ListOfValues.Add(container.Name, container.Name);
+         metaDataCategory.ListOfValues.Add(container.Name, displayValue ?? container.Name);
 
          var icon = ApplicationIcons.IconByName(container.Icon);
          if (icon != ApplicationIcons.EmptyIcon)
@@ -478,7 +489,7 @@ namespace MoBi.Presentation.Tasks
       {
          var category = new MetaDataCategory();
          action(category);
-         return category.ListOfValues.Values;
+         return category.ListOfValues.Keys;
       }
 
       public IReadOnlyList<string> DefaultMetaDataCategories => new[]
--- a/tests/MoBi.Tests/Presentation/Tasks/ObservedDataTasksSpecs.cs
+++ b/tests/MoBi.Tests/Presentation/Tasks/ObservedDataTasksSpecs.cs
@@ -22,6 +22,8 @@ using OSPSuite.Infrastructure.Import.Core;
 using OSPSuite.Infrastructure.Import.Services;
 using OSPSuite.Utility.Extensions;
 using ImporterConfiguration = OSPSuite.Core.Import.ImporterConfiguration;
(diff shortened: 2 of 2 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05:
@coderabbitai full review

Yuri05:
@coderabbitai full review

Yuri05:
@coderabbitai review

Yuri05:
@rwmcintosh @PavelBal If we have a molecule with the same name, which is defined in several modules but has **different** mol weights there - what should happen? 
The scenario does not make much sense, but is technically feasible - should we at least check and produce a warning for this? (In PK-Sim we can only have 1 molecule with the same name, so MW-Assignment cannot be ambiguous there).

At the moment, one of those molecules will be randomly picked by `allMolecules()` and its mol weight will be assigned.
https://github.com/Open-Systems-Pharmacology/MoBi/blob/b7131888ffc0eadcb3f3a069a1f9568a48e959d3/src/MoBi.Presentation/Tasks/ObservedDataTask.cs#L377-L381

Yuri05:
Ok, just saw the comment in the PR description. Ok...
> Ambiguity when several modules define the same molecule name is out of scope here and tracked in https://github.com/Open-Systems-Pharmacology/MoBi/issues/2530.

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -360,7 +360,18 @@ private void addPredefinedMoleculesForImporter(MetaDataCategory metaDataCategory
...
+      }
+
+      private static string molWeightInDefaultUnitFor(IContainer molecule)
+      {
+         var molWeightParameter = molecule.Parameter(Parameters.MOL_WEIGHT);
+         if (molWeightParameter == null || double.IsNaN(molWeightParameter.Value))
+            return string.Empty;
+
+         var dimension = molWeightParameter.Dimension;
+         return dimension.BaseUnitValueToUnitValue(dimension.DefaultUnit, molWeightParameter.Value).ToString();
```
Yuri05, line 374: This will use the decimal separator from the system language settings 
(e.g. on my system 123.456 g/mol => "123,456").
Not a problem when converting MW back to double from the saved string?

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
```diff
@@ -478,7 +489,7 @@ private IEnumerable<string> predefinedValuesForCategory(Action<MetaDataCategory>
       {
          var category = new MetaDataCategory();
          action(category);
-         return category.ListOfValues.Values;
+         return category.ListOfValues.Keys;
```
Yuri05, line 492: why that?

rwmcintosh: `addInfoToCategory` used to store name -> name, so .Keys and .Values were interchangeable. Now that the value slot holds the MW (which is where Core's importer reads it from), PredefinedValuesFor would hand the meta data editor's Molecule dropdown 148000/147220 instead of ADC/nAb. Organ and Compartment are unaffected, they still store name -> name.

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
rwmcintosh: Core parses with `double.TryParse(molecularWeightValueAsString, out var molWeight)` using CurrentCulture. We probably should unify conversions on invariant, but that's beyond the scope here.

That's by inspection though. If you would be so kind as to verify this one yourself using German locale, I'd appreciate it.

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
Yuri05: Ok, I see, thx.

src/MoBi.Presentation/Tasks/ObservedDataTask.cs
Yuri05: Will do after PR merge (if I'll not forget :))

Yuri05 (approved):

PR #2533, opened 2026-09-01
Fixes #2531 Resolve initial conditions by path when inverting update commands

### LINKED ISSUE #2531: Error when undoing deletion of IC entries
- Delete some entries in an IC BB
- Try to rollback to that state in the history:

Application:
MoBi® 13.0.53

Stack trace:
```
at MoBi.Core.Commands.UpdateInitialConditionIsPresentCommand..ctor(IBuildingBlock`1 initialConditionsBuildingBlock, InitialCondition initialCondition, Boolean isPresent)
   at MoBi.Core.Commands.UpdateInitialConditionIsPresentCommand.GetInverseCommand(IMoBiContext context)
   at OSPSuite.Core.Commands.OSPSuiteReversibleCommand`1.InverseCommand(TContext context)
   at OSPSuite.Core.Commands.Core.RollBackCommand`1.<>c__DisplayClass3_0.<Execute>b__0(IReversibleCommand`1 command)
   at OSPSuite.Utility.Extensions.EnumerableExtensions.Each[T](IEnumerable`1 list, Action`1 action)
   at OSPSuite.Core.Commands.Core.RollBackCommand`1.Execute(TExecutionContext context)
   at OSPSuite.Core.Commands.Core.HistoryManager`1.RollBackTo(Int32 state)
```

rwmcintosh:
Fixes #2531

The three `UpdateInitialCondition*` commands re-resolved their entry from the global id repository. Entries added after their building block was registered — by Extend, by import, or as the inverse of a delete — are never registered there, so the lookup returned `null` and building the inverse command threw an NRE during rollback. This is why only entries added during the session reproduced it.

They now resolve the entry by path from its building block, as every other path-and-value-entity command already does. The type parameter widens to `ILookupBuildingBlock<InitialCondition>` to get `ByPath`; both implementing building blocks and all call sites already have that type.

The two revert specs now stub the id lookup to `null` — the production behaviour for an unregistered entry — and fail on the old code. The scale-divisor revert spec was vacuous (old and new value both `0`, empty building block) and was tightened.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when updating initial-condition settings, including presence, negative-value handling, and scale divisors.
  * Undo and redo operations now continue to work when initial conditions are added after their building block is registered.

* **Tests**
  * Expanded coverage for restoring and reverting initial-condition changes in dynamically updated building blocks.

### DIFF
--- a/src/MoBi.Core/Commands/UpdateInitialConditionIsPresentCommand.cs
+++ b/src/MoBi.Core/Commands/UpdateInitialConditionIsPresentCommand.cs
@@ -2,20 +2,21 @@ using MoBi.Assets;
 using MoBi.Core.Domain.Model;
 using OSPSuite.Assets;
 using OSPSuite.Core.Commands.Core;
+using OSPSuite.Core.Domain;
 using OSPSuite.Core.Domain.Builder;
 
 namespace MoBi.Core.Commands
 {
-   public class UpdateInitialConditionIsPresentCommand : BuildingBlockChangeCommandBase<IBuildingBlock<InitialCondition>>
+   public class UpdateInitialConditionIsPresentCommand : BuildingBlockChangeCommandBase<ILookupBuildingBlock<InitialCondition>>
    {
-      private readonly string _initialConditionId;
+      private readonly ObjectPath _initialConditionPath;
       private InitialCondition _initialCondition;
       private readonly bool _oldIsPresent;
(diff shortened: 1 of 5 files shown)


rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

PR #2540, opened 2026-09-14
Fixes #2539 Only publish nuget packages from nightly builds

### LINKED ISSUE #2539: Only publish nuget packages from nightly builds
We need to make sure that the nuget packages consumed by OSPSuite-R always come from a nightly because we want the binaries included in OSPSuite-R to match the binaries released with the app.

Right now, we publish nuget each merge to develop and so it would actually be impossible to have the same binaries in both release paths.

rwmcintosh:
Fixes #2539

The nuget packages consumed by OSPSuite-R have to come from the same build that produces the released installer. Publishing on every merge to develop made that impossible, since the two workflows have independent run counters.

### `build-nightly_v13.yml`
- Added the `Pack the project` step (after `Test`) and the `Publish to GitHub registry` step, both lifted unchanged from the develop workflow.
- `Publish` is the **last** step, after the installer and portable artifacts are uploaded, so a package is only ever pushed from a run that also produced a signed installer.
- `permissions: packages: read` -> `write`.
- The single `APP_VERSION` now drives the assembly version, the nupkg `PackageVersion`, the MSI and the portable zip, so `MoBi.13.0.180.msi` ships alongside `MoBi.Core.13.0.180.nupkg`.

### Version offset
The nightly run counter is far behind the counter of the workflow that has been publishing (nightly is at 68, publish at 174), so an un-offset nightly would try to push `13.0.69` and collide with versions already in the registry. `APP_VERSION` is therefore offset by `+111`, putting the first nightly at `13.0.180` — just past the last published `13.0.174`, with a little headroom. The offset is commented and should be dropped at the next version bump.

GitHub Actions expressions have no arithmetic operators, so the arithmetic is done in PowerShell and `run-name` cannot show the computed version; it now shows the run number instead, with the real version visible on the artifact names.

### `build-and-publish_v13.yml` deleted, folded into `build-and-test.yml`
Once the publishing is gone, that workflow was just a second, hand-rolled copy of the build-and-test that already runs on pull requests — and the only thing it did differently was stamp a live `13.0.<run_number>`, which is meaningless now that nothing is published. So it is deleted and `build-and-test.yml` gains a `push: develop` trigger, delegating to the shared `test-csharp.yml` with the throwaway `temp-version: 13.0.9999`.

This is also how `build-and-test.yml` was set up in 12.3, before the publish workflow was split out.

Merges to develop are still built and tested, they are just no longer published.

The identical change for PK-Sim is Open-Systems-Pharmacology/PK-Sim#3754.


.github/workflows/build-nightly_v13.yml
```diff
@@ -1,5 +1,5 @@
 name: Build Nightly 13.0
-run-name: Version 13.0.${{ github.run_number }}
+run-name: Nightly build ${{ github.run_number }}
```
Yuri05, line 2: let's keep 13.0 in the run name
`run-name: Nightly build 13.0.${{ github.run_number }}` or 
`run-name: Version 13.0.${{ github.run_number }}` (as before) or
`run-name: Nightly 13.0.${{ github.run_number }}`

.github/workflows/build-and-test_v13.yml
```diff
@@ -1,16 +1,17 @@
-name: Build and Publish V13
-run-name: Version 13.0.${{ github.run_number }} (${{ github.event.head_commit.message }})
+name: Build and Test V13
```
Yuri05: We have already Build and Test workflow (on PR) which internal name is exactly the same
(`name: Build and Test 13.0`)
Any problems/side effects to expect? 
Also yaml file names are now **build-and-test.yml** and **build-and-test_v13.yml**) - somehow confusing anyways. 
Should we try to combine both workflows into one yaml instead?

.github/workflows/build-nightly_v13.yml
```diff
@@ -52,8 +52,10 @@ jobs:
           nuget sources add -username Open-Systems-Pharmacology -password ${{ secrets.GITHUB_TOKEN }} -name OSP-GitHub-Packages -source "https://nuget.pkg.github.com/Open-Systems-Pharmacology/index.json"
 
       - name: define env variables
+        # offset the run number past 13.0.174, the last version published from the develop workflow
+        # drop the offset when the version is bumped to the next release
         run: |
-          echo "APP_VERSION=13.0.${{ github.run_number }}" | Out-File -FilePath $env:GITHUB_ENV -Append
+          echo "APP_VERSION=13.0.$(${{ github.run_number }} + 111)" | Out-File -FilePath $env:GITHUB_ENV -Append
```
Yuri05, line 58: +111 to continue the current build number? lol. I mean, why not

Yuri05 (changes requested):
I think we should combine on_push and on_pr workflows into single yaml then, othewise it becomes pretty confusing.

.github/workflows/build-and-test_v13.yml
rwmcintosh: let's rename one to build_and_test_pr.yml

.github/workflows/build-nightly_v13.yml
rwmcintosh: the run number is not the publish version anymore. That's the problem

.github/workflows/build-nightly_v13.yml
rwmcintosh: yeah. It's the delta between the merge to develop and the nightly run numbers.

.github/workflows/build-and-test_v13.yml
rwmcintosh: combined into one

.github/workflows/build-nightly_v13.yml
Yuri05: Can you put the same offset there? 
`run-name: Version 13.0.$(${{ github.run_number }} + 111)`
Or is it not properly evaluated in the run-name?

rwmcintosh: Not evaluated at the time of naming

Yuri05 (approved):

PR #2543, opened 2026-09-15
Fixes #2542 Deserialize building blocks created by the PK-Sim starter

### LINKED ISSUE #2542: Individual - reload from snapshot - unexpected differences
1. Create an individual in an empty MoBi project
2. Right-click -> Snapshot -> Reload Individual
3. Compare the individuals

Origin data "BMI" is missing in the recreated individual:

original:

Image

recreated:

Image

Comparison of the two individuals lists every path, but without stating the actual difference:

Image

rwmcintosh:
Fixes #2542 — MoBi half. Companion PK-Sim changes: https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3760 (Starter returns PKML, https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3758) and https://github.com/Open-Systems-Pharmacology/PK-Sim/pull/3759 (BMI origin data, https://github.com/Open-Systems-Pharmacology/PK-Sim/issues/3757).

## Root cause
`PKSimStarter.executeAndCloneBuildingBlock` received the **live** `IndividualBuildingBlock` / `ExpressionProfileBuildingBlock` from `PKSim.Starter` and only cloned it, so the object kept PK-Sim's `IDimension` instances (and anything else it referenced from the PK-Sim container). Snapshot reloads already go through PKML serialization, so the reloaded copy carries MoBi's instances — comparing the two reports a `Dimension` difference on every path while printing identical names.

## Fix
Every exchange with PK-Sim now goes through one door, `PKSimSnapshotConverterBase.DeserializeFromPKSim<T>`, which is what the four snapshot loaders already did inline. `executeAndCloneBuildingBlock` expects the PKML string that PK-Sim#3760 returns, deserializes it in the MoBi container, then clones as before. Cancellation (`null`) is unchanged. This also means the `PKSimSpatialStructure → MoBiSpatialStructure` conversion the serialization comment mentions now applies to created building blocks too.

## ⚠️ Coupled with PK-Sim#3760
Against a PK-Sim without #3760 the starter still returns an object; `as string` yields `null` and the creation reads as cancelled. Both must ship in V13 together. Draft until PK-Sim#3760 is in.

## Tests
- New `When_creating_an_individual_with_PKSim` (red against the old starter, green now) and `When_creating_an_individual_with_PKSim_was_cancelled`.
- 135 exchange-related specs (PKSim*, InteractionTasksFor*, Snapshot*, ExpressionProfileBuildingBlockPresenter*) pass.
- Full `MoBi.Tests` suite: 2737 passed, 0 failed, 4 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of building blocks created through PK-Sim by ensuring returned data is properly loaded before cloning.
  * Preserved cancellation behavior: cancelling PK-Sim creation continues to return no building block.

* **Tests**
  * Added coverage for successful PK-Sim individual creation, cloning, and cancellation scenarios.


rwmcintosh:
Validated end to end against PK-Sim `develop` at `f32fa910` (with #3759 and #3760 merged): create individual → Snapshot → Reload Individual → compare is clean — BMI present in both origin-data rows, no `Dimension` differences.

Side finding while validating: a stale PK-Sim path silently produced the "cancelled" behaviour, tracked in https://github.com/Open-Systems-Pharmacology/MoBi/issues/2544.

rwmcintosh:
@coderabbitai review please

Yuri05 (approved):

